PR #9, opened 2024-07-05
fix: remove savings balance card

reneaaron:


rolznz:
Decided to not go with this idea, closing for now

PR #11, opened 2024-07-05
Feat: node health indicator

rolznz:
See https://github.com/getAlby/nostr-wallet-connect-next/pull/532


rolznz:
@reneaaron I got the :ok: from @stackingsaunter :-)

PR #13, opened 2024-07-05
feat: support limit and offset in list transactions

im-adithya:
- Adds support for `limit` and `offset` parameters
- Also adds the transaction hint in the Onboardsing Checklist

Screenshot 2024-07-04 at 6 03 00 PM

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -487,12 +487,11 @@ func (api *api) GetBalances(ctx context.Context) (*BalancesResponse, error) {
 	return balances, nil
 }
 
-// TODO: accept offset, limit params for pagination
-func (api *api) ListTransactions(ctx context.Context) (*ListTransactionsResponse, error) {
+func (api *api) ListTransactions(ctx context.Context, limit uint64, offset uint64) (*ListTransactionsResponse, error) {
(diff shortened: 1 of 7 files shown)


im-adithya:
See https://github.com/getAlby/nostr-wallet-connect-next/pull/572 for more info

rolznz:
@im-adithya should we wait for the dynamic budgets / internal transactions PR to be merged then? because if we add this code but it is not used on the frontend, it seems wrong to me

im-adithya:
Hmmm, the reason why I added this is because we need only one transaction to check the transaction item in the list, so there I'm using the limit, if you think this causes many conflicts in dynamic budgets, I can close, upto you 👍

rolznz (approved):
tACK

rolznz:
@im-adithya ah, I see. Thanks!

rolznz:
@im-adithya we could do even better by also supporting the type param (See https://github.com/nostr-protocol/nips/blob/master/47.md#list_transactions) and then we can have checklist items for both sending and receiving by getting the first incoming and outgoing transactions

PR #14, opened 2024-07-05
feat: revamp permissions component

im-adithya:
### No Params

Screenshot 2024-07-12 at 7 58 20 PM

### `max_amount=172166`

Screenshot 2024-07-12 at 7 59 17 PM

### `expires_at=1721673000`

Screenshot 2024-07-12 at 7 58 45 PM

### `request_methods=make_invoice pay_invoice&max_amount=172166`

Screenshot 2024-07-12 at 7 59 49 PM

### `request_methods=make_invoice pay_invoice&expires_at=1721673000`

Screenshot 2024-07-12 at 8 00 37 PM

### `request_methods=make_invoice pay_invoice&expires_at=1721673000`

Screenshot 2024-07-12 at 8 00 37 PM

### `request_methods=make_invoice pay_invoice&expires_at=1721673000&max_amount=212121`

Screenshot 2024-07-12 at 8 04 43 PM


frontend/src/components/Permissions.tsx
```diff
@@ -107,213 +157,240 @@ const Permissions: React.FC<PermissionsProps> = ({
...
-            new Date(permissions.expiresAt).getFullYear() !== 1
-              ? new Date(permissions.expiresAt).toString()
-              : "This app will never expire"}
-          </p>
-        </>
+      {expireOption && (
+        <div className="mb-6">
+          <p className="font-medium text-sm mb-2">Connection expiration</p>
+          <div className="grid grid-cols-2 md:grid-cols-6 gap-2 text-xs mb-4">
+            {Object.keys(expiryOptions).map((expiry) => {
```
rolznz, line 332: if you pass an `expires_at` it should not be editable and there should not be options to choose from

frontend/src/components/Permissions.tsx
```diff
@@ -1,50 +1,119 @@
...
   validBudgetRenewals,
 } from "src/types";
 
+const getTimeZoneDirection = () => {
+  const offset = new Date().getTimezoneOffset();
+
+  return offset <= 0 ? +1 : -1;
+};
+
+const daysFromNow = (date?: Date) => {
```
rolznz, line 41: we have dayjs

frontend/src/components/Permissions.tsx
im-adithya: After writing the logic, I felt maybe editable is not necessary(?)

We show the user what all options have been marked and the user can simply click next. If they want to change anything they can. Instead of adding more logic we can maybe add a alert box on top which says "The app might not function properly without the following permissions, change them at your own risk" (probably something less threatening)

frontend/src/components/Permissions.tsx
```diff
@@ -1,50 +1,119 @@
...
   Scope,
+  WalletCapabilities,
   budgetOptions,
   expiryOptions,
   iconMap,
   scopeDescriptions,
   validBudgetRenewals,
 } from "src/types";
 
+const getTimeZoneDirection = () => {
```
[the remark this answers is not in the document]
im-adithya: This is because the Date Picker component gives 00:00 <your time zone> date. And we deal with UTC date. So we have to +/- day depending on the time zone direction. I still feel expiring the app on 23:59 of the time zone of the user is more sensible than UTC 23:59

frontend/src/components/Permissions.tsx
```diff
@@ -1,50 +1,119 @@
...
   budgetUsage,
 }) => {
   const [permissions, setPermissions] = React.useState(initialPermissions);
-  const [days, setDays] = useState(isNewConnection ? 0 : -1);
-  const [expireOptions, setExpireOptions] = useState(!isNewConnection);
-  const { data: capabilities } = useCapabilities();
+  const [expiryDays, setExpiryDays] = useState(
+    daysFromNow(permissions.expiresAt)
+  );
+  const [budgetOption, setBudgetOption] = useState(
```
rolznz, line 82: can you please do better with naming and review all the variables in this PR?

Think what this variable does and what it should be called so I understand when I read it

frontend/src/components/Permissions.tsx
```diff
@@ -107,213 +157,240 @@ const Permissions: React.FC<PermissionsProps> = ({
...
-                              {permissions.budgetRenewal || "Never"}
-                            </td>
-                          </tr>
-                        </tbody>
-                      </table>
-                    )}
+                    Custom...
+                  </div>
+                </div>
+                {customBudget && (
```
rolznz, line 293: if the app provides `max_amount` then this should not be editable

frontend/src/components/Permissions.tsx
```diff
@@ -107,213 +157,240 @@ const Permissions: React.FC<PermissionsProps> = ({
...
-              <li
-                key={index}
-                className={cn(
-                  "w-full",
-                  scope == "pay_invoice" ? "order-last" : "",
-                  !canEditPermissions && !permissions.scopes.has(scope)
-                    ? "hidden"
-                    : ""
-                )}
+  return !canEditPermissions ? (
```
rolznz, line 160: `canEditPermissions` is completely wrong. As we discussed, if the app sets  something, then that thing cannot be edited. No matter if it's the name, request methods, max amount, expiry, ...

frontend/src/types.ts
```diff
@@ -58,10 +65,12 @@ export type Scope =
   | "sign_message"
   | "notifications"; // covers all notification types
 
+export type ScopeGroupType = "send_receive" | "only_receive" | "custom";
+
 export type Nip47NotificationType = "payment_received" | "payment_sent";
 
 export type IconMap = {
-  [key in Scope]: LucideIcon;
+  [key in string]: LucideIcon;
```
rolznz, line 73: Now there is no type safety and we will not know when we introduce a new scope. Can you undo the change and add a `ScopeIconMap` and any `OtherIconMap` that is typed

frontend/src/types.ts
```diff
@@ -75,6 +84,12 @@ export const iconMap: IconMap = {
   [NIP_47_NOTIFICATIONS_PERMISSION]: Bell,
 };
 
+export const scopeGroupIconMap: IconMap = {
+  [SCOPE_GROUP_SEND_RECEIVE]: ArrowDownUp,
```
rolznz, line 88: The way you implemented scope groups has no type safety. Can you add a `type ScopeGroup = "send_receive" | "only_receive" | "custom"`

rolznz: Ah, I see you have `ScopeGroupType` but it's not always used. Sometimes you just use strings

frontend/src/components/Scopes.tsx
```diff
@@ -0,0 +1,181 @@
...
+  scopeDescriptions,
+  scopeGroupDescriptions,
+  scopeGroupIconMap,
+  scopeGroupTitle,
+} from "src/types";
+
+// TODO: this runs everytime, use useEffect
+const scopeGrouper = (scopes: Set<Scope>) => {
+  if (
+    scopes.size === 2 &&
```
rolznz, line 24: why would our default group only have these 2 scopes? I would expect the default group to have all scopes

frontend/src/components/Scopes.tsx
```diff
@@ -0,0 +1,181 @@
...
+    scopes.has(NIP_47_PAY_INVOICE_METHOD)
+  ) {
+    return "send_receive";
+  } else if (scopes.size === 1 && scopes.has(NIP_47_MAKE_INVOICE_METHOD)) {
+    return "only_receive";
+  }
+  return "custom";
+};
+
+const validScopeGroups = (capabilities: WalletCapabilities) => {
```
rolznz, line 35: all our LNClients support make invoice and pay invoice, this is not necessary

frontend/src/components/Scopes.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+const Scopes: React.FC<ScopesProps> = ({
+  capabilities,
+  scopes,
+  onScopeChange,
+}) => {
+  const [scopeGroup, setScopeGroup] = React.useState(scopeGrouper(scopes));
+  const scopeGroups = validScopeGroups(capabilities);
+
+  // TODO: EDITABLE PROP
```
rolznz, line 60: review TODO

rolznz:
@im-adithya we had a meeting with @stackingsaunter and @reneaaron to review this, and the implementation is not the same as I thought we agreed upon. If an app passes anything in, then that should not be able to be edited.

And the scope group of "Send and Receive" makes no sense to me as a default if it discards ALL the scopes except for those two scopes. Why do we do that? why do we want to restrict apps to only these two scopes by default?

frontend/src/screens/apps/NewApp.tsx
```diff
@@ -53,31 +55,44 @@ const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
...
-  const maxAmountParam = queryParams.get("max_amount") ?? "";
-  const expiresAtParam = queryParams.get("expires_at") ?? "";
 
   const initialScopes: Set<Scope> = React.useMemo(() => {
     const methods = reqMethodsParam
       ? reqMethodsParam.split(" ")
-      : capabilities.methods;
-
+      : // this is done for scope grouping
+        capabilities.methods.includes(NIP_47_MAKE_INVOICE_METHOD)
```
rolznz, line 77: please undo this stuff, it makes no sense

frontend/src/screens/apps/NewApp.tsx
```diff
@@ -53,31 +55,44 @@ const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
...
+          : [NIP_47_MAKE_INVOICE_METHOD]
+        : capabilities.methods;
     const requestMethodsSet = new Set<Nip47RequestMethod>(
       methods as Nip47RequestMethod[]
     );
+
+    const notificationTypes = notificationTypesParam
+      ? notificationTypesParam.split(" ")
+      : // this is done for scope grouping
+        !capabilities.methods.includes(NIP_47_MAKE_INVOICE_METHOD)
```
rolznz, line 89: here too

frontend/src/types.ts
im-adithya: I just skipped it here as it was causing issues (or could've added icon maps separately for methods and scopes) Should I proceed with the later?

frontend/src/types.ts
```diff
@@ -100,6 +126,18 @@ export const scopeDescriptions: Record<Scope, string> = {
...
 };
 
+export const scopeGroupTitle: Record<ScopeGroupType, string> = {
+  [SCOPE_GROUP_FULL_ACCESS]: "Full Access",
+  [SCOPE_GROUP_READ_ONLY]: "Read Only",
+  [SCOPE_GROUP_CUSTOM]: "Custom",
+};
+
+export const scopeGroupDescriptions: Record<ScopeGroupType, string> = {
+  [SCOPE_GROUP_FULL_ACCESS]: "Complete wallet control",
```
im-adithya, line 136: Review scope group descriptions

frontend/src/components/Scopes.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const readOnlyScopes: Set<ReadOnlyScope> = React.useMemo(() => {
+    const scopes: Scope[] = capabilities.scopes;
+    return new Set(
+      scopes.filter((method): method is ReadOnlyScope =>
+        [
+          NIP_47_GET_BALANCE_METHOD,
+          NIP_47_GET_INFO_METHOD,
+          NIP_47_LOOKUP_INVOICE_METHOD,
+          NIP_47_LIST_TRANSACTIONS_METHOD,
+          NIP_47_NOTIFICATIONS_PERMISSION,
```
im-adithya, line 52: Are these good for Read Only Scopes?

frontend/src/components/Permissions.tsx
```diff
@@ -1,274 +1,235 @@
...
+                  <p className="text-sm">{scopeDescriptions[rm]}</p>
+                </div>
+              );
+            })}
+          </div>
+        </>
+      )}
+      {capabilities.scopes.includes(NIP_47_PAY_INVOICE_METHOD) &&
+        permissions.scopes.has(NIP_47_PAY_INVOICE_METHOD) &&
+        (!isBudgetAmountEditable ? (
```
im-adithya, line 144: If budget amount is set and budget renewal isn't (in the params) then it assumes renewal as "Monthly" and the user can't change it. Avoiding option to edit budget renewal in these cases as that adds unnecessary complexity.

frontend/src/components/Permissions.tsx
```diff
@@ -1,274 +1,235 @@
...
+    isNewConnection ? !initialPermissions.scopes.size : canEditPermissions
+  );
+  const [isBudgetAmountEditable, setBudgetAmountEditable] = React.useState(
+    isNewConnection
+      ? Number.isNaN(initialPermissions.maxAmount)
+      : canEditPermissions
+  );
+  const [isExpiryEditable, setExpiryEditable] = React.useState(
+    isNewConnection ? !initialPermissions.expiresAt : canEditPermissions
+  );
```
im-adithya, line 48: Current flow (given nothing is passed in parameters)
permissions are set in NewApp as

`scopes: []`
`maxAmount:NaN` (as `AppPermissions` type expects a number)
`expiresAt: undefined` (thankfully it expects `Date | undefined`)

And hence we use these checks to identify if the params had any permission data

frontend/src/components/Permissions.tsx
```diff
@@ -1,274 +1,235 @@
...
-    } else {
-      newScopes.add(scope);
-      if (scope === "pay_invoice") {
-        budgetRenewal = "monthly";
-      }
-    }
+  const handlePermissionsChange = React.useCallback(
+    (changedPermissions: Partial<AppPermissions>) => {
+      const updatedPermissions = { ...permissions, ...changedPermissions };
+      setPermissions(updatedPermissions);
```
im-adithya, line 78: Setting local permissions separately as depending on initialPermissions might cause issues due to param data being changed. (Param data is only set from permissions and not initialPermissions so it doesn't change when it is updated, and the isEditable states are not disturbed post that in NewApp screen)

frontend/src/screens/apps/NewApp.tsx
```diff
@@ -171,12 +180,17 @@ const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
...
+      return;
+    }
+
     try {
       const createAppRequest: CreateAppRequest = {
         name: appName,
         pubkey,
         budgetRenewal: permissions.budgetRenewal,
-        maxAmount: permissions.maxAmount,
+        maxAmount: permissions.maxAmount || 0,
```
im-adithya, line 193: ^if user doesn't change and it stays in `NaN`
This whole `NaN` thing feels hacky, but changing `maxAmount` to `number | undefined` would need a lot of changes

frontend/src/components/Permissions.tsx
rolznz: Using NaN seems a bit strange, I'll take a look and see if there's another way

frontend/src/components/Scopes.tsx
rolznz: note: the variables you used are not even scopes (as you see `_METHOD` etc)

frontend/src/components/Permissions.tsx
rolznz: Yeah, I believe default should be never. Budgets from a deeplink-created app are different than the default user flow, and monthly in this case does not make sense

frontend/src/components/Scopes.tsx
```diff
@@ -0,0 +1,168 @@
...
+        ].includes(method)
+      )
+    );
+  }, [capabilities.scopes]);
+
+  const [scopeGroup, setScopeGroup] = React.useState<ScopeGroup>(() => {
+    if (!scopes.size || isSetEqual(scopes, fullAccessScopes)) {
```
rolznz, line 59: We don't need to do this, if an app passes in any request methods it can just be the custom option

frontend/src/components/Scopes.tsx
```diff
@@ -0,0 +1,168 @@
...
+      return SCOPE_GROUP_FULL_ACCESS;
+    } else if (isSetEqual(scopes, readOnlyScopes)) {
+      return SCOPE_GROUP_READ_ONLY;
+    }
+    return SCOPE_GROUP_CUSTOM;
+  });
+
+  // we need scopes to be empty till this point for isScopesEditable
```
rolznz, line 67: this comment sounds like a hack

same with `// stop setting scopes on re-renders`

frontend/src/types.ts
```diff
@@ -20,6 +23,14 @@ export const NIP_47_SIGN_MESSAGE_METHOD = "sign_message";
 
 export const NIP_47_NOTIFICATIONS_PERMISSION = "notifications";
 
+export const NIP_47_PAY_KEYSEND_METHOD = "pay_keysend";
+export const NIP_47_MULTI_PAY_KEYSEND_METHOD = "multi_pay_keysend";
+export const NIP_47_MULTI_PAY_INVOICE_METHOD = "multi_pay_invoice";
+
+export const SCOPE_GROUP_FULL_ACCESS = "full_access";
```
rolznz, line 30: We should try to only put types here which are used in multiple files.

Also, there is a mix of these constant variables, and string unions. We should be consistent here

frontend/src/screens/apps/NewApp.tsx
```diff
@@ -110,34 +117,34 @@ const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
...
-      requestMethodsSet.has("pay_invoice") ||
-      requestMethodsSet.has("multi_pay_invoice") ||
-      requestMethodsSet.has("multi_pay_keysend")
+      requestMethodsSet.has(NIP_47_PAY_KEYSEND_METHOD) ||
+      requestMethodsSet.has(NIP_47_PAY_INVOICE_METHOD) ||
+      requestMethodsSet.has(NIP_47_MULTI_PAY_INVOICE_METHOD) ||
+      requestMethodsSet.has(NIP_47_MULTI_PAY_KEYSEND_METHOD)
     ) {
-      scopes.add("pay_invoice");
+      scopes.add(NIP_47_PAY_INVOICE_METHOD);
```
rolznz, line 125: Scopes are not the same as methods.

rolznz:
@im-adithya there are a lot of issues here still and some things that are fundamentally wrong, and this will take too long to fix if I keep adding review comments like this. We also need to ensure this works with the new isolated app type, so I have started a new branch off of this one, started cleaning things up, and have also merged the dynamic budgets PR into it. I will be continuing there: https://github.com/getAlby/hub/pull/273

Please read through my commits to get an idea of what I have changed and let me know if you have any questions.

PR #223, opened 2024-07-08
feat: fire events on failed sync

rolznz:
I think we need better monitoring of core parts of our system. If syncs fail we need to quickly be notified and check our electrs service, or reach out to users who may be unable to sync for whatever other reason and provide as much help as we can.


bumi:
is there anything else that we can do? what does the user then need to do? restart?

rolznz:
> is there anything else that we can do? what does the user then need to do? restart?
If it's a temporary issue on startup (e.g. they run Alby Hub locally and have a bad internet connection) then yes, they can just try again.

If it fails consistently though or for multiple users, it would be a good indicator there is something wrong on our side that we need to look into.

Issue #122, opened 2024-05-08
Improve budget calculation


rolznz:
- Implement https://github.com/getAlby/hub/issues/123
- Implement transactions around payment creation and budget checking to ensure multiple payments cannot be created that exceed the budget

Issue #52, opened 2024-06-20
Create a common interface for NIP-47 controller methods


rolznz:
It is a bit more difficult to work with the NIP-47 controllers as they are right now (all different classes with different constructors)

See https://github.com/getAlby/nostr-wallet-connect-next/pull/461

Issue #25, opened 2024-07-01
Dynamic Budgets


rolznz:
TODOs
- [x] Transactions table with payment status
- [x] Update API, NIP-47 handlers, oauth handler to use transactions service
- [x] UI / planning for creating a dynamic budget vs normal NWC connection
- [x] keysend payments
- [x] Update payment logic to not allow budgets to be exceeded
- [x] migration of old payments - not necessary
- [x] Planning for isolated vs global view of data - e.g. for Alby Mobile
- [x] transaction list - show which app the transaction was for
- [x] LDK payment events rather than manually checking
- [x] internal payments
- [x] LND notifications support

Follow-ups
- [ ] onboard your friend/family flow with Dynamic Budgets and Alby Mobile
- [ ] undo LDK database changes in Alby fork (probably by undoing changes and then deleting old payments in a migration as they will be unused after we have our own payment DB)
- [ ] Blog post

Related tickets
https://github.com/getAlby/hub/issues/122

Issue #105, opened 2024-05-16
Expand what data is saved in NWC database


rolznz:
Do some planning around:

- What are our current limitations/state and where are we trying to get to (better notifications for users, better app connection budgets, improved performance, less app complexity)
- Transactions - currently we only store payments, not invoices. and no metadata, status, fees
  - Do we want to have one table for both payments and invoices? e.g. "Transactions" table as that's what it's called in NIP-47
  - improved budget requirements (fees, payment state)
  - custom data (LUD-12, LUD-18) and custom TLVs (Podcasting 2.0)
  - timing info (created, updated)
  - bolt11 invoice
  - payment hash (should be unique)
  - correct indexes and constraints
- Can we undo ALL our changes to the LDK-node payment table - this is difficult to maintain and will cause migration issues
- Channels - currently we only have access to open channels through LDK - closed ones disappear
  - How to fetch the data and when
  - Additional state - like if it closed, why
  - What to do with existing channels
- ~~DB size issues (my nwc.db is 400+mb already)~~

PR #226, opened 2024-07-08
Feat: dynamic budgets

rolznz:
TLDR:
- New "Isolated" app type which has its own balance and cannot interact with the rest of the wallet
- Budgets are now secure, with transactions around payments, unsettled payments, fees and fee reserves are now taken into account
- This is possible with a new internal transactions table - now all outgoing and incoming payments are tied to apps

Fixes https://github.com/getAlby/hub/issues/25
Fixes https://github.com/getAlby/hub/issues/52
Fixes https://github.com/getAlby/hub/issues/122
Partially addresses https://github.com/getAlby/hub/issues/284 (no multi-connection setup though)
Partially addresses https://github.com/getAlby/hub/issues/105

# Breaking Changes
- `Payments` table is deleted as part of the migration.
- Now we use internal transactions. Transactions from the individual LNClients are NOT migrated.

# What this enables
- [x] more transaction list metadata - which app made or received which payment
</img>
- [x] removes broken constraint between request events and payments (see https://github.com/getAlby/nostr-wallet-connect-next/pull/388)
- [x] fee and transaction states are saved internally, so improved budgeting and database transactions around payments are possible. See https://github.com/getAlby/hub/issues/122
- [x] Self payments 

# Changes
## Architecture
- New Transactions Layer (layer between api/nip-47 handlers and LN clients)
  - Creating a transaction will internally call the LNClient to create the transaction, then save it to our DB with additional metadata, such as the app the transaction was made for.
  - The transactions service is a consumer of our events system, so it receives notifications when transactions are received or sent transactions are settled. We then update our internal transactions.
  - Self-payments can now easily be intercepted by checking the payee of the invoice in the transactions service
  - for backends that do not receive events for updates, we check recently created unsettled transactions when lookup_invoice or list_transactions is called (Note: this will not work for receiving keysend payments, so the LND backend needs to implement notifications)
- LNClient
  - LDK `nwc_payment_received` and `nwc_payment_sent` has been updated to send entire payment details to the event publisher, which the transactions service will consume.
## Database
- New transaction table with additional data than the old payment table, notably:
  - `type` incoming or outgoing
  - `state` which can be used to ensure budgets are not exceeded
  - `metadata` JSON-encoded transaction metadata
- New properties on app permissions
  - balance_type (`full` or `isolated`) - isolated balance is summed from that app's received and sent payments
    - [x] move to apps instead of app_permissions
  - visibility (`full` or `isolated`) - isolated only allows accessing that app's created invoices
    - [x] move to apps instead of app_permissions
    - [x] merge balance_type and visibility into a single property. can just be an `isolated` boolean
- rename app_permissions max amount to be clearly in sats
  
TODOs:
- [x] update all transaction related usages of lnclient to go through transactions service
- [x] Test payment timeouts correctly apply to transactions (2 usages)
- [x] Ensure not found error propagates (see NIP-47 NOT_FOUND)
- [x] Add fees to budget calculation
- [x] Keysend (sends)
- [x] Keysend (receives)
- [x] Test payment metadata (boostagrams)
- [x] LDK event firing -> update transaction service
  - [x] received payment
    - [x] keysends
    - [x] BOLT11
  - [x] payment sent
    - [x] keysend
    - [x] BOLT11
  - [x] payment failed
    - [x] keysend
    - [x] BOLT11
- [x] Intercept self payments
- [x] Review unique constraints
  - [x] Duplicate outgoing payments for the same payment hash are possible if a payment fails and we retry - now sorted by settled des, then created desc
- [x] Should the internal wallet have it's own app ID - not needed
- [x] NIP-47 notifier receives a notification of a payment, but we do not know the app ID - test both for apps and global view
  - [x] notifier should be updated to check if the transaction app ID matches the app ID, or it's an app with a global view
  - [x] make sure notifier gets updated transaction (it must be consumed by transactions service first)
- [x] remove old payments table
- [x] review it's ok NOT to migrate transaction history. I would email all early users and let them know.
- [x] review keysend transactions initially do not have a payment hash, as it's generated by the LNClient - this is ok except for Greenlight/Breez, which are not a priority right now
- [x] review newly added app_permissions properties - move to app, internal boolean
- [x] To confirm: either implement notifications support in LND using `SubscribeInvoices` and `SubscribeTransactions` RPC methods and ensure it works for keysend, OR change the method of checking unsettled payments to list all payments from LNClient for a certain period of time. -> implement notifications support in LND
- [x] nwc_payment_failed event handling
- [x] fix lookup invoice returning not found
- [x] isolated balance check for keysend
- [x] remove duplication in isolated balance code
- [x] budget checks inside transaction
- [x] exceeded balance - add a proper error type so correct code is passed in NIP-47 response
- [x] exceeded budget - add a proper error type so correct code is passed in NIP-47 response
- [x] apply fee reserves in budget/balance checks
- [x] fee reserves for unsettled transactions
- [x] remove amount checks in NIP-47 handlers (now checked by transactions service) - PermissionsService.HasPermission should not check amounts, this is done by the transactions service in a database transaction
- [x] DB Transactions around payments
- [x] Add fee reserve to budget calculation for unsettled payments
- [x] Change code names of satoshi fields in DB models (using column rename)
- [ ] review  fee_reserve implementation for unsettled payments (column on transactions DB)
- [x] isolated app should not have access to node info
- [x] isolated app should not be able to sign messages -> merge the permissions revamp
- [x] add AUTOINCREMENT on the transactions table
- [x] add AUTOINCREMENT on the existing tables, drop request and response events
- [x] unknown external sent payments should not throw an error
- [x] fix autoincrement migration, add `DELETE FROM app_permissions WHERE app_id NOT IN (SELECT id FROM apps);`
- [x] review extra SQLite pragma commands

UI
- [x] transaction list
- [x] app cards with isolated balance
- [x] how to create/edit apps to be isolated? how does it affect the rest of the UI?

Unit tests
- [x] get_balance controller for isolated balance
- [x] event handler tests
- [x] update nip-47 controller tests
- [x] transactions service tests

Testing
- [x] NIP-47 notifications
- [x] Non-LDK LNClient (e.g. Cashu)
- [x] LND async payments
- [x] LND self-payments (do they expose the preimage? otherwise we can generate it, same as LNDHub)

To move to new issues:
- [ ] Enable Roman's cleanup task for old request and response events
- [ ] isolated app possible follow-ups (in case we wish to spend more time on isolated apps)
  - internal payments should trigger NIP-47 notifications (this can be done by the transactions service publishing its own notifications rather than the notifier subscribing to LNClient notifications - this would simplify things too)
  - restrict isolated balance to not receive more than a certain budget (it's hard to do?)
  - prevent hub not spending balance that is set aside for other apps
  - prevent paying more than actual balance from main wallet into isolated wallets (you can send an infinite amount)
  - cashu isolated app payments does not work and causes money to be "lost"
- [ ] currently there are no DB indexes. I am not sure how this will perform when the number of payments grow
- [ ] ListTransactions on LNClients that support notifications should actually just throw an error (they should never be used)
- [ ] fee reserve calculation could be per-LNClient to completely match what the node sets
- [ ] Extra nip-47 controller tests
  - [ ] multi_pay_keysend (partial success)
  - [ ] list_transactions (paging params)


transactions/transactions_service.go
```diff
@@ -0,0 +1,690 @@
...
+			return errors.New("failed to cast event")
+		}
+
+		var dbTransaction db.Transaction
+		result := svc.db.Find(&dbTransaction, &db.Transaction{
+			Type:        constants.TRANSACTION_TYPE_OUTGOING,
+			PaymentHash: lnClientTransaction.PaymentHash,
+		})
+
+		if result.RowsAffected == 0 {
```
im-adithya, line 555: What if an LND user did a payment externally? Then #281 would send a `nwc_payment_sent ` event to return an error here

rolznz: fixed

lnclient/lnd/lnd.go
```diff
@@ -380,46 +378,46 @@ func (svc *LNDService) SendKeysend(ctx context.Context, amount uint64, destinati
...
 			"customRecords": custom_records,
 			"paymentError":  resp.PaymentError,
 		}).Errorf("Keysend payment has payment error")
-		return "", errors.New(resp.PaymentError)
+		return paymentHash, "", 0, errors.New(resp.PaymentError)
 	}
-	respPreimage = hex.EncodeToString(resp.PaymentPreimage)
-	if respPreimage == "" {
+	respPreimage := hex.EncodeToString(resp.PaymentPreimage)
+	if respPreimage == preimage {
```
im-adithya, line 400: This should be `!=`

im-adithya: Fixed in #281

db/migrations/202407012100_transactions.go
```diff
@@ -0,0 +1,59 @@
...
+	description_hash text,
+	amount_msat integer,
+	fee_msat integer,
+	fee_reserve_msat integer,
+	created_at datetime,
+	updated_at datetime,
+	expires_at datetime,
+	settled_at datetime,
+	metadata text,
+	self_payment boolean
```
bumi, line 40: we probably should have an index on 
+ app_id
+ state
+ type
+ payment_hash
+ created_at / settled_at? 

need to check what queries we make, but I think we need something here. 
for example for this a combined might be helpful

`"app_id = ? AND type = ? AND (state = ? OR state = ?) AND created_at > ?"`
`app_id = ? AND type = ? AND state = ?`


make sure `type` is no reserved word.

db/migrations/202407012100_transactions.go
```diff
@@ -0,0 +1,59 @@
...
+// - Adds new properties to apps
+//   - isolated boolean
+//
+// - Renames max amount on app permissions to be clear its in sats
+var _202407012100_transactions = &gormigrate.Migration{
+	ID: "202407012100_transactions",
+	Migrate: func(tx *gorm.DB) error {
+
+		if err := tx.Exec(`
+CREATE TABLE transactions(
```
bumi, line 21: transactions is also often a reserved word - esp in the DB context this can lead to strange things. are we good there?

db/queries/get_budget_usage.go
```diff
@@ -0,0 +1,44 @@
...
+	"gorm.io/gorm"
+)
+
+func GetBudgetUsageSat(tx *gorm.DB, appPermission *db.AppPermission) uint64 {
+	var result struct {
+		Sum uint64
+	}
+	tx.
+		Table("transactions").
+		Select("SUM(amount_msat + coalesce(fee_msat, 0) + coalesce(fee_reserve_msat, 0)) as sum").
```
bumi, line 17: why do we need the `coalesce` ? isn't a fee_msat default of 0 ok?

db/migrations/202407012100_transactions.go
rolznz: it works, but I also don't like the same terminology being used for 2 different things. I added a note to https://github.com/getAlby/hub/issues/299

db/queries/get_budget_usage.go
rolznz: Currently fee_msat is nullable, because we don't always know the fee (for some connectors) and it seems wrong for it to be zero (I think we chatted about this in the past)

db/migrations/202407012100_transactions.go
bumi: why would you want to wait? what is an argument of not adding an index?
you know which queries you do and I am not sure if we can identify slow queries.

db/queries/get_budget_usage.go
bumi: I'd actually add zero in our wallet - imo such exceptions because some connectors will make the app complicated and now we have to do things like `coalesce` because some connector misses a feature. 
and in the next code this `coalesce` or similar will be forgotten and just causes bugs.

PR #231, opened 2024-07-09
feat: changebg color to black for all themes for intro screens

pavanjoshi914:


frontend/src/themes/alby.css
```diff
@@ -1,5 +1,6 @@
 .theme-alby {
   --background: 0 0% 100%;
+  --intro-background: 240 10% 3.9%;
```
rolznz, line 3: why do we need a new variable?

and best to wait for @stackingsaunter to confirm what we should do

PR #232, opened 2024-07-09
fix: default redirect on settings route

pavanjoshi914:
properly tested


frontend/src/routes.tsx
```diff
@@ -102,7 +101,7 @@ const routes = [
       },
       {
         path: "settings",
-        element: <SettingsLayout />,
```
rolznz, line 105: Can you see anything wrong in the UI by doing this?

rolznz:
@pavanjoshi914 can you fix the issue rather than just closing it?

frontend/src/routes.tsx
pavanjoshi914: should be fixed now

reneaaron (approved):
tACK

Issue #230, opened 2024-07-08
Bug: LNClients expect pay_keysend tlv values as plaintext instead of hex


rolznz:
As per https://github.com/nostr-protocol/nips/blob/master/47.md#pay_keysend the values should be hex encoded.

Also see https://github.com/getAlby/js-sdk/pull/229

PR #234, opened 2024-07-09
fix: correctly pass TLVs in NIP-47 pay_keysend

rolznz:
as per the nip-47 spec, pay_keysend TLV values are hex-encoded

Fixes https://github.com/getAlby/hub/issues/230

### DIFF
--- a/lnclient/greenlight/greenlight.go
+++ b/lnclient/greenlight/greenlight.go
@@ -132,7 +132,7 @@ func (gs *GreenlightService) SendKeysend(ctx context.Context, amount uint64, des
 	for _, customRecord := range custom_records {
 		extraTlvs = append(extraTlvs, glalby.TlvEntry{
 			Ty:    customRecord.Type,
-			Value: customRecord.Value,
+			Value: customRecord.Value, // glalby expects hex-encoded TLV values
 		})
 	}
 
--- a/lnclient/greenlight/greenlight.go
+++ b/lnclient/greenlight/greenlight.go
@@ -2,7 +2,6 @@ package greenlight
 
 import (
 	"context"
-	"encoding/hex"
 	"errors"
 	"log"
 	"math/rand"
@@ -133,7 +132,7 @@ func (gs *GreenlightService) SendKeysend(ctx context.Context, amount uint64, des
 	for _, customRecord := range custom_records {
 		extraTlvs = append(extraTlvs, glalby.TlvEntry{
 			Ty:    customRecord.Type,
-			Value: hex.EncodeToString([]byte(customRecord.Value)),
+			Value: customRecord.Value,
 		})
 	}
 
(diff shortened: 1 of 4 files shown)


lnclient/greenlight/greenlight.go
```diff
@@ -133,7 +132,7 @@ func (gs *GreenlightService) SendKeysend(ctx context.Context, amount uint64, des
 	for _, customRecord := range custom_records {
 		extraTlvs = append(extraTlvs, glalby.TlvEntry{
 			Ty:    customRecord.Type,
-			Value: hex.EncodeToString([]byte(customRecord.Value)),
+			Value: customRecord.Value,
```
im-adithya, line 135: Does greenlight expect encoded value? Can you add a comment so it's understandable

lnclient/lnd/lnd.go
```diff
@@ -359,7 +359,11 @@ func (svc *LNDService) SendKeysend(ctx context.Context, amount uint64, destinati
 
 	destCustomRecords := map[uint64][]byte{}
 	for _, record := range custom_records {
-		destCustomRecords[record.Type] = []byte(record.Value)
+		decodedValue, err := hex.DecodeString(record.Value)
+		if err != nil {
+			return "", err
+		}
+		destCustomRecords[record.Type] = decodedValue
```
im-adithya, line 366: LND actually accepts keysend record as a hex encoded value (see [this](https://docs.lightning.engineering/lightning-network-tools/lnd/send-messages-with-keysend#send-a-message-to-other-nodes))

rolznz: @im-adithya you linked to the cli docs. The LND DestCustomRecords is a `map[uint64][]byte`

im-adithya: Yup, I tried it from the CLI in polar, and what we have right now does it fine. Because after passing the hex encoded in the cli, it still returned the same thing while listing transactions (the hex encoded value `48617070792047656E6573697320426C6F636B2044617921 ` but not the actual string `Happy Genesis Block Day!`)

rolznz: @im-adithya I don't understand, why are you testing the cli?

lnclient/greenlight/greenlight.go
rolznz, line 135:
```suggestion
			Value: customRecord.Value, // glalby expects hex-encoded TLV values
```

lnclient/lnd/lnd.go
im-adithya: Discard above, it accepts a hex encoded value in bytes and the above line does exactly that. tACK ✅

im-adithya (approved):

lnclient/lnd/lnd.go
rolznz: Just to confirm - it accepts the value as *bytes* not as hex-encoded bytes - that's why the line above decodes the hex first

im-adithya: Yup got that, thanks!

PR #235, opened 2024-07-09
feat: appcard improvements & linking account

reneaaron:


alby/alby_oauth_service.go
```diff
@@ -262,7 +261,7 @@ func (svc *albyOAuthService) DrainSharedWallet(ctx context.Context, lnClient lnc
 
 	amountSat := int64(math.Floor(
 		balanceSat- // Alby shared node balance in sats
-			(balanceSat*(8/1000))- // Alby service fee (0.8%)
+			(balanceSat*(8.0/1000.0))- // Alby service fee (0.8%)
```
reneaaron, line 264: @rolznz I think that could be another reason why the migration doesn't work currently. Without the `.` the calculation result is an integer, which will always be 0 (so the full balance will be deducted.

rolznz: @reneaaron nice catch!

frontend/src/components/Permissions.tsx
```diff
@@ -1,26 +1,19 @@
 import { PlusCircle } from "lucide-react";
 import React, { useEffect, useState } from "react";
+import BudgetAmountSelect from "src/components/BudgetAmountSelect";
+import BudgetRenewalSelect from "src/components/BudgetRenewalSelect";
```
im-adithya, line 4: Can we leave the refactoring here for now? This might bring conflicts in https://github.com/getAlby/hub/pull/14 🥲

reneaaron: I don't think that makes sense, no. It's literally just encapsulating markup and, no changes to state variables and such.

frontend/src/components/connections/AlbyConnectionCard.tsx
```diff
@@ -62,10 +80,43 @@ function AlbyConnectionCard({ connection }: { connection?: App }) {
...
+                  <DialogTrigger asChild>
+                    <LoadingButton loading={loading}>
+                      {!loading && <Link2Icon className="w-4 h-4 mr-2" />}
+                      Link your Alby Account
+                    </LoadingButton>
+                  </DialogTrigger>
+                  <DialogContent>
+                    <DialogHeader>Link to Alby Account</DialogHeader>
+                    <DialogDescription className="flex flex-col gap-4">
+                      After you link your account, every app you access through
```
rolznz, line 93: Can we also include info about the lightning address?

rolznz:
image

@reneaaron  @stackingsaunter can we adjust the illustration a bit? I think if it had links to a lightning address, a podcasting app, etc. it would be much easier to understand.

Right now it's just a button with a cursor on it

alby/models.go
```diff
@@ -29,6 +29,11 @@ type AlbyPayRequest struct {
 	Invoice string `json:"invoice"`
 }
 
+type AlbyLinkAccountRequest struct {
+	Budget  int    `json:"budget"`
```
rolznz, line 33: if you make this a `uint64` you don't need to cast it elsewhere

reneaaron:
> image
> 
> @reneaaron @stackingsaunter can we adjust the illustration a bit? I think if it had links to a lightning address, a podcasting app, etc. it would be much easier to understand.
> 
> Right now it's just a button with a cursor on it

How about this one? 

image

rolznz:
image

Weird "dai" instead of "daily"

rolznz (approved):
tACK

bumi:
<3

stackingsaunter:
Updated illustration @reneaaron 

alby-account-mobile-hub
alby-account-mobile-hub

PR #242, opened 2024-07-10
fix: make copy to clipboard work in non-https environments

### LINKED ISSUE #56: Bug: "Copy Invoice" uncaught exception in "Buy an Instant Channel"
Please see screenshot below.

I would like to try to fix this issue, however I could use a little bit of direction here. 

Error is happening here:

https://github.com/getAlby/bitcoin-connect/blob/b1306775814135ff7fe14e0e4758d60286321cf1/src/components/pages/bc-send-payment.ts#L213

Would you suggest that I dig into the `bitcoin-connect` package to try to debug this issue? Or are we using `bitcoin-connect` incorrectly?

copy-invoice-failure

reneaaron:
Fixes #56 

Tested on a local domain.

### DIFF
--- a/frontend/src/lib/clipboard.ts
+++ b/frontend/src/lib/clipboard.ts
@@ -11,12 +11,15 @@ export async function copyToClipboard(content: string) {
       textArea.style.position = "absolute";
       textArea.style.opacity = "0";
       document.body.appendChild(textArea);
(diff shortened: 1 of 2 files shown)


im-adithya (approved):
tACK nice fix, thanks 🙌

PR #244, opened 2024-07-10
fix: validate theme from storage

### LINKED ISSUE #221: Clouds not showing in onboarding any more (only when system appearance = light mode)
to reproduce, delete your cookies/local storage and set system appearance to lightmode. Then go through the onboarding.

reneaaron:
Partly fixes #221

Older hubs had "dark" saved as the ui theme which now was rename to a explicit darkmode setting, however there is no theme named "dark". Thus the theme appears broken if you upgrade an older version of the hub, for new installs it should be fine.


rolznz (approved):
utACK

Issue #20, opened 2024-07-04
"Open First Channel" goes to wrong place if Alby Funds Migration is possible


rolznz:
Also, remove the "View other options" link

PR #248, opened 2024-07-10
fix: link to open first channel when alby shared funds are available

rolznz:
Fixes https://github.com/getAlby/hub/issues/20


bumi:
lgtm

Issue #236, opened 2024-07-09
Re-add primary "Open Channel" button on node page


rolznz:
Issue: there is no primary action on the node page and users don't easily see where to go first

Which goes to the main (outgoing) channel flow, but we can add a button to link to the receiving flow

And change wording of "Topup" for increasing spending balance

![image](https://github.com/getAlby/hub/assets/33993199/f9c2a148-297f-4ca0-b624-3e74c5dd4eed)

rolznz:
Extra screenshots of confusion - we also need to clearly explain whether it is bitcoin or lightning on the pages:
![image](https://github.com/getAlby/hub/assets/33993199/c8132fff-8afa-4d68-8444-2fb697291ded)
![image](https://github.com/getAlby/hub/assets/33993199/41133c6e-eaff-4971-ae76-57215fb3c28a)

PR #249, opened 2024-07-10
feat: improve open channel ux flows

rolznz:
Fixes https://github.com/getAlby/hub/issues/236

- re-add primary open channel button
- add links to increase receiving capacity from outgoing and channel order pages so the user can always find it

image
image
image


bumi (approved):

reneaaron (approved):
utACK

PR #254, opened 2024-07-11
feat: new illustration for linking account

reneaaron:
@rolznz You we're a bit too fast. :grimacing:


rolznz (approved):
Oops, thanks!

utACK

PR #255, opened 2024-07-11
chore: remove unnecessary dark classes

im-adithya:
title

### DIFF
--- a/frontend/src/components/TransactionItem.tsx
+++ b/frontend/src/components/TransactionItem.tsx
@@ -67,14 +67,14 @@ function TransactionItem({ tx }: Props) {
                 className={cn(
                   "w-6 h-6 md:w-8 md:h-8",
                   type === "outgoing"
-                    ? "text-orange-400 dark:text-amber-600 stroke-orange-400 dark:stroke-amber-600"
-                    : "text-green-500 dark:text-emerald-500 stroke-green-400 dark:stroke-emerald-500"
+                    ? "stroke-orange-400 dark:stroke-amber-600"
+                    : "stroke-green-400 dark:stroke-emerald-500"
                 )}
               />
             </div>
           </div>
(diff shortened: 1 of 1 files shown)


frontend/src/components/TransactionItem.tsx
```diff
@@ -67,14 +67,14 @@ function TransactionItem({ tx }: Props) {
                 className={cn(
                   "w-6 h-6 md:w-8 md:h-8",
                   type === "outgoing"
-                    ? "text-orange-400 dark:text-amber-600 stroke-orange-400 dark:stroke-amber-600"
-                    : "text-green-500 dark:text-emerald-500 stroke-green-400 dark:stroke-emerald-500"
+                    ? "stroke-orange-400 dark:stroke-amber-600"
```
rolznz, line 70: I don't know what happened, I thought I already commented this but I think we should also not hardcode sent and received colors. Is there a good way we can do this? CC @reneaaron

reneaaron: We could of course introduce new variables, but since it's the only placed used right now I would be good to keep it as is. 

As soon as we need those colors in other places as well I'd probably introduce separate theme variables for that purpose. Does that make sense for you?

rolznz: @reneaaron sounds good!

im-adithya: 👍 and @rolznz I opened a similar PR to clean up wrong CSS classnames in the extension repo, you have commented [there](https://github.com/getAlby/lightning-browser-extension/pull/3190#pullrequestreview-2173621760) 😄

rolznz: :man_facepalming: thanks @im-adithya !

rolznz (approved):
utACK

PR #257, opened 2024-07-11
fix: add unique id to app created page

### LINKED ISSUE #252: Creating a new app instantly navigates
There seems to be a bug sometimes creating a second app connection after the first was successfully setup (and received an event). It looks like a bug with the old app pubkey being cached somewhere, causing an instant navigation on the `/created` page.

https://github.com/getAlby/hub/assets/33993199/5825c21b-5402-4fbe-b269-6afaf0e7744e

A temporary workaround is to do a hard refresh before creating a new app connection

reneaaron:
Fixes #252 

State seems to be persisted in browsers history, we fix that by using a unique URI for every app that has been created.

### DIFF
--- a/frontend/src/screens/apps/NewApp.tsx
+++ b/frontend/src/screens/apps/NewApp.tsx
@@ -200,9 +200,13 @@ const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
         window.location.href = createAppResponse.returnTo;
         return;
       }
-      navigate(`/apps/created${app ? `?app=${app.id}` : ""}`, {
-        state: createAppResponse,
-      });
+      // Adding a unique id prevents https://github.com/getAlby/hub/issues/252
+      navigate(
+        `/apps/created?id=${createAppResponse.pairingPublicKey}${app ? `&app=${app.id}` : ""}`,
+        {
+          state: createAppResponse,
+        }
+      );
       toast({ title: "App created" });
     } catch (error) {
       handleRequestError(toast, "Failed to create app", error);


im-adithya (approved):
Nice 🙌

im-adithya:
Did you try passing `{ state: null }` after we go to /apps on having lastEventAt [here](https://github.com/getAlby/hub/blob/master/frontend/src/screens/apps/AppCreated.tsx#L56)? I can't reproduce even the actual issue

frontend/src/screens/apps/NewApp.tsx
```diff
@@ -200,9 +200,13 @@ const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
         window.location.href = createAppResponse.returnTo;
         return;
       }
-      navigate(`/apps/created${app ? `?app=${app.id}` : ""}`, {
-        state: createAppResponse,
-      });
+      // Adding a unique id prevents https://github.com/getAlby/hub/issues/252
+      navigate(
+        `/apps/created?id=${createAppResponse.pairingPublicKey}${app ? `&app=${app.id}` : ""}`,
```
rolznz, line 205: It was already here, but why is there a conditional here? this seems wrong?

Also, the `pairingPublicKey` should be unique per app

rolznz (changes requested):
something is wrong here (not sure what)

frontend/src/screens/apps/NewApp.tsx
rolznz: sorry, I completely misread this. I am wrong :cry:

rolznz: but it's weird, the navigation state should work. And it seems like it's not

rolznz:
@im-adithya @reneaaron I found how to consistently reproduce the issue:

1. Create a new app and do at least one NWC event for it
2. Delete the app
3. Create a new app

Step 3. Will lead to this bug.

The pubkey is different for each app, but SQLITE is reusing app IDs rather than properly auto-incrementing.

We fetch events by pubkey from the API, but internally we query by app ID.

And for some reason when we delete apps, the request_events are not deleted.

So this PR does actually not fix the issue, and there's nothing wrong with the routing (which passes different state each time)

rolznz:
For some reason the foreign key on delete constraint is not working

```
sqlite> .schema request_events
CREATE TABLE IF NOT EXISTS "request_events" (`id` integer,`app_id` integer null,`nostr_id` text UNIQUE,`state` text,`created_at` datetime,`updated_at` datetime, method TEXT, content_data TEXT,PRIMARY KEY (`id`),CONSTRAINT `fk_request_events_app` FOREIGN KEY (`app_id`) REFERENCES `apps`(`id`) ON DELETE CASCADE);
CREATE UNIQUE INDEX `idx_request_events_nostr_id` ON `request_events`(`nostr_id`);
CREATE INDEX `idx_request_events_app_id` ON `request_events`(`app_id`);
CREATE INDEX idx_request_events_app_id_and_id ON request_events(app_id, id);
CREATE INDEX `idx_request_events_method` ON `request_events`(`method`);
```

PR #260, opened 2024-07-11
fix: improvements for smaller screens

reneaaron:
- Make sidebar scrollable if not all content fits into screen height
 - 2 rows of buttons on wallet page
 - channels cards on node page in mobile view
 - wrap settings menu items on mobile


rolznz (approved):
tACK

PR #274, opened 2024-07-15
fix: migrate existing tables to have autoincrementing primary keys

rolznz:
Fixes https://github.com/getAlby/hub/issues/252

This PR adds autoincrement to the primary keys of all our tables which fixes weird bugs caused by ID reuse which is the default in SQLITE (if you delete app ID 7, then the next app you create will have ID 7 again). Only apps, app_permissions and user_configs are preserved. The other tables are dropped (payments are dropped as part of dynamic budgets, request_events and response_events are not critical)

TESTS:
- [x] diff select * from `table-name` before and after migration (no change)
- [x] diff schema changes (only expected changes - autoincrement, one missed index rename for scopes)
- [x] autoincrement
  - [x] apps
  - [x] app_permissions
  - [x] user_configs
  - [x] request_events
  - [x] response_events
  - [x] transactions
- [x] cascading delete
  - [x] apps -> app_permissions
  - [x] apps -> request_events
  - [x] request_events -> response_events
- [x] create app with one NWC request, delete app, create new app -> new ID and bug fixed?
- [x] check: should index creation of request and response events be done after new app table is created?

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -15,21 +15,19 @@ func NewDB(uri string) (*gorm.DB, error) {
 	if err != nil {
 		return nil, err
 	}
-	err = gormDB.Exec("PRAGMA foreign_keys=ON;").Error
+	err = gormDB.Exec("PRAGMA foreign_keys = ON", nil).Error
 	if err != nil {
 		return nil, err
 	}
-	err = gormDB.Exec("PRAGMA auto_vacuum=FULL;").Error
+	err = gormDB.Exec("PRAGMA auto_vacuum = FULL", nil).Error
 	if err != nil {
 		return nil, err
 	}
 
-	// sqlDb, err = DB.DB()
-	// if err != nil {
-	// 	return err
-	// }
-	// this causes errors when concurrently saving DB entries and otherwise requires mutexes
-	// sqlDb.SetMaxOpenConns(1)
+	err = gormDB.Exec("PRAGMA busy_timeout = 5000", nil).Error
+	if err != nil {
+		return nil, err
+	}
 
 	err = migrations.Migrate(gormDB)
 	if err != nil {
--- a/db/migrations/202407151352_autoincrement.go
+++ b/db/migrations/202407151352_autoincrement.go
@@ -58,7 +58,7 @@ CREATE INDEX idx_app_permissions_app_id ON app_permissions(app_id);
 
 			// create fresh request and response event tables
 			if err := tx.Exec(`
-CREATE TABLE "request_events" (id integer PRIMARY KEY AUTOINCREMENT,app_id integer null,nostr_id text UNIQUE,state text,created_at datetime,updated_at datetime, method TEXT, content_data TEXT,CONSTRAINT fk_request_events_app FOREIGN KEY (app_id) REFERENCES apps(id) ON DELETE CASCADE);
+CREATE TABLE "request_events" (id integer PRIMARY KEY AUTOINCREMENT,app_id integer,nostr_id text UNIQUE,state text,created_at datetime,updated_at datetime, method TEXT, content_data TEXT,CONSTRAINT fk_request_events_app FOREIGN KEY (app_id) REFERENCES apps(id) ON DELETE CASCADE);
 CREATE INDEX idx_request_events_app_id ON request_events(app_id);
 CREATE INDEX idx_request_events_app_id_and_id ON request_events(app_id, id);
 CREATE INDEX idx_request_events_method ON request_events(method);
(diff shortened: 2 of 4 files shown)


db/db.go
```diff
@@ -15,21 +15,19 @@ func NewDB(uri string) (*gorm.DB, error) {
...
 		return nil, err
 	}
 
-	// sqlDb, err = DB.DB()
-	// if err != nil {
-	// 	return err
-	// }
-	// this causes errors when concurrently saving DB entries and otherwise requires mutexes
-	// sqlDb.SetMaxOpenConns(1)
+	err = gormDB.Exec("PRAGMA busy_timeout = 5000", nil).Error
```
rolznz, line 27: Added so that we won't get a SQL busy error thrown if there are two concurrent DB queries

bumi: good idea!

db/migrations/202407151352_autoincrement.go
```diff
@@ -0,0 +1,80 @@
...
+// - Adds AUTOINCREMENT to the primary key of:
+// - apps, app_permissions, request_events, response_events, user_configs
+//
+// request_events and response_events are not critical (and also payments are dropped in the same release)
+// so we just drop those tables and re-create them.
+var _202407151352_autoincrement = &gormigrate.Migration{
+	ID: "202407151352_autoincrement",
+	Migrate: func(db *gorm.DB) error {
+
+		if err := db.Transaction(func(tx *gorm.DB) error {
```
bumi, line 20: gorm migrations are not by default in a transaction?

or should we split up this transaction into several ones for each table? seems a bit a lot for the one transaction

db/migrations/202407151352_autoincrement.go
```diff
@@ -0,0 +1,80 @@
...
+`).Error; err != nil {
+				return err
+			}
+
+			// User configs - create new table, copy old values, delete old table, rename new table
+			if err := tx.Exec(`
+CREATE TABLE "user_configs_2" ("id" integer PRIMARY KEY AUTOINCREMENT, "key" text NOT NULL UNIQUE, "value" text, "encrypted" numeric, created_at datetime,updated_at datetime);
+INSERT INTO user_configs_2 (id, key, value, encrypted, created_at, updated_at) SELECT id, key, value, encrypted, created_at, updated_at FROM user_configs;
+DROP TABLE user_configs;
+ALTER TABLE user_configs_2 RENAME TO user_configs;
```
bumi, line 35: wondering if we should do this in a second migration? those are the most valuable data. 
and those actually also don't have associations thus the auto increment is not that important.

db/migrations/202407151352_autoincrement.go
```diff
@@ -0,0 +1,80 @@
...
+`).Error; err != nil {
+				return err
+			}
+
+			// Apps & app permissions (interdependent)
+			// create new tables, copy old values, delete old tables, rename new tables, create new indexes
+			if err := tx.Exec(`
+CREATE TABLE apps_2 (id integer PRIMARY KEY AUTOINCREMENT,name text,description text,nostr_pubkey text UNIQUE,created_at datetime,updated_at datetime, isolated boolean);
+INSERT INTO apps_2 (id, name, description, nostr_pubkey, created_at, updated_at, isolated) SELECT id, name text, description, nostr_pubkey, created_at, updated_at, isolated FROM apps;
```
bumi, line 44: do we already have the `isolated` column? this is because it's not into master, and in the other branch we will have it?

db/migrations/202407151352_autoincrement.go
rolznz: @bumi the default is off - we can turn it on in the gorm settings

If we split this into multiple transactions, then if any of them fail we will have trouble re-running the migration (next time the migration will fail because the DB state is not expected).

I would instead split this into multiple migrations. What do you think?

db/migrations/202407151352_autoincrement.go
rolznz: @bumi yeah, I am wondering if I should just remove this from the migration and we just leave this table without autoincrement?

db/migrations/202407151352_autoincrement.go
rolznz: @bumi yes, this is branched off (and will be merged into) the dynamic budgets PR.

Also, in that PR, the transactions table correctly has autoincrement on its primary key

db/migrations/202407151352_autoincrement.go
rolznz: I actually don't think there is much data though - It's only the user configs and apps/app permissions. Everything else is cleared

db/migrations/202407151352_autoincrement.go
rolznz: Decision: let's not update this one

PR #280, opened 2024-07-15
Render.com phoneixd

bumi:
This makes finding the matching phoneixd in cloud environments easier. we default to http if that is not privided


rolznz (approved):
utACK

PR #281, opened 2024-07-16
feat: subscribe for payments and invoices

im-adithya:
Adds a goroutine to fetch invoices and payments on starting the node

Needs input:
- Do we need historic payments/invoices?
- The `nwc_payment_sent` event consumer currently blocks external payments, what do there?


lnclient/lnd/lnd.go
```diff
@@ -448,14 +415,89 @@ func NewLNDService(ctx context.Context, lndAddress, lndCertHex, lndMacaroonHex s
...
+		paymentStream, err := lndClient.SubscribePayments(lndCtx, &routerrpc.TrackPaymentsRequest{
+			NoInflightUpdates: true,
+		})
+		if err != nil {
+			logger.Logger.Errorf("Error subscribing to payments: %v", err)
+			return
+		}
+		for {
+			payment, err := paymentStream.Recv()
+			if lndCtx.Err() == context.Canceled {
```
rolznz, line 433: hmm, can you select on context.Done() instead?

lnclient/lnd/lnd.go
```diff
@@ -448,14 +415,89 @@ func NewLNDService(ctx context.Context, lndAddress, lndCertHex, lndMacaroonHex s
...
+
+	lndService := &LNDService{client: lndClient, pubkey: info.IdentityPubkey, cancel: cancel}
+
+	// Subscribe to payments
+	go func() {
+		paymentStream, err := lndClient.SubscribePayments(lndCtx, &routerrpc.TrackPaymentsRequest{
+			NoInflightUpdates: true,
+		})
+		if err != nil {
+			logger.Logger.Errorf("Error subscribing to payments: %v", err)
```
rolznz, line 428: instead of just returning (killing the gorouting completely) can we try again?

lnclient/lnd/lnd.go
im-adithya: I tried adding 
```go
select {
case <-lndCtx.Done():
	return
default:
	payment, err := paymentStream.Recv()
	if err != nil {
		logger.Logger.Errorf("Failed to receive payment: %v", err)
	...
```

But it still prints the error
`{"level":"error","msg":"Failed to receive payment: rpc error: code = Canceled desc = context canceled","time":"2024-07-17T15:55:25+05:30"}`
on context cancelling

lnclient/lnd/lnd.go
im-adithya: You mean wrap it in a for loop and keep trying if we result in an error?

rolznz: Yeah, otherwise the subscription will stop working. It's similar to what we do when we reconnect to the relay

lnclient/lnd/lnd.go
rolznz: Is it just an error log? there's no issue, right? I think we should still add this switch checking for ctx.Done.

im-adithya: Yeah I'll maybe add it, select is used like you said also in lndhub.go too ([here](https://github.com/getAlby/lndhub.go/blob/main/lib/service/invoicesubscription.go#L243-L248)) 🤔 But I'm wondering because the context done should trigger the select above it instead of resulting in this, I think it's because we do this in default and by the time context is killed it goes till the execution of `paymentStream.Recv()`

lnclient/lnd/lnd.go
im-adithya: Done!

lnclient/lnd/lnd.go
```diff
@@ -448,14 +415,97 @@ func NewLNDService(ctx context.Context, lndAddress, lndCertHex, lndMacaroonHex s
...
+			}
+		}
+	}()
+
+	// Subscribe to invoices
+	go func() {
+		for {
+			invoiceStream, err := lndClient.SubscribeInvoices(lndCtx, &lnrpc.InvoiceSubscription{})
+			if err != nil {
+				logger.Logger.Errorf("Error subscribing to invoices: %v", err)
```
rolznz, line 470: can you please use the JSON logging format - Logger.WithError(err).Error("message")

rolznz:
- Can we be notified if the user tries to make a payment and it fails sometime later?
- LND LNClient was not updated to signal it supports notifications

lnclient/lnd/lnd.go
im-adithya: Done!

lnclient/lnd/lnd.go
```diff
@@ -448,14 +415,104 @@ func NewLNDService(ctx context.Context, lndAddress, lndCertHex, lndMacaroonHex s
...
+			}
+			for {
+				select {
+				case <-lndCtx.Done():
+					return
+				default:
+					payment, err := paymentStream.Recv()
+					if err != nil {
+						logger.Logger.WithError(err).Error("Failed to receive payment")
+						time.Sleep(2 * time.Second)
```
rolznz, line 440: Why do we sleep here?

im-adithya: Just to wait if that changes anything, otherwise we might keep looping into errors for some time

lnclient/lnd/lnd.go
```diff
@@ -693,6 +748,47 @@ func (svc *LNDService) GetBalances(ctx context.Context) (*lnclient.BalancesRespo
...
+		Preimage:        payment.PaymentPreimage,
+		PaymentHash:     payment.PaymentHash,
+		Amount:          payment.ValueMsat,
+		FeesPaid:        payment.FeeMsat,
+		CreatedAt:       time.Unix(0, payment.CreationTimeNs).Unix(),
+		Description:     description,
+		DescriptionHash: descriptionHash,
+		ExpiresAt:       expiresAt,
+		SettledAt:       settledAt,
+		//TODO: Metadata:  (e.g. keysend),
```
rolznz, line 788: can this be done?

rolznz: Update: transactions service will save metadata for keysend payments

rolznz (approved):
utACK

PR #286, opened 2024-07-17
feat: add channel alerts in send and receive pages

### LINKED ISSUE #241: Add alerts on wallet send/receive pages if receiving/sending capacity is insufficient
Similar to the Node page (see the alerts there)

im-adithya:
Fixes #241 

What if the user doesn't have any channels? Should we hide the Send/Receive buttons from the Wallet screen

Screenshot 2024-07-19 at 12 43 33 PM

## Send
Screenshot 2024-07-19 at 12 44 14 PM

## Receive
Screenshot 2024-07-19 at 12 43 59 PM


frontend/src/screens/wallet/Send.tsx
```diff
@@ -106,6 +119,27 @@ export default function Send() {
...
+        <>
+          {/* If all channels have less or equal balance than their reserve, show a warning */}
+          {channels?.every(
+            (channel) =>
+              channel.localBalance <=
+              channel.unspendablePunishmentReserve * 1000
+          ) && (
+            <Alert>
+              <AlertTriangle className="h-4 w-4" />
+              <AlertTitle>Channel reserves unmet</AlertTitle>
```
bumi, line 132: Shouldn't this be something like: 

You don't have enough spending balance?

frontend/src/screens/wallet/Receive.tsx
```diff
@@ -116,6 +127,27 @@ export default function Receive() {
         title="Receive"
         description="Create a lightning invoice that can be paid by any bitcoin lightning wallet"
       />
+      {!!channels?.length && (
+        <>
+          {/* If all channels have less than 20% incoming capacity, show a warning */}
+          {channels?.every(
+            (channel) =>
+              channel.remoteBalance <
+              (channel.localBalance + channel.remoteBalance) * 0.2
```
bumi, line 136: can this also check the amount? that if I enter a high amount it will also show the message?

im-adithya: As in if amount is > than receiving capacity of every channel / we consider MPP and check with net receiving capacity?

Or we can also show an error if the amount entered is > receiving capacity (But is it good to restrict the user like that?)

frontend/src/screens/wallet/Send.tsx
im-adithya: Changed to "Low spending balance"

frontend/src/screens/wallet/Receive.tsx
im-adithya: Done!

frontend/src/screens/wallet/Receive.tsx
```diff
@@ -116,6 +127,28 @@ export default function Receive() {
         title="Receive"
         description="Create a lightning invoice that can be paid by any bitcoin lightning wallet"
       />
+      {!!channels?.length && (
```
rolznz, line 130: need to update this check to also only happen if the backend type has channel management (the useInfo hook exposes this, see other places in the source code)

frontend/src/screens/wallet/Receive.tsx
```diff
@@ -116,6 +127,28 @@ export default function Receive() {
...
+        <>
+          {/* If all channels have less than 20% incoming capacity, show a warning */}
+          {channels?.every(
+            (channel) =>
+              channel.remoteBalance <
+                (channel.localBalance + channel.remoteBalance) * 0.2 ||
```
rolznz, line 136: I don't think we need to loop through channels, can we use the balances instead? (see `useBalances`)

frontend/src/screens/wallet/Receive.tsx
rolznz: (also applies to send)

frontend/src/screens/wallet/Receive.tsx
rolznz: (also applies to send)

rolznz:
@im-adithya 

> What if the user doesn't have any channels? Should we hide the Send/Receive buttons from the Wallet screen

Good question, I think we should have a specific alert for this on the main wallet page, and not allow the user to enter the send or receive pages (If the LNClient has channel management and the user has no channels)

frontend/src/screens/wallet/Receive.tsx
im-adithya: Done!

rolznz (approved):
tACK

PR #287, opened 2024-07-17
chore: return if info is not loaded in app layout

### LINKED ISSUE #253: On fresh start, App flickers on home page before going to intro screen
image

The issue is because the homepage is the default route, the info hasn't loaded yet, so no redirect occurs.

Maybe if there's no info we can not render the router (will there still be a flicker as soon as the info is loaded?)

im-adithya:
Fixes #253

Should I return `Outlet` instead of nothing? Didn't right now as it would be weird if the Sidebar pops up after a second everywhere, and info should be present for the Hub in any case.

## Video (previously vs now)
https://github.com/user-attachments/assets/ae65cb3d-e1f4-4196-921d-7323dc3bbe64

### DIFF
--- a/frontend/src/components/layouts/AppLayout.tsx
+++ b/frontend/src/components/layouts/AppLayout.tsx
@@ -85,6 +85,10 @@ export default function AppLayout() {
 
   const isHttpMode = window.location.protocol.startsWith("http");
 
+  if (!info) {
+    return;
+  }
+
   function UserMenuContent() {
     return (
       <DropdownMenuContent align="end">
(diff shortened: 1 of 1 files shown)


frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -85,6 +85,10 @@ export default function AppLayout() {
 
   const isHttpMode = window.location.protocol.startsWith("http");
 
+  if (!info) {
+    return;
```
rolznz, line 89:
```suggestion
    return null;
```

rolznz (approved):
utACK

PR #294, opened 2024-07-18
feat: onboarding checklist 

reneaaron:
- Allow only the next item to be clicked, disable everything else
 - Add numbers


rolznz (approved):
utACK

PR #295, opened 2024-07-19
fix: change mainnet to bitcoin

### LINKED ISSUE #290: LND returns wrong network type (mainnet instead of bitcoin)
This is used in places such as channel partners for increasing receiving capacity

im-adithya:
Fixes #290


rolznz (approved):
utACK

Issue #299, opened 2024-07-19
Isolated Apps / Dynamic Budgets followup tasks


rolznz:
From https://github.com/getAlby/hub/pull/226

- [ ] Enable Roman's cleanup task for old request and response events
- [ ] isolated app possible follow-ups (in case we wish to spend more time on isolated apps)
  - internal payments should trigger NIP-47 notifications (this can be done by the transactions service publishing its own notifications rather than the notifier subscribing to LNClient notifications - this would simplify things too)
  - restrict isolated balance to not receive more than a certain budget (it's hard to do?)
  - prevent hub not spending balance that is set aside for other apps
  - prevent paying more than actual balance from main wallet into isolated wallets (you can send an infinite amount)
  - cashu isolated app payments does not work and causes money to be "lost"
- [ ] currently there are no DB indexes. I am not sure how this will perform when the number of payments grow
- [ ] ListTransactions on LNClients that support notifications should actually just throw an error (they should never be used)
- [ ] fee reserve calculation could be per-LNClient to completely match what the node sets
- [ ] Extra nip-47 controller tests
  - [ ] multi_pay_keysend (partial success)
  - [ ] list_transactions (paging params)

on4r:
> prevent hub not spending balance that is set aside for other apps

this seems very important if you want to share your hub with friends and family

max-peter:
> > prevent hub not spending balance that is set aside for other apps
> 
> this seems very important if you want to share your hub with friends and family

Yes, I second that. Should be a more priority. At least it would be helpfull show it if this happens. Right now I have to watch this myself and that is tricky without the exact balances from each isolated / family wallets!

PR #305, opened 2024-07-20
chore: add transaction indexes, make fees non-nullable

rolznz:
Part of https://github.com/getAlby/hub/issues/299


bumi:
I am in the camp that indexes do not hurt and it's in our case fine to add them. @reneaaron do you want some to be removed? 

regarding the fees: my thinking is that we just try to avoid as much conditionals as possible due to different backends.

rolznz:
@reneaaron merging for now - if you have any feedback on the indexes I'll fix it as a follow-up PR

PR #306, opened 2024-07-20
fix: always calculate budget usage

rolznz:
App card (and app page) will show usage even if no budget is set.

image


rolznz:
This is a bit strange because the field is called `budget` usage. But the app effectively has an infinite budget.

PR #307, opened 2024-07-21
feat: bolt12 (WIP)

rolznz:
Supports:
- [x] Paying amountless Bolt12 offers
- [x] Receiving Bolt12 payments in transaction list

TODO:
- [ ] Decode offer (frontend - needed to check if the offer is amountless or not)
- [ ] Pay Bolt12 offers with a set amount
- [ ] Make an amountless offer - receive screen could have options: [Lightning address | Invoice | Bolt12 (beta)]?
- [ ] NWC planning - pay_offer and make_offer?
- [ ] Review metadata
  - [ ] When paying data can now be passed in? (New!) - max is currently 512 chars, and then is truncated! https://docs.rs/lightning/latest/lightning/offers/invoice_request/constant.PAYER_NOTE_LIMIT.html 
  - [ ] How large a description can be stored on an offer?
- [ ] Questions for LDK
  - [ ] How do I get an offer from an offer ID?
  - [ ] How do I get the payer message in a payment received event? - _they are working on a PR_

TBC:
- [ ] Offer decoding (Go)
- [ ] Offer storage in transaction table

Notes:
- no way to create offers in the UI - an offer is created on LDK node startup
- no online BOLT12 decoder interface? - can use our CLN instance (lightning-cli decode `offer-here`)
- variable vs non-variable amounts - only variable is supported
- no BOLT12 decoder on frontend
- our CLN does not have offers support so Bolt12 routing does not work. I think this means also all LND nodes will not route it because they don't handle the onion messages?, and a lot of payments will not work
- offer creation fails if your node doesn't have a channel partner with a peer that has at least 3 channels and at least one public. I think this will be fixed in the next LDK version.
  - offers are based on the channels you have at the time of creation. So if you open new channels, your old offer probably won't work... so people can't save offers to their address list. They still need to save a lightning address (or BIP 353 address)
- offers can include a payer message. This is cool (but it's limited to just a message?).

# Dev setup:  
2 paired nodes with one node that has at least 3 channels, one of which is public (I think.) See https://github.com/getAlby/hub/issues/308

### DIFF
--- a/api/models.go
+++ b/api/models.go
@@ -35,7 +35,7 @@ type API interface {
 	RedeemOnchainFunds(ctx context.Context, toAddress string) (*RedeemOnchainFundsResponse, error)
 	GetBalances(ctx context.Context) (*BalancesResponse, error)
 	ListTransactions(ctx context.Context, limit uint64, offset uint64) (*ListTransactionsResponse, error)
-	SendPayment(ctx context.Context, invoice string) (*SendPaymentResponse, error)
+	SendPayment(ctx context.Context, invoice string, amount *uint64) (*SendPaymentResponse, error)
 	CreateInvoice(ctx context.Context, amount int64, description string) (*MakeInvoiceResponse, error)
 	LookupInvoice(ctx context.Context, paymentHash string) (*LookupInvoiceResponse, error)
 	RequestMempoolApi(endpoint string) (interface{}, error)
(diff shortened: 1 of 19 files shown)


rolznz:
Update: this should be simpler now as the send and receive UIs have been split into multiple pages.

chrisguida:
OMG YES

rolznz:
closing in favor of https://github.com/getAlby/hub/pull/1242

PR #312, opened 2024-07-22
chore: include esplora url in sync failed events

rolznz:


bumi (approved):
we should be careful to abuse this not too much as logger.

PR #317, opened 2024-07-22
fix: add warning for importing your mnemonic

### LINKED ISSUE #311: Warn before importing seed that Alby Hub cannot be run with the same seed on multiple devices
A user did this, and the 2nd didn't show the channels or the whole onchain activity. We should warn against the user doing this as there could be a lot of unknown/unexpected things happen (especially regarding onchain address reuse).

Also recommend users to do the standard migrate flow.

The only case for this actually is if they want to reuse their existing seed from the browser extension.

reneaaron:
Fixes #311 

image

@rolznz Not sure if we really need to explain what could go wrong in detail, I think it's okay to just have this warning in there?


rolznz (approved):
utACK, looks great!

PR #334, opened 2024-07-24
fix: remove page reload

### LINKED ISSUE #333: Cancel edit permissions doesn't work
Clicking cancel just reloads the page:

https://github.com/getAlby/hub/blob/fix/responsive/frontend/src/screens/apps/ShowApp.tsx#L225

There is a GET param `edit` that can be used to link into edit mode right away, which renders this button useless as you will end up on the same page after that.

reneaaron:
Fixes #333

### DIFF
--- a/frontend/src/screens/apps/ShowApp.tsx
+++ b/frontend/src/screens/apps/ShowApp.tsx
@@ -222,7 +222,7 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
                             type="button"
                             variant="outline"
                             onClick={() => {
-                              window.location.reload();
+                              setEditMode(false);
                             }}
                           >
                             Cancel


frontend/src/screens/apps/ShowApp.tsx
```diff
@@ -222,7 +222,7 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
                             type="button"
                             variant="outline"
                             onClick={() => {
-                              window.location.reload();
+                              setEditMode(false);
```
im-adithya, line 225: No this won't work (try setting the permissions to custom and cancel without saving, you won't see any permissions)

It used to work this way before but @rolznz changed this to window reload because this adds a lot of complexity to the Permissions component (that's why the prop passed ot Permissioins is renamed from `isEditMode` to `readOnly`)

im-adithya: We can somehow remove the edit param and then do the reload to fix the issue instead

rolznz: It's just so if they cancel and edit again, it resets the values. But how many people are going to do this and if they edit it incorrectly they can just edit it again

I actually think @reneaaron 's fix is ok?

im-adithya: It fixes the issue of not reloading with edit mode again but I think it would feel buggy to see the edited state even after clicking cancel, but yeah the chances of that happening are slim

rolznz: oh, it even applies to what's displayed on the screen, hmm...

Maybe let's leave this open then and re-think it

im-adithya: Yeah, but if we want to fix the actual issue for now (#333) we can remove the param value and do the reload

rolznz: I will close this for now - let's do a new PR to fix it

PR #337, opened 2024-07-25
feat: add back button in setup screen

### LINKED ISSUE #310: Add back button in Wails app
There's no browser navigation so it's impossible to go back

im-adithya:
Fixes #310 

## Screenshot
Screenshot 2024-07-25 at 7 23 03 PM


rolznz:
@im-adithya this is for Wails/Desktop mode where it has no browser navigation, so shouldn't we always have a back button in that case?

Therefore, it also doesn't need to be shown in non-desktop mode

rolznz:
@im-adithya sounds good and fixes the main issue. But I think we should whitelist, not blacklist pages since you missed some:

image

rolznz:
@im-adithya I pushed a fix, can you check it?

rolznz (approved):
tACK

im-adithya:
> I pushed a fix, can you check it?

Looks better now, thanks!

PR #338, opened 2024-07-25
 feat: reconnect immediately on first retry

### LINKED ISSUE #297: Immediately try to reconnect to relay first time
Currently we always wait 10 seconds to re-connect, but the problem is this is 10 seconds of lost data.

If it fails again, we need to have some back off though

im-adithya:
Fixes #297 

Is this what we want here @rolznz?


rolznz:
@im-adithya please re-read the issue and think through it. If you don't understand it, please do ask before writing code

im-adithya:
@rolznz can you re-review?

service/start.go
```diff
@@ -46,10 +46,11 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
 		defer svc.wg.Done()
 		//Start infinite loop which will be only broken by canceling ctx (SIGINT)
 		var relay *nostr.Relay
+		immediateRetry := false
```
rolznz, line 49: I think it'd be easier to understand to flip this and call it `waitToReconnectSeconds`, and use a number instead of a boolean, so that later we are able to easily adjust the algorithm (e.g. do an exponential back off algorithm)

service/start.go
```diff
@@ -46,10 +46,11 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
 
 		for i := 0; ; i++ {
 			// wait for a delay before retrying except on first iteration
-			if i > 0 {
+			if i > 0 && !immediateRetry {
```
rolznz, line 53: I would remove the iteration check here and do `if waitToReconnectSeconds > 0`

rolznz:
@im-adithya can you address the feedback?

im-adithya:
Ready now @rolznz!

rolznz (approved):
tACK

Issue #335, opened 2024-07-25
Add a warning for redeeming funds


reneaaron:
https://feedback.getalby.com/-alby-hub-request-a-feature/posts/when-redeeming-funds-only-allow-confirmed-address

PR #340, opened 2024-07-25
fix: improve copy and add confirmation on redeem onchain funds

rolznz:
Fixes https://github.com/getAlby/hub/issues/335

image

image

image


rolznz:
Updated images

PR #343, opened 2024-07-25
fix: copy on open channel button on first channel page

rolznz:


frontend/src/screens/channels/first/FirstChannel.tsx
```diff
@@ -190,7 +190,7 @@ export function FirstChannel() {
             )}
             <LoadingButton loading={isLoading} onClick={openChannel}>
               Open Channel
-              {albyBalance && albyBalance?.sats > 0 && <> and Migrate Funds</>}
+              {canPayForFirstChannel && <> and Migrate Funds</>}
```
reneaaron, line 193: @rolznz I thought also when the balance is too low to pay for a channel, the hosted balances is migrated anyway, right?

rolznz: Ah, yes this is wrong. Thanks! I'll make a new PR

Issue #289, opened 2024-07-18
Unlink Alby Account


rolznz:
Sometimes the user connects with the wrong OAuth account. They should be able to completely unlink it from the app, from the settings page.

PR #347, opened 2024-07-26
feat: unlink alby account

rolznz:
Fixes https://github.com/getAlby/hub/issues/289

Adds a new settings page http://localhost:5173/#/settings/alby-account

image

This currently does not remote the Alby Oauth account's NWC connection. I don't think this should necessarily be the same action

The first link is a bit wrong because it always goes to your currently logged in account on getalby.com which might not be the one logged in in the hub. It would be cool if we can pass something to getalby.com to detect that?

### DIFF
--- a/alby/alby_oauth_service.go
+++ b/alby/alby_oauth_service.go
@@ -48,6 +48,8 @@ const (
 	userIdentifierKey    = "AlbyUserIdentifier"
 )
 
+const ALBY_ACCOUNT_APP_NAME = "getalby.com"
+
 func NewAlbyOAuthService(db *gorm.DB, cfg config.Config, keys keys.Keys, eventPublisher events.EventPublisher) *albyOAuthService {
 	conf := &oauth2.Config{
 		ClientID:     cfg.GetEnv().AlbyClientId,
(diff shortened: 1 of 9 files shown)


frontend/src/screens/settings/AlbyAccount.tsx
```diff
@@ -0,0 +1,93 @@
...
+        throw new Error("No CSRF token");
+      }
+      await request("/api/alby/unlink-account", {
+        method: "POST",
+        headers: {
+          "X-CSRF-Token": csrf,
+          "Content-Type": "application/json",
+        },
+      });
+      await refetchInfo();
```
rolznz, line 48: navigate to the URL instead of refetching the info, and pass a query param so we can include `force_login` in the auth URL

frontend/src/screens/settings/AlbyAccount.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { request } from "src/utils/request";
+
+export function AlbyAccount() {
+  const { data: csrf } = useCSRF();
+  const { toast } = useToast();
+  const { mutate: refetchInfo } = useInfo();
+
+  const unlink = async () => {
+    if (
+      !confirm(
```
rolznz, line 24: remove

alby/alby_oauth_service.go
```diff
@@ -395,6 +395,15 @@ func (svc *albyOAuthService) GetAuthUrl() string {
 	return svc.oauthConf.AuthCodeURL("unused")
 }
 
+func (svc *albyOAuthService) UnlinkAccount(ctx context.Context) error {
+	svc.cfg.SetUpdate(userIdentifierKey, "", "")
```
rolznz, line 399: TODO: also destroy the alby account NWC node

rolznz: Therefore we should also delete the any getalby.com connections on this side

PR #349, opened 2024-07-26
feat: add channels cards on mobile

rolznz:
Extracts Channels table into new component and adds ChannelCards component for mobile


reneaaron (approved):
The diff seems to be a bit odd as there seem to be changes from other branches in here as well? :thinking: 

The channel cards look great! :+1:

PR #355, opened 2024-07-26
fix: button usage for links

bumi:
Button tags in forms submit the form.
This means clicking the "Buy Bitcoin" link did not open the link but actually submit the form. According to the HTML spec a button tag must also not be nested into a hyperlink.

This PR removes those links from the form and uses hyperlinks with button styles (instead of a button tag)

note: I have no idea about those react elements/buttons


rolznz (approved):
Thanks for fixing this! tACK

bumi:
there are some more buttons like this that we probably also should fix and try to have valid HTML.

PR #359, opened 2024-07-27
chore: no default invoice memo

bumi:
what value does it give? then all the entries say lightning invoice and it can only be a lightning invoice


rolznz:
I see, this is especially wrong for keysend payments.

rolznz (approved):
utACK

Issue #360, opened 2024-07-28
Bug: duplicate keysend payment entries in transactions table


rolznz:
One keysend payment results in two transactions, with one not assigned to the app

![image](https://github.com/user-attachments/assets/aeea1cb1-735c-4525-8706-3b8edc7c41b4)

PR #361, opened 2024-07-28
fix: duplicate transaction entry in keysend payments

rolznz:
Fixes https://github.com/getAlby/hub/issues/360

This is due to the transaction entry in the DB initially not having a payment hash, and we will not know it until the LNClient keysend method returns (which internally may fire a pay_keysend notification in the LDK/LND case)

### DIFF
--- a/transactions/transactions_service.go
+++ b/transactions/transactions_service.go
@@ -321,14 +321,28 @@ func (svc *transactionsService) SendKeysend(ctx context.Context, amount uint64, 
 
 	// the payment definitely succeeded
 	now := time.Now()
-	dbErr := svc.db.Model(&dbTransaction).Updates(map[string]interface{}{
-		"State":          constants.TRANSACTION_STATE_SETTLED,
-		"PaymentHash":    paymentHash,
-		"Preimage":       &preimage,
-		"FeeMsat":        &fee,
-		"FeeReserveMsat": 0,
-		"SettledAt":      &now,
-	}).Error
+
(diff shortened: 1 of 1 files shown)


transactions/transactions_service.go
```diff
@@ -321,14 +321,28 @@ func (svc *transactionsService) SendKeysend(ctx context.Context, amount uint64,
...
+
+	dbErr := svc.db.Transaction(func(tx *gorm.DB) error {
+		// Because we don't know the payment hash in advance when creating the original keysend transaction,
+		// there may have been another transaction created with the same payment hash by the nwc_payment_sent event.
+		// Delete that one and update the one we already created.
+		// This would not be needed if we always can pass the payment hash when doing keysend.
+		err := tx.Delete(&db.Transaction{}, "payment_hash = ?", paymentHash).Error
+		if err != nil {
+			return err
+		}
```
bumi, line 333: hmm, is there no other way. feels like a hack and just thrown code at it.

bumi: I think in LNDHub we create the preimage and the payment hash

rolznz: @bumi we can do this if all the connectors that support passing a payment hash to the keysend method.

- More LDK custom changes on our fork to pass a payment hash rather then letting LDK generate one
- Disable Keysend on Breez and Greenlight since they don't conform and we don't want to spend time on them

Then Keysend will only be supported by LDK and LND.

What do you think?

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/385 (WIP)

PR #362, opened 2024-07-29
fix: do not allow duplicate getalby.com apps

rolznz:


alby/alby_oauth_service.go
```diff
@@ -396,6 +396,11 @@ func (svc *albyOAuthService) GetAuthUrl() string {
 }
 
 func (svc *albyOAuthService) LinkAccount(ctx context.Context, lnClient lnclient.LNClient, budget uint64, renewal string) error {
+	appName := "getalby.com"
```
bumi, line 399: what happens if I create one and call it like that?

rolznz: There will be :dragon:

PR #363, opened 2024-07-29
fix: keep anchor reserves when withdrawing savings balance

### LINKED ISSUE #344: LDK uses wrong method to redeem funds
The one we use also spends anchor reserves, putting channels at higher risk of not being able to be closed

https://docs.rs/ldk-node/latest/ldk_node/payment/struct.OnchainPayment.html#method.send_to_address

reneaaron:
Fixes #344

### DIFF
--- a/lnclient/ldk/ldk.go
+++ b/lnclient/ldk/ldk.go
@@ -1000,6 +1000,8 @@ func (ls *LDKService) GetOnchainBalance(ctx context.Context) (*lnclient.OnchainB
 
 func (ls *LDKService) RedeemOnchainFunds(ctx context.Context, toAddress string) (string, error) {
 	spendableBalance := ls.node.ListBalances().SpendableOnchainBalanceSats
+	// TODO: estimate the transaction fee and subtract that from the spendable balance
+	// to avoid spending any of the reserved anchor channel balance
 	txId, err := ls.node.OnchainPayment().SendToAddress(toAddress, spendableBalance)
 	if err != nil {
 		logger.Logger.WithError(err).Error("SendToAddress failed")
--- a/lnclient/ldk/ldk.go
+++ b/lnclient/ldk/ldk.go
@@ -999,9 +999,10 @@ func (ls *LDKService) GetOnchainBalance(ctx context.Context) (*lnclient.OnchainB
 }
 
 func (ls *LDKService) RedeemOnchainFunds(ctx context.Context, toAddress string) (string, error) {
-	txId, err := ls.node.OnchainPayment().SendAllToAddress(toAddress)
+	spendableBalance := ls.node.ListBalances().SpendableOnchainBalanceSats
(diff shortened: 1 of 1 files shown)


lnclient/ldk/ldk.go
```diff
@@ -999,9 +999,10 @@ func (ls *LDKService) GetOnchainBalance(ctx context.Context) (*lnclient.OnchainB
 }
 
 func (ls *LDKService) RedeemOnchainFunds(ctx context.Context, toAddress string) (string, error) {
-	txId, err := ls.node.OnchainPayment().SendAllToAddress(toAddress)
+	spendableBalance := ls.node.ListBalances().SpendableOnchainBalanceSats
```
rolznz, line 1002: I don't think this will account for onchain fees, but it is definitely better than now :raised_hands:

lnclient/ldk/ldk.go
```diff
@@ -999,9 +999,10 @@ func (ls *LDKService) GetOnchainBalance(ctx context.Context) (*lnclient.OnchainB
...
+	txId, err := ls.node.OnchainPayment().SendToAddress(toAddress, spendableBalance)
```
rolznz, line 1003: The argument should be in millisats - can you multiply the spendable balance by 1000?

reneaaron: I am wondering where this is going wrong because I tested this feature while developing it and the amount seems to be correct.

However, after multiplying the spendableBalance with 1000 to get the amount in msat, I get an error "not enough funds available". See this example with the current code I have:

1. Request 1M sats from mutinynet
tb1qnga845us4t4hafkp8aneq5f7tf9za9sr9u55f5
https://mutinynet.com/tx/7fc3a8cf78a3872774dcd6c112609915388ed586ba4ae7bc339446c78234487d
2. Savings balance updates to 999,786 sats
3. Generate new onchain address
tb1quq39g744c9e926tt6uzdrerq225d8pue9wffnh
3. Withdraw Savings Balance to this address
https://mutinynet.com/tx/a89dfa542b6efdd24d3c95738dc5a66dbc087a2917e19e51739164ba8e47ac6e
4. Spending balance updates to 999,572 sats
5. Send funds back to Mutinynet
tb1qd28npep0s8frcm3y7dxqajkcy2m40eysplyr9v
6. Savings balance updates to 0 sats

### Docs
https://sourcegraph.com/github.com/lightningdevkit/ldk-node/-/blob/src/payment/onchain.rs?L47
https://docs.rs/ldk-node/latest/ldk_node/payment/struct.OnchainPayment.html#method.send_to_address

rolznz: I created a bug in LDK, thanks for testing this!

rolznz (approved):
utACK

lnclient/ldk/ldk.go
rolznz, line 1003: Maybe would be nice to add a TODO above to try detect the transaction fee and subtract that from the spendable balance

PR #365, opened 2024-07-29
feat: show tlv records in transaction popup

### LINKED ISSUE #31: Improve transaction lists
Follow-up to getAlby/nostr-wallet-connect-next#465 

- ~Add pagination to transactions (remove hack to get all transactions in one go)~ It's complicated as LND has separate methods to fetch incoming and outgoing
- ~Add modal to show transaction details on click~

### LINKED ISSUE #372: Change transactions table metadata column to JSON type instead of string
This would make metadata more easy to query. But I'm not sure if it simplifies things with the ORM - it still looks like it needs to be manually encoded and decoded? (see https://github.com/go-gorm/datatypes)

im-adithya:
Fixes #31
Fixes #372
Adds custom record info in the transaction popup

## Screenshot
Screenshot 2024-07-29 at 8 47 18 PM


frontend/src/components/TransactionItem.tsx
```diff
@@ -191,52 +191,143 @@ function TransactionItem({ tx }: Props) {
...
+                      // Chat message
+                      return (
+                        <div className="mt-6" key={record.type}>
+                          <p>Message</p>
+                          <p className="text-muted-foreground break-all">
+                            {hexToString(record.value)}
+                          </p>
+                        </div>
+                      );
+                    } else if (record.type === 7629169) {
```
bumi, line 220: can this be a separate component?
to make this a bit smaller here (also if we have more)

im-adithya: > to make this a bit smaller here

Yup I wanted to but wasn't sure where exactly to break it into a different component

Also do you mean a component just for the podcasting info?

bumi: yep, I thought for the type 7629169

im-adithya: Done!

bumi:
I am not sure how the data processing works there. can we get that message on the transaction list? where we have the "description" ? cc @rolznz

im-adithya:
Yes we can, but we have to discuss on the priority order
For example `boostagram > keysend message > description`

frontend/src/types.ts
```diff
@@ -377,7 +377,33 @@ export type Transaction = {
   fees_paid: number;
   created_at: string;
   settled_at: string | undefined;
-  metadata: unknown;
+  metadata: {
+    tlv_records: TLVRecord[];
```
rolznz, line 381: We have control over our own API - we could actually decode this on the backend

frontend/src/components/PodcastingInfo.tsx
```diff
@@ -0,0 +1,63 @@
+import { Boostagram } from "src/types";
+
+function PodcastingInfo({ boost }: { boost: Boostagram }) {
+  return (
+    <>
+      {boost?.message && (
```
rolznz, line 6: Everywhere `boost` has a `?` afterwards - this should not be needed

rolznz:
> Yes we can, but we have to discuss on the priority order
For example boostagram > keysend message > description

Sounds good (and the description would be empty in this case I think anyway)

bumi:
but we would need to persist that data in the DB. we should not on the fly always decode those keysends. 

that's why we have the DB. prepare the data as our app needs it.

frontend/src/components/PodcastingInfo.tsx
im-adithya: Oops yes ✅

frontend/src/types.ts
bumi: +1 we should store the data we need it now as we have the payments table. 
also for example have a special field for the name, message metadata.

a well defined API is most important to be able to easily and flexible build features.

though I am not sure how much podcasting knowledge should be in the hub - but guess it's important.

rolznz:
@bumi @im-adithya I think we should go with Bumi's recommendation, right?

@im-adithya would you like to propose the changes you would make? or how should we proceed here?

frontend/src/types.ts
im-adithya: Done, now the API returns decoded tlv values. We already store them in metadata column, so I changed the api struct to give decoded tlv records in metadata

im-adithya:
TODOs
- [ ] Fix/Add tests
- [ ] Store proper description (boostagram > keysend message > description) in db

api/models.go
```diff
@@ -192,18 +192,36 @@ type ListTransactionsResponse = []Transaction
...
+	FeesPaid        uint64             `json:"fees_paid"`
+	CreatedAt       string             `json:"created_at"`
+	SettledAt       *string            `json:"settled_at"`
+	AppId           *uint              `json:"app_id"`
+	Metadata        *lnclient.Metadata `json:"metadata,omitempty"`
+	Boostagram      *Boostagram        `json:"boostagram,omitempty"`
+}
+
+type Boostagram struct {
+	AppName        string `json:"app_name"`
```
rolznz, line 211: Most of our JSON payloads are in `camelCase` - the transaction model and now this are inconsistent with the rest of the app - any chance you could update these?

db/migrations/202408061737_add_boostagrams_and_use_json.go
```diff
@@ -0,0 +1,32 @@
...
+
+// This migration adds boostagram column to transactions
+var _202408061737_add_boostagrams_and_use_json = &gormigrate.Migration{
+	ID: "202408061737_add_boostagrams_and_use_json",
+	Migrate: func(tx *gorm.DB) error {
+		if err := tx.Exec("ALTER TABLE transactions ADD COLUMN boostagram JSONB").Error; err != nil {
+			return err
+		}
+
+		if err := tx.Exec(`
```
rolznz, line 18: can this be wrapped in a transaction (note: `tx` is not actually a transaction by default in gormigrate - see one of the recent migrations of how I did it)

rolznz: Thanks for updating the metadata column!

frontend/src/types.ts
```diff
@@ -378,7 +378,35 @@ export type Transaction = {
   fees_paid: number;
   created_at: string;
   settled_at: string | undefined;
-  metadata: unknown;
+  metadata?: {
+    tlv_records: TLVRecord[];
```
rolznz, line 382: can you remove the old tlv_records

frontend/src/types.ts
```diff
@@ -378,7 +378,35 @@ export type Transaction = {
...
+  };
+  boostagram?: Boostagram;
+};
+
+export type TLVRecord = {
```
rolznz, line 387: this is no longer needed

transactions/transactions_service.go
```diff
@@ -263,13 +265,17 @@ func (svc *transactionsService) SendKeysend(ctx context.Context, amount uint64,
 	metadata := map[string]interface{}{}
 
 	metadata["destination"] = destination
-	metadata["tlv_records"] = customRecords
-	metadataBytes, err := json.Marshal(metadata)
 
+	// we copy records because parseMetadata decodes record values
```
rolznz, line 269: Is this actually needed? and I think it would be good for `parseMetadata` to not do this anyway so there is no side effect

transactions/transactions_service.go
```diff
@@ -607,14 +620,19 @@ func (svc *transactionsService) ConsumeEvent(ctx context.Context, event *events.
 
 			if result.RowsAffected == 0 {
 				// Note: brand new payments cannot be associated with an app
-				var metadata string
+				var metadataBytes []byte
```
rolznz, line 623: I actually don't think we should do this on `nwc_payment_sent` because we already save the metadata when we initiate the payment

frontend/src/components/TransactionItem.tsx
```diff
@@ -191,52 +192,68 @@ function TransactionItem({ tx }: Props) {
...
+                Details
+                {showDetails ? (
+                  <ChevronUp className="w-4 h-4" />
+                ) : (
+                  <ChevronDown className="w-4 h-4" />
+                )}
+              </div>
+              {showDetails && (
+                <>
+                  {tx.metadata?.tlv_records.map((record) => {
```
rolznz, line 210: We discussed that we should not the TLV records from the frontend

Note: LDK also does not support this record type currently and we need to make a change in our fork to support it. I don't think we really need this one? CC @bumi 

It looks like this is untested anyway?

transactions/transactions_service.go
im-adithya: Not that important but what if payments happen outside (they don't have metadata since it is a payment, atleast in LND but still I thought it was good to have if we are anyways adding metadata)

Also we should DRY up the "brand new" lnClient.Transaction processing because it is almost the same in both sent/received

transactions/transactions_service.go
im-adithya: Done!

db/migrations/202408061737_add_boostagrams_and_use_json.go
im-adithya: > can this whole migration be wrapped in a transaction 

Is this because we don't want only one to succeed in case of failures?

Done btw 🙌

frontend/src/components/TransactionItem.tsx
im-adithya: Yeah the frontend is not yet ready.

> we need to make a change in our fork to support it

Then I'll remove the Whatsat message from the frontend (which will also remove the need for tlv in frontend completely).

If it is present in the backend I'll save it in the transactions db description column in whatever backend type it is supported 👍

db/migrations/202408061737_add_boostagrams_and_use_json.go
rolznz: We don't want the migration to half-succeed - it's hard to recover from

db/migrations/202408061737_add_boostagrams_and_use_json.go
```diff
@@ -0,0 +1,30 @@
...
+// This migration adds boostagram column to transactions
+var _202408061737_add_boostagrams_and_use_json = &gormigrate.Migration{
+	ID: "202408061737_add_boostagrams_and_use_json",
+	Migrate: func(db *gorm.DB) error {
+		err := db.Transaction(func(tx *gorm.DB) error {
+			return tx.Exec(`
+			ALTER TABLE transactions ADD COLUMN boostagram JSONB;
+			ALTER TABLE transactions ADD COLUMN metadata_temp JSONB;
+			UPDATE transactions SET metadata_temp = "";
+			UPDATE transactions SET metadata_temp = json(metadata) where metadata != "";
```
rolznz, line 19: a few things seem wrong here, it doesn't work with existing transactions, it uses the json() function but the column types are jsonb, I'm not sure what to do with existing metadata values that were an empty string - if they should be changed NULL in this case

No json queries are working for me.

db/migrations/202408061737_add_boostagrams_and_use_json.go
```diff
@@ -0,0 +1,30 @@
...
+
+// This migration adds boostagram column to transactions
+var _202408061737_add_boostagrams_and_use_json = &gormigrate.Migration{
+	ID: "202408061737_add_boostagrams_and_use_json",
+	Migrate: func(db *gorm.DB) error {
+		err := db.Transaction(func(tx *gorm.DB) error {
+			return tx.Exec(`
+			ALTER TABLE transactions ADD COLUMN boostagram JSONB;
+			ALTER TABLE transactions ADD COLUMN metadata_temp JSONB;
+			UPDATE transactions SET metadata_temp = NULL;
```
rolznz, line 18: this line is not needed

db/migrations/202408061737_add_boostagrams_and_use_json.go
```diff
@@ -0,0 +1,30 @@
...
+			UPDATE transactions SET metadata_temp = json(metadata) where metadata != "";
```
rolznz, line 19: we have a mismatch between json and jsonb

im-adithya: Fixed to use JSON

rolznz:
Example queries:
```
SELECT *
FROM transactions
WHERE EXISTS (
    SELECT 1
    FROM json_each(metadata, '$.tlv_records') AS records
    WHERE json_extract(records.value, '$.type') = 7629169
);
```

rolznz:
Rolands tests (WIP):
- [x] migration with transactions with both metadata and no metadata
- [x] query of metadata from migrated transactions
- [ ] query of metadata from new transactions
- [ ] querying boostagrams

api/transactions.go
```diff
@@ -73,9 +73,9 @@ func toApiTransaction(transaction *transactions.Transaction) *Transaction {
 		preimage = transaction.Preimage
 	}
 
-	var metadata interface{}
-	if transaction.Metadata != "" {
-		jsonErr := json.Unmarshal([]byte(transaction.Metadata), &metadata)
+	var metadata *Metadata
```
rolznz, line 76: I don't think this should be a pointer - then you use `&metadata` which is basically getting a pointer to a pointer

api/transactions.go
```diff
@@ -84,6 +84,19 @@ func toApiTransaction(transaction *transactions.Transaction) *Transaction {
 		}
 	}
 
+	var boostagram *Boostagram
```
rolznz, line 87: here too

frontend/src/components/PodcastingInfo.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <p>Total amount</p>
+          <p className="text-muted-foreground break-all">
+            {new Intl.NumberFormat(undefined, {}).format(
+              Math.floor(boost.value_msat_total / 1000)
+            )}{" "}
+            {Math.floor(boost.value_msat_total / 1000) == 1 ? "sat" : "sats"}
+          </p>
+        </div>
+      )}
+      {boost.sender_name && (
```
rolznz, line 47: these were not updated after changing to camelCase

frontend/src/types.ts
```diff
@@ -380,7 +380,25 @@ export type Transaction = {
   fees_paid: number;
   created_at: string;
   settled_at: string | undefined;
-  metadata: unknown;
+  metadata?: Record<string, unknown>;
+  boostagram?: Boostagram;
+};
+
+export type Boostagram = {
+  app_name: string;
```
rolznz, line 388: these were not renamed. I will do this

api/transactions.go
rolznz: oops, not this one

api/transactions.go
```diff
@@ -84,6 +84,19 @@ func toApiTransaction(transaction *transactions.Transaction) *Transaction {
...
+	if transaction.Boostagram != nil {
+		var txBoostagram *transactions.Boostagram
+		jsonErr := json.Unmarshal(transaction.Boostagram, &txBoostagram)
+		if jsonErr != nil {
+			logger.Logger.WithError(jsonErr).WithFields(logrus.Fields{
+				"id":         transaction.ID,
```
rolznz, line 93: use payment hash instead of transaction id

transactions/transactions_service.go
```diff
@@ -836,40 +795,6 @@ func makePreimageHex() ([]byte, error) {
 	return bytes, nil
 }
 
-func (svc *transactionsService) parseMetadata(metadata lnclient.Metadata) ([]byte, error) {
```
im-adithya: Why are we removing this, we decided not to decode it on the fly right? Or are you pushing another change which does this during db save?

rolznz: We don't decode boostagrams on the fly. This is the metadata - which we also do not decode on the fly (it's returned as-is in NWC requests)

rolznz: The frontend does not use the `metadata`.

api/models.go
```diff
@@ -200,15 +200,35 @@ type Transaction struct {
...
+}
+
+type Metadata = map[string]interface{}
+
+type Boostagram struct {
+	AppName        string `json:"appName"`
+	Name           string `json:"name"`
+	Podcast        string `json:"podcast"`
+	URL            string `json:"url"`
+	Episode        string `json:"episode,omitempty"`
```
rolznz, line 222: @im-adithya why do some have omitempty others not? is it required?

im-adithya: I took this from LBE repo [here](https://github.com/getAlby/lightning-browser-extension/blob/f1a7dd86fbad8e71d7095fc7507b55d79bdc86e6/src/types.ts#L957), whereever we have `?:` in the Boostagram type there, I've omitted

transactions/transactions_service.go
im-adithya: Ah okay, then it should be simpler 👍

transactions/transactions_service.go
rolznz: I decided to remove it for now - we need to confirm if we want to add this support and if so, also import transactions from LND (all this is a lot of extra work and code to do properly)

api/models.go
rolznz: but it looks completely arbitrary and doesn't seem to follow the spec - https://github.com/lightning/blips/blob/master/blip-0010.md#fields

rolznz (approved):
tACK

PR #368, opened 2024-07-30
feat: add channels cards on mobile

rolznz:
Moves the channel status calculation to the backend (returned in the API response), moves the channels table to a new component, adds a channels cards component, and extracts some shared code e.g. the ChannelWarning component


rolznz:
@reneaaron merging for now, let me know if you have any comments

PR #369, opened 2024-07-30
feat: setup i18n

### LINKED ISSUE #225: feat: i18n support
I think now is a good time to start adding support for translations

im-adithya:
Fixes #225

Adds i18n support

## Screenshot
Screenshot 2024-07-30 at 8 37 24 PM


frontend/src/components/LocaleSwitcher.tsx
```diff
@@ -0,0 +1,41 @@
...
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "src/components/ui/select";
+import { toast } from "src/components/ui/use-toast";
+import i18n, { supportedLocales } from "src/i18n/i18nConfig";
+
+export default function LocaleSwitcher() {
+  const fallbackLng = i18n.options.fallbackLng?.[0 as keyof FallbackLng];
```
rolznz, line 14: The fallback should not be optional, it's specified in the config

rolznz: Are you sure this is used correctly? it's not an array in the config

frontend/src/components/LocaleSwitcher.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const [dropdownLang, setDropdownLang] = useState(
+    i18n.language || fallbackLng
+  );
+
+  const languageHandler = async (newLanguage: string) => {
```
rolznz, line 19:
```suggestion
  const onLanguageChange = async (newLanguage: string) => {
```

frontend/src/components/LocaleSwitcher.tsx
im-adithya: Yeah this works, when I tried locally, it is also done the same way in LBE

frontend/src/i18n/locales/en/translation.json
```diff
@@ -0,0 +1,21 @@
+{
+  "translation": {
+    "settings": {
+      "language": {
+        "title": "Language",
+        "description": "Translations are not 100% finished yet."
```
rolznz, line 6:
```suggestion
        "description": "Choose the language to display in Alby Hub"
```

frontend/src/i18n/i18nConfig.ts
```diff
@@ -0,0 +1,47 @@
...
+  // pass the i18n instance to react-i18next.
+  .use(initReactI18next)
+  // detect user language
+  // learn more: https://github.com/i18next/i18next-browser-languageDetector
+  .use(LanguageDetector)
+  // init i18next
+  // for all options read: https://www.i18next.com/overview/configuration-options
+  .init({
+    fallbackLng: "en",
+    ns: ["translation", "common", "components", "permissions"],
```
rolznz, line 41: we don't use any of these except "translation" and "permissions" seems to not be added to the json files

rolznz:
Nice work @im-adithya !

frontend/src/i18n/i18nConfig.ts
im-adithya: Permissions got in here by mistake 😅 
Regarding others, I wasn't sure whether to add all in this PR itself as that would cause conflicts 🤔  best to limit this PR to only i18n setup no?

rolznz: Sounds good. I would just add the ones that are actually used in this PR, and others can be added later

frontend/src/components/LocaleSwitcher.tsx
```diff
@@ -0,0 +1,45 @@
...
+      setDropdownLang(newLanguage);
+      i18n.changeLanguage(newLanguage);
+      toast({ title: t("success") });
+    }
+  };
+
+  return (
+    <Select value={dropdownLang} onValueChange={onLanguageChange}>
+      <SelectTrigger className="w-[150px]">
+        <SelectValue placeholder="Language" />
```
im-adithya, line 34: aNot sure if we need a translation for Language too, if yes where should it go? In components or common?

rolznz: I put it in the component for now

rolznz (approved):
tACK

rolznz:
Unfortunately we decided to postpone this as 3-6 months. Marking as draft for now.

PR #376, opened 2024-07-31
chore: include whether the channel is public or not in channel ready event

rolznz:
With this we can have custom emails and monitors/workflows if a user opens a public and private channel for example

### DIFF
--- a/lnclient/ldk/ldk.go
+++ b/lnclient/ldk/ldk.go
@@ -1244,12 +1244,15 @@ func (ls *LDKService) handleLdkEvent(event *ldk_node.Event) {
 			logger.Logger.WithField("event", eventType).Error("Failed to find channel by ID")
 			return
 		}
+		channel := channels[channelIndex]
 		ls.eventPublisher.Publish(&events.Event{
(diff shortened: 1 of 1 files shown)


lnclient/ldk/ldk.go
```diff
@@ -1236,11 +1236,20 @@ func (ls *LDKService) handleLdkEvent(event *ldk_node.Event) {
...
+		if channelIndex == -1 {
+			logger.Logger.WithField("event", eventType).Error("Failed to find channel by ID")
+			return
+		}
 		ls.eventPublisher.Publish(&events.Event{
 			Event: "nwc_channel_ready",
 			Properties: map[string]interface{}{
 				"counterparty_node_id": eventType.CounterpartyNodeId,
 				"node_type":            config.LDKBackendType,
+				"public":               channels[channelIndex].IsPublic,
```
bumi, line 1252: is there other relevant information? like the capacity?

rolznz: I added capacity and also whether the channel is outbound or not

bumi (approved):

PR #378, opened 2024-07-31
feat: add close channel dialog

im-adithya:
## Description
`window.alert`, `prompt` and `confirm` don't work in wails. We can use [Dialog](https://wails.io/docs/reference/runtime/dialog/) but it has some limitations in [MacOS](https://stackoverflow.com/questions/26898941/ios-wkwebview-not-showing-javascript-alert-dialog). Also the flow for closing a channel involves the user to type "force close" which is not ideal. This PR replaces those with AlertDialog from shadcn for Wails support as well as design consistency

Screenshot 2024-07-31 at 9 10 18 PM
Screenshot 2024-07-31 at 9 10 30 PM
Screenshot 2024-08-07 at 4 45 05 PM
Screenshot 2024-08-07 at 4 45 20 PM
Screenshot 2024-07-31 at 9 11 01 PM

### DIFF
--- a/frontend/src/components/CloseChannelDialog.tsx
+++ b/frontend/src/components/CloseChannelDialog.tsx
@@ -1,0 +1,243 @@
+import { AlertTriangleIcon, CopyIcon, ExternalLinkIcon } from "lucide-react";
+import React from "react";
+import ExternalLink from "src/components/ExternalLink";
+import { Alert, AlertDescription, AlertTitle } from "src/components/ui/alert";
+import { Button } from "src/components/ui/button";
+import { Checkbox } from "src/components/ui/checkbox";
+import { toast } from "src/components/ui/use-toast";
+import { useChannels } from "src/hooks/useChannels";
+import { useCSRF } from "src/hooks/useCSRF";
+import { copyToClipboard } from "src/lib/clipboard";
+import { Channel, CloseChannelResponse } from "src/types";
+import { request } from "src/utils/request";
+import {
+  AlertDialogCancel,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from "./ui/alert-dialog";
+
+type Props = {
+  alias: string;
+  channel: Channel;
+};
+
+export function CloseChannelDialog({ alias, channel }: Props) {
+  const [closeType, setCloseType] = React.useState("normal");
+  const [step, setStep] = React.useState(channel.active ? 2 : 1);
+  const [fundingTxId, setFundingTxId] = React.useState("");
+  const { data: channels, mutate: reloadChannels } = useChannels();
+  const { data: csrf } = useCSRF();
+
+  const onContinue = () => {
+    setStep(step + 1);
+  };
+
+  const copy = (text: string) => {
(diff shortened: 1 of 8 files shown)


reneaaron (commented):
Just from a quick glance at the screenshots: 

1) "Proceed only if you still want to continue" - Either remove or rephrase, but this feels wrong. 

2) If you present the user with 2 options (where the user can only pick one) rather use selects / radios than checkboxes, checkboxes always suggest you can pick multiple ones)

im-adithya:
> "Proceed only if you still want to continue" - Either remove or rephrase, but this feels wrong.

Yeah felt weird to me as well, will remove it entirely, maybe we can change the "Continue" in the button text to something else?

> rather use selects / radios than checkboxes

I was a bit hesitant to use Radio as that would add another component, so I used existing one. I didn't use Select/Boxes like budget options as I wanted to add description so it's more understandable for the user. So `Radio`?

reneaaron:
> > "Proceed only if you still want to continue" - Either remove or rephrase, but this feels wrong.
> 
> Yeah felt weird to me as well, will remove it entirely, maybe we can change the "Continue" in the button text to something else?

Sure, maybe "Confirm"? 

> > rather use selects / radios than checkboxes
> 
> I was a bit hesitant to use Radio as that would add another component, so I used existing one. I didn't use Select/Boxes like budget options as I wanted to add description so it's more understandable for the user. So `Radio`?

Yeah, I'd go for that one. :+1:

im-adithya:
Done!

frontend/src/components/CloseChannelDialog.tsx
```diff
@@ -0,0 +1,246 @@
...
+                  />
+                  <div className="grid gap-1.5">
+                    <Label
+                      htmlFor="force"
+                      className="text-primary font-medium cursor-pointer"
+                    >
+                      Force Close
+                    </Label>
+                    <p className="text-sm text-muted-foreground">
+                      Closes the channel unilaterally, can take longer and might
```
rolznz, line 187: I don't think people will understand what this means

rolznz:
@reneaaron @im-adithya I'm not sure about the UI. Ideally users do normal closures, so I don't think they should have to pick between the two like this. The force closure option should be a last resort

frontend/src/components/channels/ChannelsCards.tsx
```diff
@@ -80,55 +79,51 @@ export function ChannelsCards({
...
-                            </DropdownMenuItem>
-                            {channel.public && (
-                              <DropdownMenuItem
-                                className="flex flex-row items-center gap-2 cursor-pointer"
-                                onClick={() => editChannel(channel)}
-                              >
-                                <HandCoins className="h-4 w-4" />
-                                Set Routing Fee
+                        <AlertDialog>
+                          <DropdownMenu>
```
rolznz, line 83: Can you move this dropdown menu to a separate component in this folder? I realized its the same for both the channels table and cards

im-adithya: Done!

im-adithya:
We agreed on making the copy understandable and dealing with the Normal / Force Close UI later. Ready for review!

rolznz (approved):
tACK

PR #386, opened 2024-08-02
fix: use private channels in invoice routing

im-adithya:
This solves people with private channels having issues with invoice creation

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -241,13 +241,6 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
 		expiry = lnclient.DEFAULT_INVOICE_EXPIRY
 	}
 
-	addInvoiceRequest := &lnrpc.Invoice{
-		ValueMsat:       amount,
-		Memo:            description,
-		DescriptionHash: descriptionHashBytes,
-		Expiry:          expiry,
-	}
-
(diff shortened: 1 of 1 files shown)


lnclient/lnd/lnd.go
```diff
@@ -241,7 +241,7 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
 		expiry = lnclient.DEFAULT_INVOICE_EXPIRY
 	}
 
-	resp, err := svc.client.AddInvoice(ctx, &lnrpc.Invoice{ValueMsat: amount, Memo: description, DescriptionHash: descriptionHashBytes, Expiry: expiry})
+	resp, err := svc.client.AddInvoice(ctx, &lnrpc.Invoice{ValueMsat: amount, Memo: description, DescriptionHash: descriptionHashBytes, Expiry: expiry, Private: true})
```
rolznz, line 244: I am not sure we should enable this always. LDK only enables it if you have no public channels

im-adithya: But what is the guarantee that the public channels connect all the way from sender to us? Aren't we then lessening our chances of a successful payment?

im-adithya: How does it work in LDK? Here this also uses public channels after setting the `Private` flag (tested in Polar)

rolznz (approved):
tACK

PR #387, opened 2024-08-02
fix: do not show sidebar while loading

im-adithya:
This should fix the loading issue when node is unlocked/not running

image

### DIFF
--- a/frontend/src/components/layouts/AppLayout.tsx
+++ b/frontend/src/components/layouts/AppLayout.tsx
@@ -91,6 +91,10 @@ export default function AppLayout() {
     return null;
   }
 
+  if (!info.setupCompleted || !info.running || !info.unlocked) {
+    return <Outlet />;
+  }
+
   function UserMenuContent() {
     return (
       <DropdownMenuContent align="end">
--- a/frontend/src/components/layouts/AppLayout.tsx
+++ b/frontend/src/components/layouts/AppLayout.tsx
@@ -91,10 +91,6 @@ export default function AppLayout() {
     return null;
   }
 
-  if (!info.setupCompleted || !info.running || !info.unlocked) {
-    return <Outlet />;
-  }
-
   function UserMenuContent() {
     return (
       <DropdownMenuContent align="end">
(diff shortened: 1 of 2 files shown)


pavanjoshi914:
works. but still it shows blank white screen and then goes to login screen. but maybe it's too complex to handle state or show something consistent there while things are loading.

frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -91,6 +91,10 @@ export default function AppLayout() {
     return null;
   }
 
+  if (!info.setupCompleted || !info.running || !info.unlocked) {
```
rolznz, line 94: This looks unsafe to me. What about we move this info check
```
if (!info) {
  return null;
}
```

Into App.tsx e.g.
`{info && <RouterProvider router={router} />}`

im-adithya: 👍 yup that would be good, but line 94 does the fix here, should we also move it to App.tsx is what you mean?

rolznz: Line 94 with any of these conditions not being true feels worrying to me - it could break one of our pages

im-adithya: But if any of it is false, we do the redirect in `HomeRedirect`. `Outlet` is anyways rendered even now. But it is just rendered with the Sidebar. (And it consists of the `HomeRedirect` component which then transfers to the intro/start/unlock pages) This change just removes the Sidebar in those cases by rendering just the `Outlet` (`HomeRedirect` component)

rolznz: @im-adithya this makes it more complicated and does not cleanly solve the problem. We don't even know what page to render if the info is not loaded. Rendering the outlet makes no sense

im-adithya:
This change won't fix the issue, we somehow have to change the HomeRedirect or render only the Outlet but not the Sidebar when HomeRedirect is deciding which page to render

This is on current master [7d4e7b6](https://github.com/getAlby/hub/commit/7d4e7b60ee6baf064a4e509942f695e5a91e9dbb)


https://github.com/user-attachments/assets/c907161e-cc63-472e-86b3-b6425fdc96a5

PR #395, opened 2024-08-04
feat: improve install scripts for linux

bumi:
this adds a default linux installation script to download and run Alby Hub on a plain Linux machine.


scripts/linux-x86_64/README.md
```diff
@@ -0,0 +1,53 @@
+# Alby Hub on a Linux server
+
+## Requirements
+
+- Linux distribution
+- Runs pretty much on any VPS or server
```
rolznz, line 6: For VPS maybe it's good to recommend the minimum RAM requirements?

bumi:
I think so, yes, but this would change the URLs that's why I left it.

rolznz:
@bumi I guess it is better to fix it now than later - anyway the way the update works is it will get the correct script file from getalby.com, right? so the impact should be minimal?

bumi:
@rolznz done.

@JPLACLAU can you help checking the URLs used. 
we should use this `https://getalby.com/install/hub/pi-zero-install.sh` install URL

scripts/linux-x86_64/README.md
```diff
@@ -0,0 +1,53 @@
...
+We recommend inspecting the install script and if needed adjusting it or taking inspiration from it for your setup.
+
+As a general good practice we recommend creating a new system user.
+
+    $ adduser albyhub
+
+Run the installation script on your server:
+
+    $ wget https://raw.githubusercontent.com/getAlby/hub/master/scripts/linux-x86_64/install.sh
+    $ ./install.sh
```
rolznz, line 20: if we do this, it won't run as the correct user, right? it's not the same as the ssh command below

scripts/linux-x86_64/README.md
rolznz: bump

scripts/linux-x86_64/install.sh
```diff
@@ -0,0 +1,88 @@
...
+[Unit]
+Description=Alby Hub
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=simple
+Restart=always
+RestartSec=1
+User=$USER
```
rolznz, line 71: how does this work?

rolznz: Ah, I see

scripts/linux-x86_64/README.md
bumi: why not?

nip47/event_handler.go
```diff
@@ -236,15 +236,17 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
...
+					"responseEventNostrId": resp.ID,
+					"eventKind":            event.Kind,
+					"appId":                app.ID,
 				}).WithError(err).Error("Failed to publish event")
 				requestEvent.State = db.REQUEST_EVENT_STATE_HANDLER_ERROR
 			} else {
 				requestEvent.State = db.REQUEST_EVENT_STATE_HANDLER_EXECUTED
 				logger.Logger.WithFields(logrus.Fields{
 					"requestEventNostrId": event.ID,
+					"responseEventNostrId": resp.ID,
```
bumi, line 249: sorry. no idea how this got in there. 😱 
should be in another branch - but I still want this in.

bumi:
merging it as I need it.

scripts/linux-x86_64/README.md
rolznz: we mix two different users, if the user runs a different command next time they use a different user. In my opinion this is confusing

bumi: I don't get why they run next time a differnt user?

rolznz: there are 2 commands they can run - if the next time they use a different one based on whether they are ssh'd in or not, it uses a different user. I did this by accident

bumi: which 2 commands are you talking about and why would the user be different?

bumi: this is the current version: https://github.com/getAlby/hub/pull/415/files

PR #403, opened 2024-08-05
fix: check if peer is already connected

### LINKED ISSUE #394: LND connect peer fails if the peer is already connected
we should not return an error in this case but just succeed - as the peer is already connected and that's what we wanted.

im-adithya:
Fixes #394

Checks if the peer is already connected, and returns nil

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -546,7 +546,14 @@ func (svc *LNDService) GetNodeConnectionInfo(ctx context.Context) (nodeConnectio
 }
 
 func (svc *LNDService) ConnectPeer(ctx context.Context, connectPeerRequest *lnclient.ConnectPeerRequest) error {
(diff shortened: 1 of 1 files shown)


bumi:
can we catch the error? then we don't have to always do two calls?
if the error is that a peer is already connected we treat it as success.

bumi:
utACK

rolznz:
I think this is quite fragile error handling:

```
if grpcErr, ok := status.FromError(err); ok {
		if strings.HasPrefix(grpcErr.Message(), "already connected to peer") {
			return nil
		}
	}
```

If the error message changes slightly this could break again. Does this particular error have a specific code we can use?

im-adithya:
The code we get is `codes.Unknown`, LND doesn't give any specific errors sadly

rolznz (approved):
utACK

bumi:
> I think this is quite fragile error handling:

I said and think this is OK. because I think it's better than doing a second request to look for the peers with all the addtional error cases and we don't have a code to identify it and I don't think the message changes likely.

PR #404, opened 2024-08-05
chore: stop subscription goroutines if node is stopped

### LINKED ISSUE #388: Alby Hub spams logs if LND is not running
```
{"error":"EOF","level":"error","msg":"Failed to receive invoice","time":"2024-08-02T20:50:31+07:00"}
{"error":"rpc error: code = Unknown desc = routerrpc server shutting down","level":"error","msg":"Failed to receive payment","time":"2024-08-02T20:50:31+07:00"}
{"error":"EOF","level":"error","msg":"Failed to receive invoice","time":"2024-08-02T20:50:31+07:00"}
{"error":"rpc error: code = Unknown desc = routerrpc server shutting down","level":"error","msg":"Failed to receive payment","time":"2024-08-02T20:50:31+07:00"}
{"error":"rpc error: code = Unknown desc = routerrpc server shutting down","level":"error","msg":"Failed to receive payment","time":"2024-08-02T20:50:31+07:00"}
```

im-adithya:
Fixes #388

We should also add a screen to display when the node is offline, which can be used by backends other than LDK, because currently Alby Hub goes blank if the node is offline with no message. But we can open a separate issue for that(?)

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -522,7 +522,7 @@ func NewLNDService(ctx context.Context, eventPublisher events.EventPublisher, ln
 				for {
 					invoice, err := invoiceStream.Recv()
 					if err != nil {
-						logger.Logger.WithError(err).Error("Failed to receive payment")
+						logger.Logger.WithError(err).Error("Failed to receive invoice")
 						select {
 						case <-lndCtx.Done():
 							return
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -6,7 +6,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"io"
 	"math"
 	"sort"
 	"strconv"
@@ -15,8 +14,6 @@ import (
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	decodepay "github.com/nbd-wtf/ln-decodepay"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
 	"github.com/getAlby/hub/events"
 	"github.com/getAlby/hub/lnclient"
@@ -438,8 +435,10 @@ func NewLNDService(ctx context.Context, eventPublisher events.EventPublisher, ln
 			})
 			if err != nil {
 				logger.Logger.WithError(err).Error("Error subscribing to payments")
+				time.Sleep(10 * time.Second)
 				continue
 			}
+		paymentsLoop:
 			for {
 				select {
 				case <-lndCtx.Done():
@@ -447,14 +446,8 @@ func NewLNDService(ctx context.Context, eventPublisher events.EventPublisher, ln
 				default:
 					payment, err := paymentStream.Recv()
 					if err != nil {
-						if grpcErr, ok := status.FromError(err); ok {
-							if grpcErr.Code() == codes.Unknown {
-								logger.Logger.Error("LND node stopped or unreachable, exiting payment subscription")
-								return
-							}
-						}
 						logger.Logger.WithError(err).Error("Failed to receive payment")
-						continue
+						break paymentsLoop
 					}
 
 					switch payment.Status {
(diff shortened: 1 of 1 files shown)


lnclient/lnd/lnd.go
```diff
@@ -444,6 +447,12 @@ func NewLNDService(ctx context.Context, eventPublisher events.EventPublisher, ln
 				default:
 					payment, err := paymentStream.Recv()
 					if err != nil {
+						if grpcErr, ok := status.FromError(err); ok {
+							if grpcErr.Code() == codes.Unknown {
```
im-adithya, line 451: `codes.Unknown` feels wrong, but we should otherwise use `grpcErr.Message() == "routerrpc server shutting down"` which looks hacky, what should I do here?

rolznz:
@im-adithya @bumi  what about we shut down the LNClient if we detect LND is not running, and also do not allow Alby Hub to start if LND is not running. Is this possible?

The current change only stops the subscriptions, and leaves the LNClient in a weird state

im-adithya:
This is how the Hub looks once the node goes down. I tried starting it again and after a few seconds (roughly a minute), it was working perfectly as it used to 👌 

Screenshot 2024-08-06 at 4 02 20 PM
Screenshot 2024-08-06 at 4 02 29 PM
Screenshot 2024-08-06 at 4 02 38 PM
Screenshot 2024-08-06 at 4 03 08 PM

lnclient/lnd/lnd.go
```diff
@@ -28,9 +28,9 @@ import (
 )
 
 type LNDService struct {
-	client *wrapper.LNDWrapper
-	pubkey string
-	cancel context.CancelFunc
+	client   *wrapper.LNDWrapper
+	nodeInfo *lnclient.NodeInfo
```
im-adithya, line 32: Lmk if this is good, then I will do this for other backends too

lnclient/lnd/lnd.go
```diff
@@ -95,7 +95,18 @@ func (svc *LNDService) ListTransactions(ctx context.Context, from, until, limit,
 }
 
 func (svc *LNDService) GetInfo(ctx context.Context) (info *lnclient.NodeInfo, err error) {
-	resp, err := svc.client.GetInfo(ctx, &lnrpc.GetInfoRequest{})
+	if svc.nodeInfo == nil {
```
rolznz, line 98: this is set on startup, we don't need this code, right?

lnclient/lnd/lnd.go
```diff
@@ -445,7 +458,7 @@ func NewLNDService(ctx context.Context, eventPublisher events.EventPublisher, ln
 					payment, err := paymentStream.Recv()
 					if err != nil {
 						logger.Logger.WithError(err).Error("Failed to receive payment")
-						continue
+						break paymentsLoop
```
rolznz, line 461: this double for loop is a bit scary and I don't think this `lndCtx.Done()` is hit if LND is not running (because this for loop will probably never be entered?)

@im-adithya this is a case with complicated logic - can you please write down the tests you did to cover these cases? Is there a simpler way we can implement it? (I'm not sure there is)

lnclient/lnd/lnd.go
rolznz: I am not sure we should just do this on every backend. If we want to do that, I'd much rather store the network on startup and add a call to get the network to the LNClient interface

lnclient/lnd/lnd.go
im-adithya: Ah lol, true

lnclient/lnd/lnd.go
im-adithya: > I don't think this lndCtx.Done() is hit if LND is not running

Yes, this won't be hit because `lndClient.SubscribePayments` call itself would fail

im-adithya: I made some changes, this should be better and will also end the goroutine in all cases so it should be safer too.

Tests done to check subscription logic:

1. Normal run to see if subscription receives event and handles it properly
2. Started Hub (with node running in Polar) and then killed context from Hub's go terminal -> Goes to the <-lndCtx.Done() case instead of breaking the paymentsLoop because ctx is ended, thereby ending the goroutine
3. Started Hub (with node running in Polar) and stopped the node in Polar after the subscription is started -> Breaks the payment loop after waiting for 2s and tries to subscribe again. It fails now and continues after waiting for 10s (this keeps going on and on)
4. After (3), I turned the node on again, and the next attempt after 10s worked and started the subscription again 🔁 
5. After (3), I cancelled the context while it was waiting for 10s and it ended the goroutine due to switch case

rolznz (approved):
utACK

lnclient/lnd/lnd.go
rolznz: @im-adithya awesome!

PR #408, opened 2024-08-06
chore: reduce number of info logs, remove echologrus middleware

rolznz:


http/http_service.go
```diff
@@ -61,13 +61,21 @@ func (httpSvc *HttpService) RegisterSharedRoutes(e *echo.Echo) {
...
+		LogHost:      true,
+		LogRequestID: true,
 		LogValuesFunc: func(c echo.Context, values middleware.RequestLoggerValues) error {
 			logger.Logger.WithFields(logrus.Fields{
-				"uri":    values.URI,
-				"status": values.Status,
-			}).Debug("handled echo request")
+				"uri":        values.URI,
+				"status":     values.Status,
+				"remote_ip":  values.RemoteIP,
```
bumi, line 74: on cloudflare this would likely be the cloudflare IP and one would like the connecting_ip from the CF- header

rolznz: I will test this with the fly gateway

PR #417, opened 2024-08-07
feat: move mnemonic decryption to backend

### LINKED ISSUE #65: Some features not available in non-https mode (e.g. albyhub.local)
- crypto.subtle which is needed to decrypt the mnemonic locally in order to allow the user to do a key backup
- ~clipboard copy - I believe we should use a copy library that has a fallback or better support for this situation~

reneaaron:
Fixes #65


api/api.go
```diff
@@ -609,11 +609,15 @@ func (api *api) GetInfo(ctx context.Context) (*InfoResponse, error) {
...
 }
 
-func (api *api) GetEncryptedMnemonic() *EncryptedMnemonicResponse {
-	resp := EncryptedMnemonicResponse{}
-	mnemonic, _ := api.cfg.Get("Mnemonic", "")
-	resp.Mnemonic = mnemonic
-	return &resp
+func (api *api) GetMnemonic(unlockPassword string) (*MnemonicResponse, error) {
+	resp := MnemonicResponse{}
+	encryptedMnemonic, err := api.cfg.Get("Mnemonic", unlockPassword)
```
rolznz, line 614:
```suggestion
	mnemonic, err := api.cfg.Get("Mnemonic", unlockPassword)
```

http/http_service.go
```diff
@@ -169,8 +169,28 @@ func (httpSvc *HttpService) infoHandler(c echo.Context) error {
...
-	responseBody := httpSvc.api.GetEncryptedMnemonic()
+func (httpSvc *HttpService) mnemonicHandler(c echo.Context) error {
+	var mnemonicRequest api.MnemonicRequest
+	if err := c.Bind(&mnemonicRequest); err != nil {
+		return c.JSON(http.StatusBadRequest, ErrorResponse{
+			Message: fmt.Sprintf("Bad request: %s", err.Error()),
+		})
+	}
+
+	if !httpSvc.cfg.CheckUnlockPassword(mnemonicRequest.UnlockPassword) {
```
rolznz, line 180: can this check be moved to the API level so it is used in wails mode too?

reneaaron, line 180: In http we seem to check the unlock password multiple times, but not for the wails handlers?

api/api.go
```diff
@@ -609,11 +609,20 @@ func (api *api) GetInfo(ctx context.Context) (*InfoResponse, error) {
...
+func (api *api) GetMnemonic(unlockPassword string) (*MnemonicResponse, error) {
+	resp := MnemonicResponse{}
+
+	if !api.cfg.CheckUnlockPassword(unlockPassword) {
+		return &resp, fmt.Errorf("wrong password")
+	}
+
+	mnemonic, err := api.cfg.Get("Mnemonic", unlockPassword)
+	if err != nil {
+		return &resp, fmt.Errorf("failed to fetch encryption key: %w", err)
```
rolznz, line 621:
```suggestion
		return nil, fmt.Errorf("failed to fetch encryption key: %w", err)
```

rolznz (approved):
tACK

api/api.go
```diff
@@ -609,11 +609,20 @@ func (api *api) GetInfo(ctx context.Context) (*InfoResponse, error) {
 	return &info, nil
 }
 
-func (api *api) GetEncryptedMnemonic() *EncryptedMnemonicResponse {
-	resp := EncryptedMnemonicResponse{}
-	mnemonic, _ := api.cfg.Get("Mnemonic", "")
+func (api *api) GetMnemonic(unlockPassword string) (*MnemonicResponse, error) {
+	resp := MnemonicResponse{}
```
rolznz, line 613: this can be moved to the bottom e.g.
```
return &MnemonicResponse {
  Mnemonic: mnemonic
}
```

Issue #381, opened 2024-07-31
Auth improvements topics


rolznz:
JWT auth + Fly proxy + possible second layer of encryption + fixing decrypting seed locally are all interconnected and need to be solved together.

- remove CSRF token with an API authentication 
- ~~What are the options for an additional layer of encryption (research additional TLS options to solve potential MITM proxy issues)~~ -> decided not to do this
- How does the Fly proxy work? does it remove the need for subdomains? -> yes, progress has started on fly gateway
- Send unlock password to hub backend and hub will return seed protected by HTTPS and possibly another layer of encryption - this will solve the issue with users unable to backup their seed on .local addresses
- JWT sounds good and saving to local storage, but how does it fit with the above? (JWT? - some key exchange that is also used for some transport encryption?)

rolznz:
All tasks have been completed :tada:

PR #418, opened 2024-08-07
feat: JWT auth

rolznz:
Part of https://github.com/getAlby/hub/issues/381

Replaces cookie auth + CSRF with JWT.

JWT is stored in local storage.

Notes:
- All CSRF code removed
- COOKIE_SECRET replaced with JWT_SECRET
- Token is saved in local storage
- added JWT_EXPIRY_DAYS so expiry can be customised so app developers can set a longer expiry. Default is 30 days.
- Restricted group with JWT middleware is used instead of custom middleware
- Logout is done clientside now - JWTs/sessions are currently not stored in the hub
- Token injection is done in the http platform-specific request - (external API access only works in http mode).
- View token in settings -> developer (EXPERIMENTAL)
- XSS protection - React has some XSS protection built in but we still need to be careful with any user data.

TODOs:
- [ ] security review
  - [x] CSP headers

Tests done:
- onboarding
- start
- login (brand new session)
- invalid token
- expired token
- logout

Screenshots

image


config/config.go
```diff
@@ -69,20 +69,20 @@ func (cfg *config) init(env *AppConfig) {
...
 
-	// set the cookie secret to the one from the env
-	// if no cookie secret is configured we create a random one and store it in the DB
-	cfg.CookieSecret = cfg.Env.CookieSecret
-	if cfg.CookieSecret == "" {
+	// set the JWT secret to the one from the env
+	// if no JWT secret is configured we create a random one and store it in the DB
+	cfg.JWTSecret = cfg.Env.JWTSecret
+	if cfg.JWTSecret == "" {
 		hex, err := randomHex(20)
```
reneaaron, line 76: Where does this number "20" come from? Is this enough entropy?

> Minimum recommended length: At least 256 bits (32 characters if using base64 or 64 characters if using hexadecimal)

reneaaron (approved):
tACK

config/config.go
rolznz: Good catch! we should increase it then. 1 hex character = 4 bits = 80 bits entropy.

http/http_service.go
```diff
@@ -264,7 +262,7 @@ func (httpSvc *HttpService) unlockHandler(c echo.Context) error {
 		Event: "nwc_unlocked",
 	})
 
-	return c.NoContent(http.StatusNoContent)
+	return c.JSON(http.StatusOK, token)
```
bumi, line 265: what JSON response is this?
If I understand this correctly this is just a string? I am wondering if some consistent JSON structure is better?

rolznz: sounds good. It can be an object with a token string field

rolznz: updated

Issue #409, opened 2024-08-06
Support environment variable to disable advanced setup option on onboarding


rolznz:
E.g. for Nodana or other platforms that have a specific backend requirement

PR #421, opened 2024-08-07
feat: add env variable for disabling advanced setup

rolznz:
Closes https://github.com/getAlby/hub/issues/409

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,10 @@ The following configuration options can be set as environment variables or in a 
 - `WORK_DIR`: directory to store NWC data files. Default: $XDG_DATA_HOME/albyhub
 - `LOG_LEVEL`: log level for the application. Higher is more verbose. Default: 4 (info)
 
+## Node-specific backend parameters
+
+- `ENABLE_ADVANCED_SETUP`: set to `false` to force a specific backend type (combined with backend parameters below)
+
 ### LND Backend parameters
 
 Currently only LND can be configured via env. Other node types must be configured via the UI.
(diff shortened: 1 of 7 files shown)


frontend/src/screens/Intro.tsx
```diff
@@ -81,7 +81,7 @@ export function Intro() {
               </div>
               <div className="mt-20">
                 <Button onClick={() => api?.scrollNext()} size="lg">
-                  Get Started
+                  Next
```
rolznz, line 84: @reneaaron I changed this so we don't have two "Get Started" - raised by @bumi a while back

README.md
```diff
@@ -144,6 +144,10 @@ The following configuration options can be set as environment variables or in a
 - `WORK_DIR`: directory to store NWC data files. Default: $XDG_DATA_HOME/albyhub
 - `LOG_LEVEL`: log level for the application. Higher is more verbose. Default: 4 (info)
 
+## Node-specific backend parameters
+
+- `ENABLE_ADVANCED_SETUP`: set to `false` to force a specific backend type (combined with backend parameters below)
```
reneaaron, line 149: If a specific backend is configured, do we still need to provide other setup options? In what environment would that be the case? :thinking:

reneaaron (approved):
utACK

README.md
rolznz: In Umbrel for example we allow the user to go to advanced and use an LDK node instead of the default (LND)

On Nodana however, they want to force a node type (Phoenixd)

Issue #414, opened 2024-08-07
Further reduce logs


bumi:
https://github.com/getAlby/hub/blob/6227c3e582a10ba799fab27393d668c122213721/nip47/notifications/nip47_notifier.go#L181-L184
I think here the full notification is too much (e.g. includes tlv records) and with streaming sats this might be a lot 

https://github.com/getAlby/hub/blob/6227c3e582a10ba799fab27393d668c122213721/nip47/notifications/nip47_notifier.go#L123-L126
same as above

https://github.com/getAlby/hub/blob/6227c3e582a10ba799fab27393d668c122213721/transactions/transactions_service.go#L593
this might lack queryable information like the payment hash. id is not a good key, because if I search for id:x then I might find all kind of different entries (e.g. apps, permissions, etc.) 

seems we have some inconsitencies between `request_event_id` and `requestEventNostrI` - I think typically snake cases is more often used in logs (at least echo does and our other apps mostly)

some nostr logs miss the response event which actually should be available

Issue #422, opened 2024-08-07
Bug: Invoice payment shows up twice


rolznz:
![image](https://github.com/user-attachments/assets/2ec2f5b9-9a15-4297-90b8-5c851c841075)

rolznz:
This is possible if:
- The user tries to pay the same invoice twice creating two outstanding transactions in the DB
- We get a payment settled event from LDK and all transactions are updated (I believe this is a bug in our code using .Find instead of .Take)

PR #423, opened 2024-08-08
fix: use take instead of find for single record lookups

rolznz:
Related to https://github.com/getAlby/hub/issues/422
Closes https://github.com/getAlby/hub/issues/414

Improves query performance. From gorm:

> Using Find without a limit for single object db.Find(&user) will query the full table and return only the first object which is not performant and nondeterministic

### DIFF
--- a/nip47/notifications/nip47_notifier.go
+++ b/nip47/notifications/nip47_notifier.go
@@ -127,7 +127,7 @@ func (notifier *Nip47Notifier) notifySubscriber(ctx context.Context, app *db.App
 	logger.Logger.WithFields(logrus.Fields{
 		"notification": notification,
 		"appId":        app.ID,
-	}).Info("Notifying subscriber")
+	}).Debug("Notifying subscriber")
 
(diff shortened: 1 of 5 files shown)


nip47/notifications/nip47_notifier.go
```diff
@@ -179,8 +183,8 @@ func (notifier *Nip47Notifier) notifySubscriber(ctx context.Context, app *db.App
 		return
 	}
 	logger.Logger.WithFields(logrus.Fields{
-		"notification": notification,
-		"appId":        app.ID,
+		// "notification": notification,
+		"appId": app.ID,
 	}).Info("Published notification event")
```
bumi, line 188: is this a useful log?

bumi (approved):
utACK.
sounds like an important one.

nip47/notifications/nip47_notifier.go
rolznz: Probably only for debugging. I changed it to a debug log so it'll be filtered out in production

Issue #419, opened 2024-08-07
Remove all instances of window alert, prompt and confirm


rolznz:
These do not work in Wails (Desktop mode)

See https://github.com/getAlby/hub/pull/378

PR #427, opened 2024-08-08
feat: edit routing fee dialog

im-adithya:
Fixes routing fee not returning in LND list channels

Adds Edit Routing Fee Dialog to replace window.prompts, related to #419 

## Screenshots
Screenshot 2024-08-12 at 2 35 37 PM


frontend/src/components/RoutingFeeDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+            value={forwardingFee}
+            onChange={(e) => {
+              setForwardingFee(e.target.value.trim());
+            }}
+          />
+          <ExternalLink
+            to="https://guides.getalby.com/user-guide/v/alby-account-and-browser-extension/alby-hub/faq-alby-hub/how-can-i-change-routing-fees#understanding-routing-fees-and-alby-hub"
+            className="underline flex items-center mt-4"
+          >
+            Learn more about routing fees
```
rolznz, line 102: we mix routing and forwarding fees. Should we use routing fees everywhere? I think it's easier to understand?

frontend/src/components/RoutingFeeDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  return (
+    <AlertDialogContent>
+      <AlertDialogHeader>
+        <AlertDialogTitle>Update Channel Forwarding Fee</AlertDialogTitle>
+        <AlertDialogDescription>
+          <p className="mb-4">
+            Adjust the fee you charge for each payment passing through your
+            channel.{" "}
```
rolznz, line 79: Should we add:

By default we set a high fee to prevent unwanted routing.

im-adithya: Not sure if we do this in LND (and I don't think we should override there for other backends), should I add this only for LDK then?

im-adithya: Not sure if this looks good, too much of text (maybe we can rearrange, lmk) and it also adds an info call:

Screenshot 2024-08-09 at 7 04 00 PM

frontend/src/components/RoutingFeeDialog.tsx
im-adithya: Yeah "Routing Fees" is better, done!

frontend/src/components/RoutingFeeDialog.tsx
rolznz: Ah I see.

Maybe we can change the copy to "A high fee (e.g. 100,000 sats) can be set to prevent unwanted routing."

The input should also show the current value, right? in your screenshot it shows an empty value

im-adithya: Yes it will, here it's 0 that's why it's empty

im-adithya: Looks good, changed (updated screenshot in PR main comment)

frontend/src/components/channels/ChannelDropdownMenu.tsx
```diff
@@ -22,14 +24,19 @@ import { Channel } from "src/types";
...
 export function ChannelDropdownMenu({
   alias,
   channel,
-  editChannel,
 }: ChannelDropdownMenuProps) {
+  const [dialog, setDialog] = React.useState<"close" | "routingFee" | null>(
+    null
+  );
+
+  const openCloseDialog = () => setDialog("close");
```
rolznz, line 37: Can you read the name of this function out loud and ask yourself does it make sense?

rolznz (approved):
tACK

PR #429, opened 2024-08-09
fix(wails): extract force flag during close channel

im-adithya:
Fixes this error in wails during channel closure

image


rolznz (approved):
tACK

The code is not so clean but until we find a better way to do the router (similar to echo) it is ok for now

PR #430, opened 2024-08-09
chore: add logo in missing places

im-adithya:
Just what the lock screen needs to look 🔥 

Screenshot 2024-08-09 at 1 27 50 PM

Screenshot 2024-08-09 at 1 26 36 PM


frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -295,8 +295,8 @@ export default function AppLayout() {
                 >
                   <nav className="grid gap-2 text-lg font-medium">
                     <div className="p-3 ">
-                      <Link to="/" className="font-semibold text-xl">
-                        <span className="">Alby Hub</span>
+                      <Link to="/">
+                        <AlbyHubLogo className="text-foreground" />
```
reneaaron, line 299: Why text-foreground? That should be the default for text, right?

frontend/src/components/layouts/TwoColumnFullScreenLayout.tsx
```diff
@@ -72,7 +73,7 @@ export default function TwoColumnFullScreenLayout() {
       >
         <div className="flex-1 w-full h-full flex flex-col">
           <div className="flex flex-row justify-between items-center">
-            <h1 className="text-lg font-medium">Alby Hub</h1>
+            <AlbyHubLogo className="text-background" />
```
reneaaron, line 76: Why text-background?

reneaaron (commented):
Nice :100:

frontend/src/components/layouts/AppLayout.tsx
im-adithya: I saw this being done similarly above, in [L220](https://github.com/getAlby/hub/blob/master/frontend/src/components/layouts/AppLayout.tsx#L220)

frontend/src/components/layouts/TwoColumnFullScreenLayout.tsx
im-adithya: This is the same color being used to highlight the Quote Author name in white, so I picked this

rolznz: Yeah, this one is needed because the background is the opposite of normal

frontend/src/components/layouts/AppLayout.tsx
rolznz: We did this so that the theme primary color does not apply. Sadly this is only the case for the yellow theme on light mode where it doesn't look good :(

rolznz (approved):
tACK

PR #431, opened 2024-08-09
fix: remove navbar breakpoints

reneaaron:
The navbar currently has some in-between breakpoints which makes it a little bit narrower. I don't think it's worth to maintain them as the differences are negligible but cause some problems: (e.g. with the version number here)

image

### DIFF
--- a/frontend/src/components/layouts/AppLayout.tsx
+++ b/frontend/src/components/layouts/AppLayout.tsx
@@ -145,7 +145,7 @@ export default function AppLayout() {
   function MainNavSecondary() {
     const { hasChannelManagement } = useInfo();
     return (
-      <nav className="grid items-start p-2 text-sm font-medium lg:px-4">
+      <nav className="grid items-start px-4 py-2 text-sm font-medium">
         {hasChannelManagement && (
           <MenuItem to="/channels">
             <CubeIcon className="h-4 w-4" />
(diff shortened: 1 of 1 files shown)


frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -145,7 +145,7 @@ export default function AppLayout() {
   function MainNavSecondary() {
     const { hasChannelManagement } = useInfo();
     return (
-      <nav className="grid items-start p-2 text-sm font-medium lg:px-4">
+      <nav className="grid items-start px-4 py-2 text-sm font-medium">
```
rolznz, line 148: @reneaaron This adds double the horizontal padding on mobile and I'm worried it will cause issues, was this intentional?

frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -210,11 +210,11 @@ export default function AppLayout() {
   return (
     <>
       <div className="font-sans min-h-screen w-full flex flex-col">
-        <div className="flex-1 h-full grid md:grid-cols-[220px_1fr] lg:grid-cols-[280px_1fr]">
+        <div className="flex-1 h-full grid md:grid-cols-[280px_1fr]">
```
rolznz, line 213: do we need to use a hardcoded pixel measurement?

frontend/src/components/layouts/AppLayout.tsx
reneaaron: On mobile we have the drawer anyway, right? This just removes smaller paddings on `md` screens...

frontend/src/components/layouts/AppLayout.tsx
reneaaron: We could use the `w-72` class potentially. I'll update it :+1:

frontend/src/components/layouts/AppLayout.tsx
rolznz: Ah true, thanks!

rolznz (approved):
utACK

PR #434, opened 2024-08-09
chore: add disconnect peer dialog

im-adithya:
Adds DisconnectPeer to LND

Adds disconnect peer dialog, related to #419 

Fixes text in toast by adding `break-all`

## Screenshot

Screenshot 2024-08-09 at 3 55 37 PM


frontend/src/screens/peers/Peers.tsx
```diff
@@ -18,20 +23,19 @@ import {
...
-import { useCSRF } from "../../hooks/useCSRF.ts";
 
 export default function Peers() {
   useSyncWallet();
-  const { data: peers, mutate: reloadPeers } = usePeers();
+  const { data: peers } = usePeers();
   const { data: channels } = useChannels();
   const [nodes, setNodes] = React.useState<Node[]>([]);
-  const { data: csrf } = useCSRF();
+  const [openPeerId, setOpenPeerId] = React.useState<string | null>(null);
```
rolznz, line 38: what is an `open peer`

rolznz: why do we need multiple dialogs? can't you just render the dialog component outside of the table?

im-adithya: I thought the Trigger is necessary for the Dialog to function, fixed now, should be good. I think this way we can also simplify the transactions list. Wil also have a look there

frontend/src/components/DisconnectPeerDialog.tsx
```diff
@@ -0,0 +1,66 @@
...
+type Props = {
+  peer: Peer;
+};
+
+export function DisconnectPeerDialog({ peer }: Props) {
+  const { mutate: reloadPeers } = usePeers();
+
+  async function disconnectPeer() {
+    try {
+      if (!peer.nodeId) {
```
rolznz, line 24: how can a peer exist but not the nodeId?

frontend/src/screens/peers/Peers.tsx
```diff
@@ -56,31 +59,20 @@ export default function Peers() {
     loadNodeStats();
   }, [loadNodeStats]);
 
-  async function disconnectPeer(peerId: string) {
+  async function checkDisconnectPeer(peer: Peer) {
     try {
-      if (!peerId) {
+      if (!peer.nodeId) {
```
rolznz, line 64: this check is not needed, check the diff

frontend/src/screens/peers/Peers.tsx
```diff
@@ -18,18 +20,19 @@ import {
...
+import { Node, Peer } from "src/types";
 import { request } from "src/utils/request";
 
 export default function Peers() {
   useSyncWallet();
-  const { data: peers, mutate: reloadPeers } = usePeers();
+  const { data: peers } = usePeers();
   const { data: channels } = useChannels();
   const [nodes, setNodes] = React.useState<Node[]>([]);
+  const [openPeer, setOpenPeer] = React.useState<Peer | null>(null);
```
rolznz, line 35:
```suggestion
  const [peer, setPeer] = React.useState<Peer | null>(null);
```

rolznz, line 35:
```suggestion
  const [openPeer, setOpenPeer] = React.useState<Peer>();
```

frontend/src/components/DisconnectPeerDialog.tsx
rolznz: @im-adithya can you please test this? otherwise we are adding extra unnecessary code and setting up poor conventions which will be copied through the app

im-adithya: I had the same question, but this check was used previously so I assumed there might be an issue with one of the backends where nodeId is not returned. Should I remove this?

frontend/src/screens/peers/Peers.tsx
im-adithya: If it can't be null then the dialog would be open always...

im-adithya: I thought this would be confusing with the peer in the mapping

frontend/src/components/DisconnectPeerDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+        title: "Failed to disconnect peer: " + e,
+      });
+      console.error(e);
+    }
+  }
+
+  return (
+    <AlertDialogContent>
+      <AlertDialogHeader>
+        <AlertDialogTitle className="break-all">
```
rolznz, line 46: Why did you add this to 3 different places in this file? surely one is enough?

frontend/src/components/ui/toast.tsx
```diff
@@ -92,7 +92,7 @@ const ToastTitle = React.forwardRef<
 >(({ className, ...props }, ref) => (
   <ToastPrimitives.Title
     ref={ref}
-    className={cn("text-sm font-semibold [&+div]:text-xs", className)}
+    className={cn("break-all text-sm font-semibold [&+div]:text-xs", className)}
```
rolznz, line 95: we should be careful with using `break-all`. Can you try `break-words`?

frontend/src/screens/peers/Peers.tsx
rolznz: Unless you pass a value to `useState` then the type will be `Peer | undefined` which means the default will be `undefined`. We do this much more often than using null

frontend/src/components/ui/toast.tsx
im-adithya: Yup was also concerned about "more" being broken into "m-ore". But `break-words` doesn't do the job, the problem is caused by one word (the pubkey)

rolznz (approved):
tACK

PR #435, opened 2024-08-09
chore: change hub update link

### LINKED ISSUE #433: Use getalby.com link for updates
Currently when the user clicks to get an update, the link only works for Alby Cloud instances. This one will work for self-hosted too.

Use `http://getalby.com/update/hub?version=oldversion` instead.

Where oldversion is the current hub version.

im-adithya:
Fixes #433

### DIFF
--- a/frontend/src/components/layouts/AppLayout.tsx
+++ b/frontend/src/components/layouts/AppLayout.tsx
@@ -223,7 +223,7 @@ export default function AppLayout() {
                       <Tooltip>
                         <TooltipTrigger>
                           <ExternalLink
-                            to="https://getalby.com/hub_deployment/edit" // TODO: link to general update page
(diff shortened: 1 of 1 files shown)


frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -223,7 +223,7 @@ export default function AppLayout() {
                       <Tooltip>
                         <TooltipTrigger>
                           <ExternalLink
-                            to="https://getalby.com/hub_deployment/edit" // TODO: link to general update page
+                            to={`https://getalby.com/update/hub${info.version && `?version=${info.version}`}`}
```
rolznz, line 226: the version should always be set, can we remove this conditional?

rolznz (approved):
utACK

Issue #437, opened 2024-08-09
Support customKey/customValue for isolated apps


rolznz:
Like LNDHub, Alby Hub could also use custom key and value for its isolated apps (subaccounts)

LNDHub:
https://getalby.com/.well-known/keysend/hello =>
```
{"status":"OK","tag":"keysend","pubkey":"030a58b8653d32b99200a2334cfe913e51dc7d155aa0116c176657a4f1722677a3","customData":[{"customKey":"696969","customValue":"017rsl75kNnSke4mMHYE"}]}
```

Alby Hub:
- custom value could be an app ID or app pubkey.
- how to give the custom value to getalby.com? could it be returned in the NWC get_info call?

PR #443, opened 2024-08-11
feat: process podcasting custom key/value records

bumi:
Fixes https://github.com/getAlby/hub/issues/437

This loads the app for a keysend transaction from the custom key/value data.

it is using the app ID as costum value - the nostr key of the connection is a bit more private than the app id of the hub

### DIFF
--- a/transactions/transactions_service.go
+++ b/transactions/transactions_service.go
@@ -11,6 +11,7 @@ import (
 	"math"
 	"slices"
 	"strings"
+	"strconv"
 	"time"
 
 	"github.com/getAlby/hub/constants"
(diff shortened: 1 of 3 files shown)


transactions/transactions_service.go
```diff
@@ -599,6 +621,7 @@ func (svc *transactionsService) ConsumeEvent(ctx context.Context, event *events.
 					ExpiresAt:       expiresAt,
 					Metadata:        datatypes.JSON(metadataBytes),
 					Boostagram:      datatypes.JSON(boostagramBytes),
+					AppId:           appId,
```
bumi, line 624: I think this is wrong. wouldn't AppId now become 0?

rolznz (approved):
utACK

PR #465, opened 2024-08-13
fix: do not require jwt for oauth callback, disable service worker

rolznz:


http/alby_http_service.go
```diff
@@ -72,6 +72,7 @@ func (albyHttpSvc *AlbyHttpService) unlinkHandler(c echo.Context) error {
 }
 
 func (albyHttpSvc *AlbyHttpService) albyCallbackHandler(c echo.Context) error {
+	logger.Logger.Info("ALBY CALLBACK HANDLER HIT")
```
bumi, line 75: does this need to scream? :)

rolznz: oops, removed

PR #471, opened 2024-08-13
feat: add alby cloud link

rolznz:
Links to Alby Cloud for users who don't have an Alby Cloud hub deployment

image

### DIFF
--- a/frontend/src/components/layouts/AppLayout.tsx
+++ b/frontend/src/components/layouts/AppLayout.tsx
@@ -1,5 +1,6 @@
 import {
   Cable,
+  Cloud,
   EllipsisVertical,
   ExternalLinkIcon,
   Home,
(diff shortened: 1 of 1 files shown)


jpl-btc:
I would like you to consider having the "Alby Cloud" section link to an external landing page that would highlights the benefits of choosing Alby Cloud for users and for us. This approach allows us to update the landing page as needed, including running A/B tests to see which version converts better, without having to modify the self-hosted Hub's UX/UI every time we want to adjust the Alby Cloud landing page recommendation/invitation.

rolznz:
@jpl-btc do we have such a page? shouldn't there also be something like this at getalby.com?

jpl-btc:
Not that I am aware of. And yes there should be something like that at getalby.com @bumi wanted me to check on a landing page-making tool, maybe that is related to this apparent need to make some additional landing pages for Alby Flavor updates, recommendations, and other stuff

Issue #480, opened 2024-08-14
Bug: Bitcoin Connect QR code not showing due to CSP


rolznz:
![image](https://github.com/user-attachments/assets/edd3f3a3-54a5-443e-a3f5-2d2ea92f4d83)

PR #483, opened 2024-08-14
fix: bump bitcoin connect dependency to fix csp issue with invoice qr

rolznz:
Fixes https://github.com/getAlby/hub/issues/480


im-adithya (approved):
tACK

Issue #472, opened 2024-08-13
Unable to restore LDK state from v1.4.1 backup


jmanero:
I don't see any existing issues or PRs relating to the failure I observed, and wanted to raise this in case there's a lurking issue with the existing backup/restore code.

`TL;DR` I noticed that the `nwc` database is using WAL journaling, but the backup does not contain `wal` and `shm` files. Is it possible that the SQLite WAL was not checkpointed before adding the `nwc.sqlite` file to the backup ZIP? This would result in any pages still in the `WAL` (and the transactions that wrote them) being effectively dropped from the database backup. Note that simply closing the SQLite handle does not compact the WAL. I believe that a [PRAGMA wal_checkpoint(FULL)](https://www.sqlite.org/pragma.html#pragma_wal_checkpoint) has to be invoked to ensure that all pages are flushed to the database (`.sqlite`) file.

---
I'm unable to provision a new AlbyHub instance from a backup archive created on a `v1.4.1` instance using LDK. I've attempted to restore with both `v1.4.1` and `v1.5.1` with the same result:

```
{"event":{"event":"nwc_started","properties":{"version":"v1.4.1"}},"global":{},"level":"info","msg":"Publishing event","time":"2024-08-13T04:50:29Z"}
{"level":"info","msg":"Using existing Alby OAuth token","time":"2024-08-13T04:50:29Z"}
{"level":"info","msg":"Using existing Alby OAuth token","time":"2024-08-13T04:50:41Z"}

[... SNIP HTTP access logs ...]

Aug 13 04:59:34 mini-01.local albyhub[72076]: {"level":"info","msg":"Launching LN Backend: LDK","time":"2024-08-13T04:59:34Z"}
Aug 13 04:59:34 mini-01.local albyhub[72076]: {"level":"warning","msg":"No LDK RGS instance set","time":"2024-08-13T04:59:34Z"}
Aug 13 04:59:34 mini-01.local albyhub[72076]: {"level":"error","msg":"Failed to create LDK node: BuildError: KvStoreSetupFailed: Failed to setup KVStore.","time":"2024-08-13T04:59:34Z"}
Aug 13 04:59:34 mini-01.local albyhub[72076]: {"error":"BuildError: KvStoreSetupFailed: Failed to setup KVStore.","level":"error","msg":"Failed to launch LN backend","time":"2024-08-13T04:59:34Z"}
[...]

PR #485, opened 2024-08-14
fix: migrate node

rolznz:
Fixes https://github.com/getAlby/hub/issues/472

We weren't shutting down the DB correctly which didn't save the full write-ahead log to the DB. This was causing issues when importing - it was using the old -shm and -wal files which were causing the imported nwc.db to be discarded.

- ensure full wal checkpoint is executed on shutdown
- correctly shutdown DB on normal exit
- shutdown service before exiting on backup restore
- remove JWT middleware from backup endpoint
- delete old nwc db files before backup restore to ensure restored files are used

Tests:
- [x] Http Mode - Backup
- [x] Http Mode - Restore
- [x] Wails mode - Backup
- [x] Wails mode - Restore

### DIFF
--- a/api/backup.go
+++ b/api/backup.go
@@ -211,6 +211,15 @@ func (api *api) RestoreBackup(unlockPassword string, r io.Reader) error {
 
 	go func() {
 		logger.Logger.Info("Backup restored. Shutting down Alby Hub...")
+		api.svc.Shutdown()
+		// ensure no -shm or -wal files exist as they will stop the restore
+		for _, filename := range []string{"nwc.db", "nwc.db-shm", "nwc.db-wal"} {
+			err = os.Remove(filepath.Join(workDir, filename))
+			if err != nil {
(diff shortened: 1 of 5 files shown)


api/backup.go
```diff
@@ -211,6 +211,15 @@ func (api *api) RestoreBackup(unlockPassword string, r io.Reader) error {
 
 	go func() {
 		logger.Logger.Info("Backup restored. Shutting down Alby Hub...")
+		api.svc.Shutdown()
+		// ensure no -shm or -wal files exist as they will stop the restore
+		for _, filename := range []string{"nwc.db", "nwc.db-shm", "nwc.db-wal"} {
```
im-adithya, line 216: This should be `DATABASE_URI` `DATABASE_URI`-shm and `DATABASE_URI`-wal

rolznz: The backup and restore only works with the default directory structure (nwc.db must be within the WORK_DIR)

im-adithya (approved):
tACK

PR #490, opened 2024-08-14
chore: add other lnd methods

### LINKED ISSUE #30: Improve LND functionality
LND currently misses some methods implemented by LDK, we have to make sure it's up to speed

im-adithya:
Fixes #30 
~Waiting on #246 for `GetDebugInfo`~

Implements the remaining functions in LND. This change should make LND support everything LDK does.


lnclient/lnd/lnd.go
```diff
@@ -719,16 +740,14 @@ func (svc *LNDService) GetOnchainBalance(ctx context.Context) (*lnclient.Onchain
 	}, nil
 }
 
+// TODO: SendCoins
```
rolznz, line 743: Note: I think this should be prioritized over getting the network graph / debug methods

im-adithya: Yes this is not related to #246, thought of working on the lnd issue so I can add both at once so shifted to that, will add both now and open this for review 👍

lnclient/lnd/lnd.go
```diff
@@ -720,15 +742,34 @@ func (svc *LNDService) GetOnchainBalance(ctx context.Context) (*lnclient.Onchain
 }
 
 func (svc *LNDService) RedeemOnchainFunds(ctx context.Context, toAddress string) (txId string, err error) {
-	return "", nil
+	resp, err := svc.client.SendCoins(ctx, &lnrpc.SendCoinsRequest{
+		Addr:    toAddress,
+		SendAll: true,
```
rolznz, line 747: We have an issue in LDK right now that this also sends anchor reserves. Is this the case in LND?

lnclient/lnd/lnd.go
```diff
@@ -562,15 +564,35 @@ func (svc *LNDService) Shutdown() error {
...
+		PubKey: pubkey,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	addresses := nodeInfo.Node.Addresses
+
+	if addresses == nil || len(addresses) < 1 {
+		return nil, errors.New("no available listening addresses")
```
rolznz, line 578: we should not return an error here, otherwise the pubkey is also not returned

lnclient/lnd/lnd.go
im-adithya: Yes, tested SendCoins with 2,3,4 and 5 channels in my LND node with `SendAll` set to true. It still left 10k, 20k, 30k, 40k and 50k sats in each case

lnclient/lnd/lnd.go
im-adithya: Fixed to always return pubkey

rolznz (approved):
utACK

PR #491, opened 2024-08-14
chore: add breadcrumbs to peer and sign message screen

### LINKED ISSUE #449: Breadcrumbs for peers and sign message
Peers has no breadcrumbs, we can add it to channels to show "Node" > Peers but is it okay to have the link as `/#/channels/peers`?

Screenshot 2024-08-12 at 2 23 21 PM

Sign Message has wrong breadcrumbs, but it is the same issue again, if we wish to show Node instead of Wallet, we would have the link as `/#/channels/sign-message`

Screenshot 2024-08-12 at 2 21 19 PM

im-adithya:
Fixes #449

## Screenshots
Screenshot 2024-08-15 at 1 57 14 AM
Screenshot 2024-08-15 at 1 57 24 AM
Screenshot 2024-08-15 at 1 57 41 AM

### DIFF
--- a/frontend/src/screens/channels/Channels.tsx
+++ b/frontend/src/screens/channels/Channels.tsx
@@ -241,12 +241,12 @@ export default function Channels() {
                 <DropdownMenuGroup>
                   <DropdownMenuLabel>Management</DropdownMenuLabel>
                   <DropdownMenuItem>
-                    <Link className="w-full" to="/peers">
+                    <Link className="w-full" to="/channels/peers">
                       Connected Peers
                     </Link>
                   </DropdownMenuItem>
                   <DropdownMenuItem>
-                    <Link className="w-full" to="/wallet/sign-message">
+                    <Link className="w-full" to="/channels/sign-message">
                       Sign Message
                     </Link>
                   </DropdownMenuItem>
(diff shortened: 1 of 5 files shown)


frontend/src/screens/peers/Peers.tsx
```diff
@@ -87,7 +87,7 @@ export default function Peers() {
         description="Manage your connections with other lightning nodes"
         contentRight={
           <>
-            <Link to="/peers/new">
+            <Link to="/channels/peers/new">
```
rolznz, line 90: I don't think this makes sense. I just think we need to be careful if we decide to change the old routes for example (/channels to /node)

frontend/src/screens/channels/Channels.tsx
```diff
@@ -241,12 +241,12 @@ export default function Channels() {
...
                   <DropdownMenuLabel>Management</DropdownMenuLabel>
                   <DropdownMenuItem>
-                    <Link className="w-full" to="/peers">
+                    <Link className="w-full" to="/channels/peers">
                       Connected Peers
                     </Link>
                   </DropdownMenuItem>
                   <DropdownMenuItem>
-                    <Link className="w-full" to="/wallet/sign-message">
+                    <Link className="w-full" to="/channels/sign-message">
```
rolznz, line 249: I don't think this is right

im-adithya:
We decided to prioritze this issue later on as this changes the routes and `/channels/peers` doesn't make sense right now as it also (can) contain peers with whom we don't have channels as well (and `/channels/sign-message`)

PR #493, opened 2024-08-15
fix: csp issue on setup node page

rolznz:


bumi:
should we relax this CSP a bit?

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/494

rolznz:
> should we relax this CSP a bit?

Adithya found a better fix! :rocket:

PR #506, opened 2024-08-16
feat: generate macos http build in actions

### LINKED ISSUE #411: Create a mac HTTP build
As we area already able to make a build for mac, is it also possible to make a HTTP build for mac.
This allows running the alby hub server similar to running it on linux without the overhead of the Desktop app.

im-adithya:
Fixes #411

### DIFF
--- a/.github/workflows/http.yml
+++ b/.github/workflows/http.yml
@@ -194,7 +194,7 @@ jobs:
           cd ../../..
 
       - name: Notarize the zip file
-        if: matrix.build.os == 'macos-12'
+        if: matrix.build.os == 'macos-12' && inputs.build-release
         run: |
           echo "Notarizing Zip Files"
           gon -log-level=info -log-json ./build/darwin/http/gon-notarize.json
--- a/.github/workflows/http.yml
+++ b/.github/workflows/http.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Setup
         run: |
           echo "EXEC_NAME=albyhub" >> $GITHUB_ENV
-          if [ "${{ matrix.build.os }}" == "macos-12" ]; then
+          if [ "${{ runner.os }}" == "macOS" ]; then
             echo "PACKAGE_NAME=albyhub-Server-MacOS" >> $GITHUB_ENV
           else
             echo "PACKAGE_NAME=albyhub-Server-Linux-${{ matrix.build.arch }}" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


im-adithya:
Ready for review!

rolznz:
@im-adithya can you look at the other workflows? I think some of your conditionals are very fragile, which will break as soon as we change the runner.

rolznz:
The yaml config is twice as long now, maybe it's better we move this to a new workflow instead?

rdmitr:
> @im-adithya can you look at the other workflows? I think some of your conditionals are very fragile, which will break as soon as we change the runner.

Yep, it might be better to check `runner.os` instead of `matrix.build.os`. The former will assume OS-version-independent values (`Linux`, `macOS`).

> The yaml config is twice as long now, maybe it's better we move this to a new workflow instead?

I have mixed feelings about this, to be honest. On one hand, yeah, it's growing quite large; on another, there's still quite a lot of common code for macOS and Linux.

In fact, the wails workflow looks the same: long yaml, lots of conditionals. Maybe we should refactor both at some point, or split them into separate jobs for different target OSes. There will be some code duplication, but each job will be more concise, easier to read and maintain.

rolznz:
Cool. I think in that case let's leave it in this work flow. @im-adithya can you address Roman's suggestion:

> Yep, it might be better to check runner.os instead of matrix.build.os. The former will assume OS-version-independent values (Linux, macOS).

.github/workflows/http.yml
```diff
@@ -107,22 +147,74 @@ jobs:
...
+        run: |
+          install_name_tool -add_rpath @executable_path/../lib ./build/bin/${{ env.PACKAGE_NAME }}/bin/${{ env.EXEC_NAME }}
+
+      - name: Add permissions
         run: |
           chmod +x build/bin/${{ env.PACKAGE_NAME }}/bin/${{ env.EXEC_NAME }}
         shell: bash
 
+      - name: Sign the MacOS binary and libraries
+        if: runner.os == 'macOS'
```
rolznz, line 173: should we run this only on release, too?

im-adithya: I think we need code signing as it links the dev build to our org, which then uses the previously saved org's permission to run the dev build without additional permissions again (only when the release build is run before on the device)

rolznz (approved):

PR #511, opened 2024-08-19
fix: use workdir for temp bkp file

### LINKED ISSUE #391: Restore backup on Mac Desktop does not work
image

im-adithya:
Fixes #391

I failed to see the "failed to create decrypted reader" error as shown in the issue comment, but it should be good as I tested it twice (and saw only the below issue)

This PR solves the issue which gives the "failed to create temporary output file: operation not permitted" error because the directory being used is outside the permitted directory

## Tests

- [x] HTTP - Wails
- [x] HTTP - HTTP
- [x] Wails - HTTP
- [x] Wails - Wails


api/backup.go
```diff
@@ -148,7 +148,7 @@ func (api *api) RestoreBackup(unlockPassword string, r io.Reader) error {
 		return fmt.Errorf("failed to create decrypted reader: %w", err)
 	}
 
-	tmpF, err := os.CreateTemp("", "nwc-*.bkp")
+	tmpF, err := os.CreateTemp(api.cfg.GetEnv().Workdir, "nwc-*.bkp")
```
bumi, line 151: should the file be renamed to `albyhub-` ?

and maybe we can make a sub directory?

.github/workflows/wails.yml
```diff
@@ -200,7 +200,7 @@ jobs:
           mv ./build/out/${{ env.PACKAGE_NAME }}.dmg ./build/bin/
 
       - name: Notarize the DMG image
-        if: runner.os == 'macOS' && inputs.build-release
+        if: runner.os == 'macOS'
```
rolznz, line 203: was this intentional?

im-adithya: Yes, we have to notarize this everytime, otherwise only signing wouldn't work (unless you already opened the app before) And we have a limit of 70 notarizations per day (with our apple dev account) so this should be good atleast for now

api/backup.go
im-adithya: Yes I wanted to change this to albyhub- as well, also the backup file. But would it have any side effects? cc @rolznz

.github/workflows/wails.yml
rolznz: @im-adithya but people only run the release builds, don't they?

api/backup.go
rolznz: aren't you creating a brand new folder? why are there side effects?

rolznz: Oh. This temp file is only used in this backup process. So if you rename it here, just check it is not used elsewhere

.github/workflows/wails.yml
im-adithya: Yeah but I thought if someone of us is testing for the first time, they might see this issue. But on second thoughts it might hit the 70 limit someday and might create problem during release, I'll add this back 👍

rolznz:
@im-adithya what tests did you do for this?

>  but it should be good as I tested it twice (and saw only the below issue)

what was the issue?

im-adithya:
> what was the issue?

I was referring to the screenshot you posted in #391

`failed to read salt: EOF`

But it is an old version so i think this fix should solve the problem

api/backup.go
rolznz: @im-adithya you didn't rename the file?

rolznz:
@im-adithya can you re-review your code and test it again

rolznz (approved):
tACK (Linux)

PR #517, opened 2024-08-20
fix: ldk use info log level by default

rolznz:
DEBUG can cause a lot of spam messages and the log files to get too large.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,7 @@ AUTO_LINK_ALBY_ACCOUNT=false
 # Optionally LDK debug log level to get more info
 #LDK_LOG_LEVEL=2
 # Logrus debug log level
-LOG_LEVEL=5
+LOG_LEVEL=4
 
 #WORK_DIR=.data
 #DATABASE_URI=nwc.db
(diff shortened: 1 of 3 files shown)


rolznz:
Thanks @matjaz !

.env.example
```diff
@@ -4,7 +4,7 @@ LOG_EVENTS=false
 # do not link your current account when you run a dev instance (so it stays pointing at your mainnet one)
 AUTO_LINK_ALBY_ACCOUNT=false
 
-# LDK trace log level
+# Optionally LDK debug log level to get more info
 #LDK_LOG_LEVEL=2
 # Logrus debug log level
 LOG_LEVEL=5
```
rolznz, line 10:
```suggestion
LOG_LEVEL=4
```

rolznz:
> also you should update this file https://getalby.com/install/hub/fly.toml

Edit: it's not set there, so it'll use the default value from Alby Hub which is fixed in this PR.

fly.toml
```diff
@@ -12,7 +12,7 @@ swap_size_mb = 2048
 
 [env]
   DATABASE_URI = '/data/nwc.db'
-  LDK_LOG_LEVEL = '2'
+  LDK_LOG_LEVEL = '3'
   LOG_LEVEL = '5'
```
matjaz, line 16: this should be 4? same as in .env ?

rolznz: image

Info level is the one we want by default, which is 3

matjaz: in `.env.example` `LOG_LEVEL` is set to 4. in fly it's 5.

PR #520, opened 2024-08-21
feat: add send to lightning address

### LINKED ISSUE #436: Support paying to lightning addresses from Alby Hub Wallet
currently we can only pay BOLT11 invoices

im-adithya:
## Description
Fixes #436 

## Screenshots
<p>

</p>

## Tests
- [x] Check with `apple23` (random string)
- [x] Check with `lnbc34567` (invalid bolt11)
- [x] Check with expired bolt11
- [x] Check with a@a.com (invalid lightning address)
- [x] Check with comment length more than `commentAllowed` while sending to LN address
- [x] Check alerts when amounts are greater than balance


frontend/vite.config.ts
```diff
@@ -82,7 +82,7 @@ const insertDevCSPPlugin: Plugin = {
         "<head>",
         `<head>
         <!-- DEV-ONLY CSP - when making changes here, also update the CSP header in http_service.go (without the nonce!) -->
-        <meta http-equiv="Content-Security-Policy" content="default-src 'self' ${DEVELOPMENT_NONCE}; img-src 'self' https://uploads.getalby-assets.com https://getalby.com;"/>`
+        <meta http-equiv="Content-Security-Policy" content="default-src 'self' ${DEVELOPMENT_NONCE}; img-src 'self' https://uploads.getalby-assets.com https://getalby.com; connect-src 'self' https://api.getalby.com"/>`
```
im-adithya, line 85: This enables fetching the invoice data from ln address via our [ln-proxy](https://github.com/getAlby/lightning-address-details-proxy)

rolznz: @im-adithya can you check the comment above this, we also need to change the CSP header in http_service.go

im-adithya: Whoops sorry, completely missed it, adding now 👍

frontend/src/screens/wallet/Send.tsx
```diff
@@ -16,98 +10,123 @@ import {
...
-import { copyToClipboard } from "src/lib/clipboard";
-import { PayInvoiceResponse } from "src/types";
-import { request } from "src/utils/request";
+
+// email regex: https://emailregex.com/
+// modified to allow _ in subdomains
+const LIGHTNING_ADDRESS_REGEX =
+  /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-_0-9]+\.)+[a-zA-Z]{2,}))$/;
+
+const BOLT11_REGEX = /^(lnbc|lntbs)([0-9a-z]+)$/i;
```
rolznz, line 29: where did you get this from? does this support standard testnet too or just mutinynet?

frontend/src/screens/wallet/Send.tsx
```diff
@@ -16,98 +10,123 @@ import {
...
 import { useChannels } from "src/hooks/useChannels";
 
 import { useInfo } from "src/hooks/useInfo";
-import { copyToClipboard } from "src/lib/clipboard";
-import { PayInvoiceResponse } from "src/types";
-import { request } from "src/utils/request";
+
+// email regex: https://emailregex.com/
+// modified to allow _ in subdomains
+const LIGHTNING_ADDRESS_REGEX =
```
rolznz, line 26: I think this is in a few projects now, I wonder if it should be moved to something like `lightning-tools`

frontend/src/screens/wallet/Send.tsx
```diff
@@ -16,98 +10,123 @@ import {
...
+            <Button
+              type="button"
+              variant="outline"
+              className="px-2"
+              onClick={paste}
+            >
+              <ClipboardPaste className="w-4 h-4" />
+            </Button>
+          </div>
+          <Button type="submit" disabled={!recipient}>
```
rolznz, line 124: Can you change this to a loading button? loading the LNURL data is not instant

frontend/src/screens/wallet/Send.tsx
rolznz: I don't think we need this actually - we can just wrap creating the invoice in a try/catch and if it fails then we know it's not a valid invoice. Then this is handled by lightning-tools.

frontend/src/screens/wallet/Send.tsx
```diff
@@ -16,98 +10,123 @@ import {
...
-        setPaymentDone(true);
-        setInvoice("");
-        toast({
-          title: "Successfully paid invoice",
-        });
+        setLnurlDetails(lnAddress);
+      } else if (BOLT11_REGEX.test(recipient)) {
+        const invoice = new Invoice({ pr: recipient });
+        setAmount(invoice.satoshi);
+        setLnurlDetails(invoice);
```
rolznz, line 61: hmm, the invoice is not an LNURL, it's a bolt11 invoice. At the least we should rename this.

frontend/src/components/wallet/BalanceCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+} from "src/components/ui/card";
+import { BalancesResponse } from "src/types";
+
+type BalanceCardProps = {
+  balances: BalancesResponse;
+  hasChannelManagement: boolean;
+  isSpending?: boolean;
+};
+
+function BalanceCard({
```
rolznz, line 19: This was unfortunately a poor refactoring and introduced at least 2 bugs, and now requires a lot of conditionals.

You could pass args to the component rather than using conditionals, and you didn't even add them all (for example the arrow icon is not right for receive, now the receiving card is shown for LN clients that do not support channel management, ...

and this is only what I spotted from looking at the code, there may be more

frontend/src/components/wallet/LnurlPay.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const [comment, setComment] = React.useState("");
+  const [preimage, setPreimage] = React.useState("");
+  const [isLoading, setLoading] = React.useState(false);
+  const { toast } = useToast();
+
+  const onSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    try {
+      setLoading(true);
+      const invoice = await lnAddress.requestInvoice({
```
rolznz, line 30: is there a way to not duplicate the invoice payment logic? you already pass callbacks as props like onReset, could there be one to pay an invoice?

frontend/src/screens/wallet/Send.tsx
```diff
@@ -16,98 +10,123 @@ import {
...
+  if (!balances || !channels) {
+    return <Loading />;
+  }
+
+  const hasLowBalance =
+    hasChannelManagement &&
+    (amount || 0) * 1000 >= 0.8 * balances.lightning.totalSpendable;
+
+  const renderScreen = () => {
+    if (lnurlDetails instanceof LightningAddress) {
```
rolznz, line 90: the flow seems kind of weird here, that the payment success cards are rendered in the individual components. A success is a success, it doesn't change based on which payment type, right?

frontend/src/screens/wallet/Send.tsx
```diff
@@ -116,156 +135,25 @@ export default function Send() {
...
-              </Link>
-            </AlertDescription>
-          </Alert>
-        )}
+      {hasLowBalance && (
+        <Alert>
+          <AlertTriangle className="h-4 w-4" />
+          <AlertTitle>Low spending balance</AlertTitle>
+          <AlertDescription>
+            You won't be able to make payments until you{" "}
```
rolznz, line 143: I think we should remove this alert - we decided to only show the alerts about receiving capacity, as the spending ones are a bit unnecessary. I must have missed this one when removing the other alerts

rolznz:
@im-adithya nice work and good attempt to split up the components, but please be careful when refactoring and consider the implications.

I also think there are a few strange things in how the payment components are structured, which I have commented on, but overall I think it is in a good direction.

frontend/src/components/wallet/BalanceCard.tsx
im-adithya: > now the receiving card is shown for LN clients that do not support channel management

I thought it would be useful to see the capacity and show the increase button only if there is channel management like we do in the Send card. Is there any reason why we show the spending card without the button and don't show it at all for receiving card?

And sorry, I missed the icon, let me know if the above is okay and depending on that I'll revert the component or add the icon (if it's okay to have another conditional)

frontend/src/screens/wallet/Send.tsx
im-adithya: Umm, it can be both an Invoice or Lightning Address so I used the name... I'll check your other comments and refactor this a bit and will see what we can do here :+1:

frontend/src/components/wallet/BalanceCard.tsx
rolznz: Let's chat about this one on a call

frontend/src/screens/wallet/Send.tsx
rolznz: But an invoice is not an LNURL, why did you choose this name?

im-adithya:
`frontend/src/screens/wallet/Send` folder should be renamed to `frontend/src/screens/wallet/send` for the tests to pass. I can't do it for some reason :(

frontend/src/components/layouts/SendLayout.tsx
```diff
@@ -0,0 +1,47 @@
...
+        description="Pay a lightning invoice created by any bitcoin lightning wallet"
+      />
+      <div className="flex gap-12 w-full">
+        <div className="w-full max-w-lg">
+          <Outlet />
+        </div>
+        <BalanceCard
+          balance={
+            balances
+              ? new Intl.NumberFormat(undefined, {}).format(
```
rolznz, line 33: can you format in the BalanceCard instead of doing it here?

frontend/src/screens/wallet/send/LnurlPay.tsx
```diff
@@ -0,0 +1,121 @@
...
+      setLoading(true);
+      const lnAddress = new LightningAddress(lnurlDetails.identifier);
+      // this is set instead of calling fetch because
+      // requestInvoice uses lnurlpData.callback
+      lnAddress.lnurlpData = lnurlDetails;
+      const invoice = await lnAddress.requestInvoice({
+        satoshi: parseInt(amount),
+        comment,
+      });
+      const payInvoiceResponse = await request<PayInvoiceResponse>(
```
rolznz, line 35: This payment is duplicated with confirm payment. We could route to the ConfirmPayment page with the invoice instead - then the user can also confirm (make sure they didn't accidentally type the wrong amount)

frontend/src/screens/wallet/Send.tsx
```diff
@@ -1,272 +1,111 @@
...
+        navigate(`/wallet/send/confirm-payment`, {
+          state: {
+            args: { paymentRequest: invoice },
+          },
+        });
+      } catch (error) {
         toast({
-          title: "Successfully paid invoice",
+          variant: "destructive",
+          title: "Failed to send payment",
```
rolznz, line 57: this is the wrong error message, right?

frontend/src/screens/wallet/send/LnurlPay.tsx
im-adithya: Hmmm, yeah we can do this to DRY it up even more!

PR #524, opened 2024-08-21
chore: add app id to CreateAppRepsonse

### LINKED ISSUE #460: Return app ID in create app response
Currently you need to do another call to fetch the app by pubkey to get it

im-adithya:
Fixes #460

Returns app id in `CreateAppRepsonse`

Screenshot 2024-08-21 at 7 33 37 PM


rolznz (approved):
utACK

PR #525, opened 2024-08-21
chore: use usetoast hook everywhere

### LINKED ISSUE #450: Fix toast usage
We should remove the `toast` export from `export { toast, useToast };` and ensure we use `useToast` everywhere - that includes additional logic.

im-adithya:
Fixes #450

`copyToClipboard` function is now the only place where we use `toast`. I think the hook here is not doing anything of much use, because we are always importing `toast` from it. The re-render (when using `useToast`) which happens on adding a new toast due to state change is not of use to us, or am I wrong?


rolznz (approved):
tACK

rolznz:
@im-adithya I passed the toast function into `copyToClipboard` so we don't have to export it and run into this issue again in the future

PR #534, opened 2024-08-23
fix: increase max dust htlc exposure from fee rate multiplier for trusted counterparties

rolznz:
Stops FC if counterparty fee rates are too high.

:warning: I haven't been able to reliably test this. Waiting for LDK to reply...

We have another workaround in place which bumps the fee estimates for the OnchainSweep block target.


rolznz:
This can be implemented once https://github.com/getAlby/ldk-node/pull/53 is merged.

rolznz:
This was implemented in another PR

Issue #396, opened 2024-08-04
Better error handling on setup


bumi:
what happens when invalid data is entered or when the requests fail. 
We have to make sure it can be retried.

rolznz:
@bumi can you add some more context? this is when the user uses a different node type or advanced setup, right?

bumi:
yes. I kept it open because I think we should review all error cases. 

so for example if the internet connection is not working then I am stuck. it says I already did the setup when I retry.

rolznz:
> so for example if the internet connection is not working then I am stuck. it says I already did the setup when I retry.

Note: this is a protection mechanism against not allowing a user's node setup to be overwritten. (maybe the node should be able to be setup again if the node has never been successfully started)

rolznz:
From ek:
> I subscribed to Alby Hub and I think trying to connect to a local node using "localhost" broke it (wanted to see if it would work). After I was done waiting for it to connect, I clicked "back" and tried to use a Cashu mint but then it said "Setup already finished".

PR #543, opened 2024-08-24
feat: better error handling on setup

rolznz:
Fixes https://github.com/getAlby/hub/issues/396

When running 1.6.0 the first time, the SetupCompleted check will rely on the user having an `AlbyUserIdentifier` or an existent LDK directory. This is not perfect, but I think the chance both are not set is very slim - and if a user does encounter this, we can give them instructions on how to do a db edit to set their user ID.

- finish can now be run again if node has never successfully started
- fix: use mutex for both setup and start endpoints
- fix: reduce setup finish timeout to 30 seconds from 3 minutes
- fix: cashu start does not handle invalid mints

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -658,20 +658,20 @@ func (api *api) SetNextBackupReminder(backupReminderRequest *BackupReminderReque
 var startMutex sync.Mutex
 
 func (api *api) Start(startRequest *StartRequest) error {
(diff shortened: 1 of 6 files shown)


rolznz:
Related issue: https://github.com/getAlby/hub/issues/545

PR #554, opened 2024-08-27
chore: secure password label id

im-adithya:
Clicking on "I understand this password cannot be recovered" toggles the above checkbox which feels buggy


rolznz:
@im-adithya btw: this should be a "fix" not a "chore" right?

im-adithya:
😬 I first named it a fix but then thought it might be a chore as it's just a typo

PR #560, opened 2024-08-28
feat: add lud16

matjaz:


rolznz:
Hi @matjaz 

Thanks for the PR!

For context, this was used so when new Damus users created a wallet (and brand new Alby Account) it would immediately return the lightning address to Damus to save it in the profile.

But this is not easily possible now because there is no seamless connection flow for Damus users to create a new alby account and hub on mobile.

For that reason, I am not sure if it is worthwhile making this change right now until we address a few of these things. CC @bumi

rolznz:
@bumi we currently do not cache it.

We could cache it in the Alby OAuth Service in memory each time we request the user info, and expose a Getter there. The API has access to the Alby OAuth Service so could easily request it when returning the response from CreateApp().

This is not a perfect solution, but might be enough in this case?

api/models.go
```diff
@@ -85,14 +85,15 @@ type UpdateAppRequest struct {
...
-	Isolated      bool     `json:"isolated"`
+	Name             string   `json:"name"`
+	Pubkey           string   `json:"pubkey"`
+	MaxAmountSat     uint64   `json:"maxAmount"`
+	BudgetRenewal    string   `json:"budgetRenewal"`
+	ExpiresAt        string   `json:"expiresAt"`
+	Scopes           []string `json:"scopes"`
+	ReturnTo         string   `json:"returnTo"`
+	Isolated         bool     `json:"isolated"`
+	LightningAddress string
```
rolznz, line 96: Interesting you also allow the lightning address to be passed from the frontend, is there a usecase for this?

matjaz: The easiest way was not to change CreateApp args.

It doesn't have json field so no. Also in http service it's always set to empty string first. https://github.com/getAlby/hub/pull/560/files#diff-3d9fb3ef05f4794887b3f5b899cf421382e8a466f4b33692389ed2af9ad31728R851

rolznz: Ok. Unless there is a clear usecase I would not add this property to the request payload

rolznz:
I realized this would also be useful for Alby Mobile (which is being released soon) to instantly be setup with the lightning address when connecting to Alby Hub

rolznz:
Related: https://github.com/getAlby/js-sdk/pull/235

matjaz:
I have refactored to separate argument

rolznz:
I simplified this and decided to store the lightning address in the user config. @bumi are you ok with this method?

matjaz:
cache (busting) is the second hardest thing in software, right after naming variables. :)
you should also handle `UnlinkAccount`

bumi (approved):

rolznz:
> cache (busting) is the second hardest thing in software, right after naming variables. :) you should also handle `UnlinkAccount`

True - this may become relevant once the Alby Account becomes optional.

api/api.go
```diff
@@ -89,24 +89,29 @@ func (api *api) CreateApp(createAppRequest *CreateAppRequest) (*CreateAppRespons
...
 			responseBody.ReturnTo = returnToUrl.String()
 		}
 	}
 
 	var lud16 string
-	// if user.LightningAddress != "" {
-	// 	lud16 = fmt.Sprintf("&lud16=%s", user.LightningAddress)
-	// }
+	if lightningAddress != "" {
+		lud16 = fmt.Sprintf("&lud16=%s", lightningAddress)
```
matjaz, line 113: @rolznz what would be the best option to add lud16 for sub-wallets? 

(Adding connection to Alby go.)

rolznz: @matjaz could you check this PR? https://github.com/getAlby/hub/pull/1556

PR #571, opened 2024-08-29
feat: show startup error

### LINKED ISSUE #545: Return last start error in info call
Since start triggers a goroutine and immediately returns, it cannot return the error to the frontend.

It would be nice to be able to set an error, and reset it when the user tries again.

im-adithya:
Fixes #545 

## Login Screen (I started the node in Polar the third time)
https://github.com/user-attachments/assets/a986347d-dab4-497a-aee9-830dc2f24a57

## Setup Finish Screen
https://github.com/user-attachments/assets/15e445e4-6d4d-48df-9ddf-b044f8163668

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -38,6 +38,7 @@ type api struct {
 	permissionsSvc permissions.PermissionsService
 	keys           keys.Keys
 	albyOAuthSvc   alby.AlbyOAuthService
+	startupError   error
 }
 
 func NewAPI(svc service.Service, gormDB *gorm.DB, config config.Config, keys keys.Keys, albyOAuthSvc alby.AlbyOAuthService, eventPublisher events.EventPublisher) *api {
@@ -605,9 +606,8 @@ func (api *api) GetInfo(ctx context.Context) (*InfoResponse, error) {
 	info := InfoResponse{}
 	backendType, _ := api.cfg.Get("LNBackendType", "")
 	info.SetupCompleted = api.cfg.SetupCompleted()
-	startupError := api.svc.GetStartupError()
-	if startupError != nil {
-		info.StartupError = api.svc.GetStartupError().Error()
+	if api.startupError != nil {
+		info.StartupError = api.startupError.Error()
 	}
 	info.Running = api.svc.GetLNClient() != nil
 	info.BackendType = backendType
@@ -662,9 +662,9 @@ func (api *api) SetNextBackupReminder(backupReminderRequest *BackupReminderReque
 var startMutex sync.Mutex
 
 func (api *api) Start(startRequest *StartRequest) (err error) {
-	api.svc.SetStartupError(nil)
+	api.startupError = nil
 	defer func() {
-		api.svc.SetStartupError(err)
+		api.startupError = err
 	}()
 	if !startMutex.TryLock() {
 		// do not allow to start twice in case this is somehow called twice
--- a/api/api.go
+++ b/api/api.go
@@ -605,6 +605,10 @@ func (api *api) GetInfo(ctx context.Context) (*InfoResponse, error) {
 	info := InfoResponse{}
 	backendType, _ := api.cfg.Get("LNBackendType", "")
 	info.SetupCompleted = api.cfg.SetupCompleted()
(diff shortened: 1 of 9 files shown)


api/api.go
```diff
@@ -657,13 +661,19 @@ func (api *api) SetNextBackupReminder(backupReminderRequest *BackupReminderReque
 
 var startMutex sync.Mutex
 
-func (api *api) Start(startRequest *StartRequest) error {
+func (api *api) Start(startRequest *StartRequest) (err error) {
+	api.svc.SetStartupError(nil)
```
rolznz, line 665: does it need to be on the service? or can it be a variable in the API? it seems it's only used here?

frontend/src/screens/Start.tsx
```diff
@@ -24,10 +24,31 @@ export default function Start() {
   const [unlockPassword, setUnlockPassword] = React.useState("");
   const [loading, setLoading] = React.useState(false);
   const [buttonText, setButtonText] = React.useState("Login");
-  useInfo(true); // poll the info endpoint to auto-redirect when app is running
+  const [startupError, setStartupError] = React.useState("");
```
rolznz, line 27: this is just a string value, there is no need to use state for it, we can use it directly from `info.startupError`

frontend/src/screens/Start.tsx
```diff
@@ -24,10 +24,31 @@ export default function Start() {
...
+
+  const { data: info, isValidating } = useInfo(true); // poll the info endpoint to auto-redirect when app is running
 
   const { toast } = useToast();
 
+  React.useEffect(() => {
+    if (!isValidating && info?.startupError) {
```
rolznz, line 34: This isValidating check feels hacky. I don't think it's necessary?

frontend/src/screens/Start.tsx
```diff
@@ -54,7 +75,7 @@ export default function Start() {
       setButtonText("Login");
       setUnlockPassword("");
       return;
-    }, 180000); // wait for 3 minutes
+    }, 30000); // wait for 30 seconds
```
rolznz, line 78: Note: this is different than the setup page, because the user has more data and syncing takes longer. But it should still be ok

frontend/src/screens/Start.tsx
im-adithya: We need it actually, because we want to check everytime info is updated (and because we do it every 3 seconds and sometimes the error is not updated and info stays same, the useEffect hook is not invoked)

im-adithya: > error is not updated

An example: suppose we try and fail (i.e. `info.startupError` is "connection error") on 1st second (assume info call is done on the same second, so the next will happen on 4s). Then we try again on 2s and as soon as we do the call, `SetStartupError(nil)` is called in the backend and usually it takes less than 3s for the call to fail, in that case the `info.startupError` would again be "connection error" (assuming the edge case where user changes nothing and tries again) and this would be set using `SetStartupError(err)` on 3s and when on 4s the info call is made, the info is the exact same although `info.startupError` changed from "connection error" to nil to "connection error" in between the 3s polling

frontend/src/screens/Start.tsx
im-adithya: We need this state variable because of [this](https://github.com/getAlby/hub/pull/571#discussion_r1741494171)

api/api.go
im-adithya: No I just thought it belongs there as it is startup information, keeping it in the api is actually better in terms of code as we won't need setters and getters, changing it 👍

api/api.go
```diff
@@ -657,13 +661,19 @@ func (api *api) SetNextBackupReminder(backupReminderRequest *BackupReminderReque
 
 var startMutex sync.Mutex
 
-func (api *api) Start(startRequest *StartRequest) error {
+func (api *api) Start(startRequest *StartRequest) (err error) {
+	api.startupError = nil
+	defer func() {
```
rolznz, line 666: I think this is not a clean way to do it with this defer and tracking setting the error variable everywhere.

rolznz (approved):
tACK

PR #572, opened 2024-08-30
fix: ensure auth with correct user if refresh token expires

rolznz:
Tests:
- [x] delete access & refresh token from DB -> triggers re-auth with the identifier as a query param
- [x] force logout from the settings page -> passes force logout param with no user identifier

### DIFF
--- a/frontend/src/screens/alby/AlbyAuthRedirect.tsx
+++ b/frontend/src/screens/alby/AlbyAuthRedirect.tsx
@@ -11,7 +11,17 @@ export default function AlbyAuthRedirect() {
   const queryParams = new URLSearchParams(location.search);
   const forceLogin = !!queryParams.get("force_login");
   const url = info?.albyAuthUrl
-    ? `${info.albyAuthUrl}${forceLogin ? "&force_login=true" : ""}`
+    ? (() => {
+        const _url = new URL(info.albyAuthUrl);
+        if (forceLogin) {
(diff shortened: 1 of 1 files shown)


frontend/src/screens/alby/AlbyAuthRedirect.tsx
```diff
@@ -11,7 +11,17 @@ export default function AlbyAuthRedirect() {
...
   const url = info?.albyAuthUrl
-    ? `${info.albyAuthUrl}${forceLogin ? "&force_login=true" : ""}`
+    ? (() => {
+        const _url = new URL(info.albyAuthUrl);
+        if (forceLogin) {
+          _url.searchParams.append("force_login", "true");
+        }
+        if (info.albyUserIdentifier) {
+          _url.searchParams.append("identifier", info.albyUserIdentifier);
+        }
```
bumi, line 21: this should be enough. if the user is logged in with that user then we don't need to prompt the login. so you can now emove the force login.?

rolznz: This is a different case. The force login is when the user explicitly wants to change the account they use for the hub

bumi: ahh, true. 👍

bumi (approved):

PR #576, opened 2024-08-30
feat: show pending payments in transactions list

### LINKED ISSUE #426: Show pending payments
This will help prevent users trying multiple times, (e.g. generating 2 invoices because the user thought the first payment failed - and then the user pays twice)

im-adithya:
Fixes #426

Currently needs design input, didn't add much there as it would result in conflicts with Roland's (upcoming?) transaction list changes in #570

### DIFF
--- a/transactions/transactions_service.go
+++ b/transactions/transactions_service.go
@@ -35,7 +35,7 @@ type TransactionsService interface {
 	events.EventSubscriber
 	MakeInvoice(ctx context.Context, amount int64, description string, descriptionHash string, expiry int64, metadata map[string]interface{}, lnClient lnclient.LNClient, appId *uint, requestEventId *uint) (*Transaction, error)
 	LookupTransaction(ctx context.Context, paymentHash string, transactionType *string, lnClient lnclient.LNClient, appId *uint) (*Transaction, error)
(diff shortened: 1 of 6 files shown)


transactions/transactions_service.go
```diff
@@ -474,7 +474,7 @@ func (svc *transactionsService) LookupTransaction(ctx context.Context, paymentHa
 	return &transaction, nil
 }
 
-func (svc *transactionsService) ListTransactions(ctx context.Context, from, until, limit, offset uint64, unpaid bool, transactionType *string, lnClient lnclient.LNClient, appId *uint) (transactions []Transaction, err error) {
+func (svc *transactionsService) ListTransactions(ctx context.Context, from, until, limit, offset uint64, pending, unpaid bool, transactionType *string, lnClient lnclient.LNClient, appId *uint) (transactions []Transaction, err error) {
```
rolznz, line 477: this introduces another bool and now there is the complexity you see below.

What about we have a `stateType` which and use that. the default would be `TRANSACTION_STATE_SETTLED`

rolznz: But this is problematic because we only want to see pending outgoing payments. This needs more thought. Maybe a function that can alter the query could be passed in :thinking:

rolznz: This is more complicated than I thought, let's move it to the next milestone

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/682

PR #580, opened 2024-09-02
feat: add buzzpay internal app

rolznz:
~~This uses the app metadata to store the readonly app connection :exploding_head: so that you can open the same PoS from any device~~

I'm not sure if it should be isolated or not?


rolznz:
@bumi edit: I'm not sure about doing it this way. It's convenient but we do not really want to store the secret like this. It might be better not not save it to the metadata.

bumi:
I don't think we should store the secret but just use the app to create the initial link

frontend/src/screens/internal-apps/BuzzPay.tsx
```diff
@@ -0,0 +1,145 @@
...
+          }
+        );
+
+        if (!createAppResponse) {
+          throw new Error("no create app response received");
+        }
+
+        const app = await request<App>(
+          `/api/apps/${createAppResponse.pairingPublicKey}`
+        );
```
bumi, line 68: we can abstract this away a bit into a `createApp` `updateApp` and `fetchApp` function maybe. 
in the future it makes it easier for such apps to be build

bumi (commented):
great

we need to explain more here how it works and what users can expect and what they need to do.

rolznz:
image

image

image

frontend/src/screens/internal-apps/BuzzPay.tsx
rolznz: done - see `requests/createApp`. Updating was removed from this PR, and fetching is a one-liner using SWR

PR #584, opened 2024-09-02
chore: better handle ports on Pi and allow albyhub to run on port 80

bumi:


rolznz (approved):
utACK

scripts/pi-arm/install.sh
```diff
@@ -12,16 +12,29 @@ wget https://getalby.com/install/hub/server-linux-armv6.tar.bz2
...
+if [[ $? -ne 0 ]]; then
   echo "Failed to unpack Alby Hub. Potentially bzip2 is missing"
   echo "Install it with sudo apt-get install bzip2"
 fi
 
 # Cleanup
 rm server-linux-armv6.tar.bz2
 
+# allow albyhub to bind on port 80
+sudo setcap CAP_NET_BIND_SERVICE=+eip /opt/albyhub/bin/albyhub
```
rolznz, line 24: should this be optional? like how we ask to setup a service?

bumi: any option makes it more complicated and more fragile. Those install scripts suck anyway because everybody wants it to be a bit different and systems are often different. 

People also need to know a little bit of what they do. 

IF we don't run as root then as far as I understand we need this. otherwise we need to run as root or on a different port.

Issue #114, opened 2024-05-11
Maintain Local Channels database


rolznz:
From @bumi :

We need to save all the required data in our DB and define single API calls to get all data.
Right now for example we do many calls to mempool triggered from the frontend to load the alias (not sure about rate limits there)

For example we can have a DB with channels that contains all the information from funding transaction to node alias and then have one GET /channels call to get this information from there. If I want information about the channel then I have a model that contains that data. right now if I want to get the node alias I would somehow have to go through this mempool API.

From Roland:

Also, we can save channel liquidity and other info that will be useful when events happen (like channel closures).

I would like to see an email notification or nostr DM when I get a channel closure:

"Your 500,000 sat channel with Olympus was closed. Reason: XYZ. You had 200,000 sats outgoing capacity when the channel was closed and these sats will be returned to your Alby Hub savings account within N blocks (about 2 weeks).

What to do next: `<Instructions here>`"

rolznz:
Feedback from @JPLACLAU 

- when a channel is forced closed maybe we should leave a message under the Savings Balance that says "x amount of sats are returning to your Savings Balance before x day. this was due to a force closed channel"
- At last: a channel forced closed, simply disappears on the ux an this creates some confusion: what happened to that channel? maybe we should leave the las state of the channel but all sort of greyed out with a warning saying the channels was forced closed, or some sort of warning inside the hub.

rolznz:
Related: https://github.com/getAlby/hub/issues/105

rolznz:
- Show closed channels on the standard table (the status would be "closed" or "closing")
- Show the balance of the closed channel as pending balance on the savings balance card (maybe an info tooltip to show that it might take up to 2 weeks)
- For channels older than 2 weeks, have a separate page (accessible from the advanced dropdown)

rolznz:
As part of this ticket, also reduce LDK sync time if a channel is currently opening

rolznz:
Removing the milestone for now - we made a significant improvement with https://github.com/getAlby/hub/pull/602

PR #590, opened 2024-09-02
[WIP] feat: display recently closed channels

rolznz:
Partially addresses https://github.com/getAlby/hub/issues/114

- [ ] pass a query param to include closed channels so it doesn't break other parts of the app which expect only open channels
- [ ] Use pending_balances_from_channel_closures to get the local channel balance for the closed channel
  - [ ] research how to get the local balance in LND - maybe the event is sufficient
- [ ] Consume Close Channel Events to get the reason for the closure and so we only show "closing" state for force closures

There is currently a big downside: we do not know how much local balance the channel has at time of closing.

And we do not check if the funding transaction funds have moved, so we also cannot show the balance of the closed channel as pending balance on the savings balance card.

For channels older than 2 weeks, these are currently not shown.

image

image

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -309,21 +309,19 @@ func (api *api) ListChannels(ctx context.Context) ([]Channel, error) {
 	if api.svc.GetLNClient() == nil {
 		return nil, errors.New("LNClient not started")
 	}
-	channels, err := api.svc.GetLNClient().ListChannels(ctx)
+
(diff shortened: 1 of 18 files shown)


rolznz:
I think instead we should do something more generalized like the events table idea https://github.com/getAlby/hub/issues/848

PR #594, opened 2024-09-03
chore: balance migration card follow up

### LINKED ISSUE #567: Alby shared balance migration card follow up tasks
From https://github.com/getAlby/hub/pull/555

- Update the icon as the user is not "moving away from Alby" - they are just taking custody of funds.
- Update copy: "still hosted by Alby" -> "still on your Alby shared wallet."
- Remove onboarding slide because it is irrelevant to users who do not have shared funds
- Use "Migrate" instead of "Transfer" on Node page

im-adithya:
Fixes #567

- Removes slide from onboarding
- Icon is replaced with ArrowLeftRight
- "still hosted by Alby" -> "still on your Alby shared wallet."
- Button text changed to "Migrate" from "Transfer"


rolznz (approved):
tACK

Issue #575, opened 2024-08-30
Send specific amount of Savings to an address 


rolznz:
Idea from @im-adithya is to alter the existing withdraw flow to add a checkbox the user can choose to drain all, otherwise specify a particular amount.

PR #597, opened 2024-09-03
feat: withdraw custom amount from savings

rolznz:
Closes https://github.com/getAlby/hub/issues/575

This allows users to send explicit amounts, e.g. to do a boltz swap, but is kept in the advanced options as Alby Hub is primarily a lightning wallet.

In the withdraw page, an input and send all checkbox are added. Because of how transaction fee calculation in bitcoin works, send all vs sending an amount are two different actions. The user is also warned if they send all when they have channels open, as it will drain their anchor reserves.

image

image


frontend/src/screens/wallet/WithdrawOnchainFunds.tsx
```diff
@@ -139,18 +152,19 @@ export default function WithdrawOnchainFunds() {
...
-            <AlertDescription>
-              You have channels open and this withdrawal will use some or all of
-              your anchor reserves to publish the transaction, which may make it
-              harder to close channels without depositing additional onchain
-              funds to your savings balance.
-            </AlertDescription>
-          </Alert>
-        )}
+        {!!balances?.onchain.reserved &&
+          (sendAll || +amount > balances.onchain.total * 0.9) && (
```
rolznz, line 156: use spendable instead of total

bumi:
btw. I find this screen a bit too complicated and I have never seen a wallet where I need to confirm my address and check a box.

why does it actually need to be that I am the owner of that address and having the seed phrase of it?

frontend/src/screens/wallet/WithdrawOnchainFunds.tsx
```diff
@@ -159,6 +173,34 @@ export default function WithdrawOnchainFunds() {
...
+                />
+                <Label htmlFor="send-all" className="text-xs">
+                  Send All
+                </Label>
+              </div>
+            </div>
+            <Input
+              id="amount"
+              type="number"
+              value={sendAll ? balances.onchain.total : amount}
```
bumi, line 196: are we sure "send all" will be that amount?

rolznz: No, good catch!

rolznz:
> btw. I find this screen a bit too complicated and I have never seen a wallet where I need to confirm my address and check a box.

@bumi this was because one of our users sent their funds to the wrong address :cry: 

I'll update the copy a bit.

rolznz:
Updated UI:
image

image

Anchor reserve variants:

image

image

Replaced checkbox and confirm input with a confirmation dialog:

image

image

bumi:
that's a nice UI.

PR #600, opened 2024-09-03
fix: do not return LNClient while shutting down

### LINKED ISSUE #391: Restore backup on Mac Desktop does not work

im-adithya:
Fixes #391 - This issue is actually because migrate node is not working and is giving the user a 0-byte file, hence resulting in an EOF error while reading

## Description

Sometimes during backup, calls to get the LNClient information are invoked (not exactly sure why, but it happens randomly), in between the LNClient stop process (see screenshot where calls to get node info and channels are requested between LDK shutdown initiation and complete) which causes the app to panic, this PR adds an `isLNClientShuttingDown` var so that `GetLNClient` returns nil without making calls to LDK during shutdown.

Screenshot 2024-09-03 at 10 15 45 PM

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -33,16 +33,17 @@ import (
 type service struct {
 	cfg config.Config
 
-	db                  *gorm.DB
-	lnClient            lnclient.LNClient
-	transactionsService transactions.TransactionsService
-	albyOAuthSvc        alby.AlbyOAuthService
-	eventPublisher      events.EventPublisher
-	ctx                 context.Context
-	wg                  *sync.WaitGroup
-	nip47Service        nip47.Nip47Service
-	appCancelFn         context.CancelFunc
-	keys                keys.Keys
+	db                     *gorm.DB
+	lnClient               lnclient.LNClient
+	transactionsService    transactions.TransactionsService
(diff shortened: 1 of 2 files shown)


service/stop.go
```diff
@@ -17,7 +17,11 @@ func (svc *service) StopApp() {
 }
 
 func (svc *service) stopLNClient() {
-	defer svc.wg.Done()
+	svc.isLNClientShuttingDown = true
```
rolznz, line 20: maybe we can do
```
lnClient := svc.lnClient
svc.lnClient = nil
```
and proceed to shut down `lnClient`

service/service.go
```diff
@@ -234,6 +234,9 @@ func (svc *service) GetEventPublisher() events.EventPublisher {
 }
 
 func (svc *service) GetLNClient() lnclient.LNClient {
+	if svc.isLNClientShuttingDown {
```
bumi, line 237: is this LNClient specific? or can it be a global boolean indicating the shutdown?

rolznz: I do not think we need one. See https://github.com/getAlby/hub/pull/605

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/605

PR #602, opened 2024-09-03
feat: show pending balances from channel closures (LDK)

rolznz:
Fixes https://github.com/getAlby/hub/issues/601

Only LDK is supported for now.

image

The LDK API is not so easy to deal with here so a lot of code to gather all the different balances, but it's isolated in one place.


bumi (commented):
great idea showing it like that.

Can't wrap my head around the calculation now, but 👍

PR #610, opened 2024-09-04
Removes any existing albyhub.service to prevent duplicating

tidley:
Albyhub service had a config error after re-running the Linux installation script [here](https://github.com/getAlby/hub/tree/master/scripts/linux-x86_64#alby-hub-on-a-linux-server) caused by the albyhub.service written twice.

### DIFF
--- a/scripts/linux-x86_64/install.sh
+++ b/scripts/linux-x86_64/install.sh
@@ -59,11 +59,7 @@ then
   exit
 fi
 
-if [ -s /etc/systemd/system/albyhub.service ]; then
-  sudo truncate -s 0 /etc/systemd/system/albyhub.service
-fi
-
-sudo tee -a /etc/systemd/system/albyhub.service > /dev/null << EOF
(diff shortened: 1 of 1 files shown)


scripts/linux-x86_64/install.sh
```diff
@@ -59,6 +59,10 @@ then
   exit
 fi
 
+if [ -s /etc/systemd/system/albyhub.service ]; then
+  sudo truncate -s 0 /etc/systemd/system/albyhub.service
+fi
+
 sudo tee -a /etc/systemd/system/albyhub.service > /dev/null << EOF
```
bumi, line 66: I think we only should remove the `-a` here which will overwrite the file and not append

bumi:
thanks for reporting this and the PR!

scripts/linux-x86_64/install.sh
tidley: Yes, more concise

rolznz (approved):
utACK

PR #613, opened 2024-09-04
fix: remove lud16 from isolated app secrets

bumi:
Isolated apps should not have anything shared from the main account. The isolated connection would also not see any transaction from that lud16.


reneaaron (approved):
utACK

Issue #623, opened 2024-09-06
Small copy improvement in Family and Friends app


MoritzKa:
What about we name the options

- Alby Go Mobile App
- Alby Web Account
- Alby Browser Extension
- ...

(at least the Alby Browser Extension would be more clear)

![image](https://github.com/user-attachments/assets/fa366ac9-9567-4ede-bc23-ffcde04e1542)

rolznz:
The other changes were made by @bumi except "Alby Account" to "Alby Web Account". I don't think this is a good change actually, the Alby Account is not just about the web. @MoritzKa are you ok with this? I'll close the issue for now.

PR #632, opened 2024-09-09
fix: make friends & family alby account instructions more clear

rolznz:
image

also deletes an unused file (bad merge?)


im-adithya:
There's also https://github.com/getAlby/hub/issues/623 which can be added in this PR

Issue #636, opened 2024-09-10
Bug: Advanced Setup -> Import Existing Mnemonic" fails with "no LNBackendType specified"


rolznz:
![image](https://github.com/user-attachments/assets/e88c36f3-a862-43cf-94da-0bcde8bdb22d)

PR #637, opened 2024-09-11
fix: import mnemonic from advanced setup does not set ldk backend type

rolznz:
Fixes https://github.com/getAlby/hub/issues/636

### DIFF
--- a/frontend/src/screens/setup/SetupAdvanced.tsx
+++ b/frontend/src/screens/setup/SetupAdvanced.tsx
@@ -1,10 +1,19 @@
-import { Link } from "react-router-dom";
+import { Link, useNavigate } from "react-router-dom";
 
 import Container from "src/components/Container";
 import TwoColumnLayoutHeader from "src/components/TwoColumnLayoutHeader";
 import { Button } from "src/components/ui/button";
+import useSetupStore from "src/state/SetupStore";
 
 export function SetupAdvanced() {
+  const navigate = useNavigate();
+  function importLDKMnemonic() {
+    useSetupStore.getState().updateNodeInfo({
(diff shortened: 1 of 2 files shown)


im-adithya (commented):
Can't we just add this in [here](https://github.com/getAlby/hub/blob/master/frontend/src/screens/setup/SetupPassword.tsx#L47) instead of importing things again?
```
store.updateNodeInfo({
  backendType: "LDK",
});
```

rolznz:
> Can't we just add this in [here](https://github.com/getAlby/hub/blob/master/frontend/src/screens/setup/SetupPassword.tsx#L47) instead of importing things again?
> 
> ```
> store.updateNodeInfo({
>   backendType: "LDK",
> });
> ```

@im-adithya I think you're right, that code is also only executed for LDK. Thanks!

alekqss:
I had this problem with ‘Import Existing Mnemonic’..
After a couple of tries I saw the "Connect you Alby account (Request authorization code)" screen
request-auth-code

After doing so I could access to my new Alby Hub, but savings balance is 0 (and it shouldn't) so I checked the recovery phrase and it is different than my original one...

It's a bit weird because the email address and lightning address of the new Alby Hub are the same than the old one ...

Should I simply send the funds in the Savings balance from my old Alby Hub to the new one? 

Or, in case this is fixed in v.1.8.0, should I repeat the ‘Import Existing Mnemonic’ process again?

rolznz:
@alekqss can you try again in 1.8.0? you will have to delete the alby hub working directory so that you can go through the onboarding again. To know where the working directory is, it should be logged to the console on startup.

Issue #484, opened 2024-08-14
Make Alby Account Optional


rolznz:
A lot of Alby features will be lost (lightning address, notifications, SCB, auto-channel on onboarding, etc), however privacy-concious users and Start9 would like this.

rolznz:
@bumi I'm not sure about doing this step by step as we talked about - I think it makes sense for a user to choose this configuration at startup otherwise events will still be sent for a short period, which the user may not want, and they're still required to login.

As you said, this can also be controlled by an environment variable, which can be disabled by default on Start9.

bumi:
> choose this configuration at startup otherwise events will still be sent for a short period, which the user may not want, and they're still required to login.

I don't think this is a problem or a good reason not to offer configuration options in the settings.

PR #639, opened 2024-09-11
Make Alby Account Optional

rolznz:
Fixes https://github.com/getAlby/hub/issues/484

This makes Alby Hub usable without an account by replacing the forced auth with an optional connect flow, and a way to disconnect from the settings. Alby Account specific cards and onboarding steps are removed from the UI.

TODOs:
- [x] suggestions API is not usable without an Alby Account
- [x] first channel flow redirects to channels/incoming - added illustration, but no "first channel" copy - I think this is ok for now?
- [x] new connect account page needs a design iteration - moved to new issue: https://github.com/getAlby/hub/issues/642  
- [x] review copy around connecting/disconnecting account
- [x] fix version check - does not show out of date, just the current version, and there is no link to update.
- [x] satoshi nakamoto name is weird but is a subtle way to motivate users to connect their account for customization
- [x] automatic re-auth if you already had an Alby Account with an expired refresh token
- [x] update to use new Alby info endpoint https://getalby.com/api/internal/info


alby/alby_oauth_service.go
```diff
@@ -839,15 +839,9 @@ func (svc *albyOAuthService) activateAlbyAccountNWCNode(ctx context.Context) err
...
-	if err != nil {
-		logger.Logger.WithError(err).Error("Failed to fetch user token")
-		return nil, err
-	}
-
-	client := svc.oauthConf.Client(ctx, token)
+	client := &http.Client{Timeout: 10 * time.Second}
 
-	req, err := http.NewRequest("GET", fmt.Sprintf("%s/internal/channel_suggestions", svc.cfg.GetEnv().AlbyAPIURL), nil)
+	req, err := http.NewRequest("GET", "https://getalby.com/api/internal/channel_suggestions", nil)
```
reneaaron, line 844: Why remove the host being loaded from the env variable here? :thinking:

reneaaron (commented):
Nice work! :100: 

Most of the things work like a charm, however when I try to re-connect my account I seem to be stuck in a redirect loop: (might be related to my dev setup?)

alby/alby_oauth_service.go
rolznz: I've moved these to constants in this file instead of being in the config, and switched this back to using a constant,

Issue #634, opened 2024-09-10
Bug: LND backend is unable to buy 0-conf channels


rolznz:
LND does not support 0-conf by default. In this case we should provide 1 confirmation instead of 0 for `required_channel_confirmations`. This needs to be applied for both the auto channel endpoint (currently we let the Alby API decide - we can no longer do this) and the LSP (manual flow) endpoint.

PR #641, opened 2024-09-11
fix: use 1 channel confirmation for non-LDK node types

rolznz:
Fixes https://github.com/getAlby/hub/issues/634

I unfortunately cannot test this currently on Mutinynet

### DIFF
--- a/alby/alby_oauth_service.go
+++ b/alby/alby_oauth_service.go
@@ -940,10 +940,14 @@ func (svc *albyOAuthService) requestAutoChannel(ctx context.Context, url string,
 		NodeType        string `json:"node_type"`
 	}
 
+	backendType, err := svc.cfg.Get("LNBackendType", "")
+	if err != nil {
+		return nil, errors.New("failed to get LN backend type")
+	}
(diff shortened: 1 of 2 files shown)


alby/alby_oauth_service.go
```diff
@@ -937,11 +937,13 @@ func (svc *albyOAuthService) requestAutoChannel(ctx context.Context, url string,
 	type autoChannelRequest struct {
 		NodePubkey      string `json:"node_pubkey"`
 		AnnounceChannel bool   `json:"announce_channel"`
+		NodeType        string `json:"node_type"`
 	}
 
 	newAutoChannelRequest := autoChannelRequest{
 		NodePubkey:      pubkey,
 		AnnounceChannel: isPublic,
+		NodeType:        svc.cfg.GetEnv().LNBackendType,
```
bumi, line 946: shouldn't we use those getters on the cfg for this? 

wondering because of that issue where the DB should be prioritized aver the env vars - which means we should always use the data from the DB I guess?

rolznz: yes, sorry!

bumi (approved):

PR #657, opened 2024-09-16
chore: precedence for user config from the DB

bumi:
With this change we always take the configuration from the user configuration in the DB and not overwrite it with the env variables.

So in case a user already has a config or imports an existing DB the environment variables will be ignored.

This has the downside that that once started env variables can not be updated/changed. But this also means that a DB import can be made on setups where env variables are pre configured.


rolznz (approved):
utACK

PR #658, opened 2024-09-17
chore: order apps by last used

### LINKED ISSUE #643: Order apps by last used on connections page
Currently they are ordered chronologically but I think it makes more sense to see the ones you actually use at the top.

im-adithya:
Fixes #643 

## Screenshot

Screenshot 2024-09-17 at 1 21 57 PM

### DIFF
--- a/frontend/src/screens/apps/AppList.tsx
+++ b/frontend/src/screens/apps/AppList.tsx
@@ -20,9 +20,13 @@ function AppList() {
   }
 
   const albyConnection = apps.find((x) => x.name === albyConnectionName);
-  const otherApps = apps.filter(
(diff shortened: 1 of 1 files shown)


reneaaron (approved):
utACK

bumi:
btw. one thing to keep in mind here is that it means the order of the list will always change. As a user if I want to get to a specific detail page of an app I will not be able to remember the position to find it.

jpl-btc:
> As a user if I want to get to a specific detail page of an app I will not be able to remember the position to find it.

Maybe there could be a sort option whether one wants to keep the order by creation date, and thus have the feeling of a static location of the apps, or by last use date, to quickly find an app that one ones to remove or edit that was used recently

Issue #650, opened 2024-09-14
Unable to pay Primal invoice (with Cashu backend)


thebullishbitcoiner:
I'm getting the following error when trying to send sats to a Primal address (e.g. thebullishbitcoiner@primal.net). 

![image](https://github.com/user-attachments/assets/4912634e-2ad7-4f1d-b409-b2d90fd7ab96)

rolznz:
@thebullishbitcoiner which mint did you use?

thebullishbitcoiner:
I used the Minibits mint.

thebullishbitcoiner:
It wasn't working on Alby Go just now but I restarted the app and was able to send to the Primal address and also pay an invoice generated from that address too. Not sure if the success was due to restarting the app or updating to v1.4.2 (3).

I'll try again from the hub and report back.

rolznz:
@thebullishbitcoiner thanks! I've made a fix here: https://github.com/getAlby/hub/pull/660

PR #660, opened 2024-09-18
fix: lnurl pay to use existing lightning address

rolznz:
fixes https://github.com/getAlby/hub/issues/650

@im-adithya how it was implemented before was very unstable. We should avoid editing internals and instead use the public API of the object.

### DIFF
--- a/frontend/src/screens/wallet/send/LnurlPay.tsx
+++ b/frontend/src/screens/wallet/send/LnurlPay.tsx
@@ -60,14 +60,12 @@ export default function LnurlPay() {
   return (
     <form onSubmit={onSubmit} className="grid gap-4">
       <div>
-        <p className="font-medium text-lg mb-2">
-          {lnAddress.lnurlpData?.identifier}
-        </p>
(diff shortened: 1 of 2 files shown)


frontend/src/screens/wallet/send/LnurlPay.tsx
```diff
@@ -14,19 +14,18 @@ export default function LnurlPay() {
...
     try {
+      if (!lnAddress) {
+        throw new Error("no lightning address set");
+      }
       setLoading(true);
-      const lnAddress = new LightningAddress(lnurlDetails.identifier);
-      // this is set instead of calling fetch because
-      // requestInvoice uses lnurlpData.callback
-      lnAddress.lnurlpData = lnurlDetails;
       const invoice = await lnAddress.requestInvoice({
```
im-adithya, line 29: I tried doing this first but for some weird reason this function was not being passed to this screen via state, but somehow it works for me now 👍

im-adithya (approved):
tACK

PR #662, opened 2024-09-18
chore: use proper title in wails

### LINKED ISSUE #466: Add version number to release artifacts
- If you download multiple versions you end up getting (1), (2) on the files (unless you delete the old ones).
- It's hard to know what version you last downloaded

im-adithya:
Changes the title from `albyhub` in wails app to `Alby Hub`

Also fixes #466

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -10,7 +10,7 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
-      - name: Download Linux server archives
+      - name: Download server archives
         uses: actions/download-artifact@v4
         with:
           pattern: albyhub-Server-*
--- a/.github/workflows/wails.yml
+++ b/.github/workflows/wails.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Setup
         shell: bash
         run: |
-          echo "EXEC_NAME=albyhub" >> $GITHUB_ENV
+          echo "EXEC_NAME=Alby Hub" >> $GITHUB_ENV
           echo "PACKAGE_NAME=albyhub-Desktop-${{ runner.os }}" >> $GITHUB_ENV
 
       - uses: actions/checkout@v4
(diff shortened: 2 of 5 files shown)


bumi:
is this good already?

rolznz:
Related: https://github.com/getAlby/hub/pull/695

im-adithya:
⚠️ Wails apps need testing in all 3 platforms

.github/workflows/http.yml
```diff
@@ -53,11 +53,12 @@ jobs:
...
       - name: Setup
         run: |
+          SANITIZED_TAG=$(echo "${GITHUB_REF_NAME}" | sed 's/[./]/-/g')
           echo "EXEC_NAME=albyhub" >> $GITHUB_ENV
           if [ "${{ runner.os }}" == "macOS" ]; then
-            echo "PACKAGE_NAME=albyhub-Server-MacOS" >> $GITHUB_ENV
+            echo "PACKAGE_NAME=AlbyHub-Server-MacOS-${SANITIZED_TAG}" >> $GITHUB_ENV
           else
-            echo "PACKAGE_NAME=albyhub-Server-Linux-${{ matrix.build.arch }}" >> $GITHUB_ENV
+            echo "PACKAGE_NAME=AlbyHub-Server-Linux-${{ matrix.build.arch }}-${SANITIZED_TAG}" >> $GITHUB_ENV
```
rolznz, line 61: will this work with our server update script?

rolznz:
Tested the Linux Desktop build, it seems to work well. I'm not sure we should do the same for the server builds though, will this create issues with our scripts? CC @bumi

.github/workflows/http.yml
bumi: I actually think file names should always be lowercased. If we change it then let's make sure everything is lowercase, then we less likely have issues on case-sensitive systems.

bumi:
if we change the package name, then let's make sure we're using also the lowercased version in the scripts (need to be searched here in the repo and on getalbycom)

.github/workflows/http.yml
im-adithya: Cool, changing it to be lowercase 👍

rolznz: I'm not sure this is the only issue. I think the Alby download links are to exact filenames, but now we add a version suffix on the filename. I think a change in getalby.com is also needed to ensure it picks up the file.

I originally suggested this because downloading two versions with the same filename ends up suffixing the second one with a (1).

bumi: > I originally suggested this because downloading two versions with the same filename ends up suffixing the second one with a (1).

that's the reason why we add the version number to the filename? why would a user do that and even if the user does that why would it be a a problem?

bumi:
ok, I missed the addition of the version to the filename. to me this just makes everything more complicated and then every script any everything needs to be aware of the changing filename.

rolznz (approved):
utACK

PR #665, opened 2024-09-19
chore: set token for olympus LSP

rolznz:


kaloudis:
LGTM

api/lsp.go
```diff
@@ -230,14 +230,19 @@ func (api *api) requestLSPS1Invoice(ctx context.Context, request *LSPOrderReques
 		requiredChannelConfirmations = 6
 	}
 
+	token := ""
+	if request.LSPUrl == "https://lsps1.lnolymp.us/api/v1" {
```
bumi, line 234: do we break things if we do that for all?

bumi (approved):
looks good. :shipit:

Issue #651, opened 2024-09-14
Payments to Cashu backend do not appear in transaction history


thebullishbitcoiner:
Was talking with Jean-Paul about this via email but I think it could use some more eyes.

The hub does not recognize that I’ve received my first sats even though my balance does increase. 

Maybe related: when I paid the invoice, the GUI just kept checking for payment indefinitely.

![image](https://github.com/user-attachments/assets/bf0a650a-6d5e-4d95-9c44-58c08d1a5a83)

thebullishbitcoiner:
Update: I was able to get things going in terms of transaction history by sending a payment to another Alby account (still unable to send to Primal address though). 

It seems like no payment I receive show up (tried from a couple different wallets). 

![image](https://github.com/user-attachments/assets/296ec1ed-a194-41ea-b304-66c67a2942d0)

rolznz:
Thanks for reporting this. I made an issue here: https://github.com/elnosh/gonuts/issues/59

PR #666, opened 2024-09-19
fix: only log a warning if there is no preimage when settling an incoming transaction

rolznz:
Fixes https://github.com/getAlby/hub/issues/651

Alternative fix as the cashu connector is a bit broken right now (incoming payments do not show because of a check to ensure there is a preimage set) and there is nothing in the cashu spec to reveal preimages for receiving payments.

I am not really sure this is good or not - ideally all the settled transactions in our DB have preimages.

### DIFF
--- a/transactions/transactions_service.go
+++ b/transactions/transactions_service.go
@@ -897,7 +897,13 @@ func (svc *transactionsService) markTransactionSettled(tx *gorm.DB, dbTransactio
 	}
 
 	if preimage == "" {
-		return nil, errors.New("no preimage in payment")
+		// outgoing payments MUST have a preimage
+		if dbTransaction.Type == constants.TRANSACTION_TYPE_OUTGOING {
+			return nil, errors.New("no preimage in payment")
+		}
+		// incoming payments SHOULD have a preimage
+		// (currently cashu backend does not)
+		logger.Logger.WithField("payment_hash", dbTransaction.PaymentHash).Warn("no preimage in settled incoming transaction")
 	}
 
 	now := time.Now()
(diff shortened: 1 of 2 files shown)


bumi:
I am always a bit hesitant with such exceptions they will always haunt one the one way or the other. At least they make the code harder to understand and cause questions in the future. why would a settled payment not require a preimage. or some other calculation is done assuming that all successful payments have preimages and then will cause issues. 

Maintaining multiple backends is tricky and with such exceptions even harder. Why does cashu not provide a preimage there?

rolznz:
@bumi I agree. I'd prefer this to be fixed on the Cashu side.

transactions/transactions_service.go
```diff
@@ -897,7 +897,13 @@ func (svc *transactionsService) markTransactionSettled(tx *gorm.DB, dbTransactio
 	}
 
 	if preimage == "" {
-		return nil, errors.New("no preimage in payment")
+		// outgoing payments MUST have a preimage
+		if dbTransaction.Type == constants.TRANSACTION_TYPE_OUTGOING {
+			return nil, errors.New("no preimage in payment")
+		}
+		// incoming payments SHOULD have a preimage
+		// (currently cashu backend does not)
```
bumi, line 905: where does this apply? only cashu? why can transactions not have a preimage. 

is cashu worth this?

thebullishbitcoiner: I also wanted to add that not only are the transactions not showing up in history, I believe the following issue is due to the same cause (not having a preimage). 

When an invoice is created, in the Hub or Go, the GUI waits for payment indefinitely.

image

thebullishbitcoiner:
Not sure if I should've added this comment about the "waiting for payment" issue here.

Here it is just in case.

https://github.com/getAlby/hub/pull/666#discussion_r1768918535

rolznz:
I've made an issue in the cashu nuts repository: https://github.com/cashubtc/nuts/issues/168

PR #671, opened 2024-09-19
fix: use channel balance api in get balances (LND)

### LINKED ISSUE #644: Hub on Umbrel | Spending Balance discrepancy between Alby Hub and LND app
There’s a discrepancy between the spending balance shown by Alby Hub on Umbrel and the LND node. While the LND node shows a fixed balance, the Alby Hub balance fluctuates by a few thousand sats daily. 

My theory, based on previous conversations about this on slack, is that this is likely due to dynamic channel reserve fees, which vary with on-chain fees. The LND node doesn’t display these reserve fees, causing the balances to differ.

im-adithya:
Fixes #644

We recently saw an issue where people saw balance differently in Umbrel, Hub and the Extension

For example:

`1 748 499` - Umbrel
`1 747 569` - Extension
`  987 439` - Hub Node page

This PR makes the Hub Node page return the same balance as the extension. The difference of 930 between Umbrel and Extension is not solved by this PR as it is because of an unsettled HTLC (again, I suspect that it is an invoice of value 500 sats but this additional 430 sats is cut from the balance and I don't know why that is the case)

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -833,12 +833,6 @@ func (svc *LNDService) GetBalances(ctx context.Context) (*lnclient.BalancesRespo
 		return nil, err
 	}
 
-	balances, err := svc.client.ChannelBalance(ctx, &lnrpc.ChannelBalanceRequest{})
-	if err != nil {
-		logger.Logger.WithError(err).Error("Failed to retrieve lightning balance")
-		return nil, err
-	}
-
 	var totalReceivable int64 = 0
 	var totalSpendable int64 = 0
 	var nextMaxReceivable int64 = 0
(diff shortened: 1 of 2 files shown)


lnclient/lnd/lnd.go
```diff
@@ -858,11 +869,11 @@ func (svc *LNDService) GetBalances(ctx context.Context) (*lnclient.BalancesRespo
 	return &lnclient.BalancesResponse{
 		Onchain: *onchainBalance,
 		Lightning: lnclient.LightningBalanceResponse{
-			TotalSpendable:    totalSpendable,
-			TotalReceivable:   totalReceivable,
+			TotalSpendable:    int64(balances.LocalBalance.Msat),
```
rolznz, line 872: this does not work - it does not remove the channel reserve.

rolznz:
@im-adithya I fixed the balances so they discount the channel reserve, and added the missing calculations (these are useful from settings -> debug tools to check why you cannot make a payment)

PR #672, opened 2024-09-19
chore: remove get balance method from LN client

### LINKED ISSUE #669: LNClient should not have two balance methods
Consider removing the `GetBalance()` method so we always use `GetBalances()`?

Double check LND reserves are not shown in the spendable balance

im-adithya:
Fixes #669

This removes the repeated `GetBalance` method which is redundant due to `GetBalances` and uses the latter in NIP47 Get Balance Handler


rolznz (approved):
tACK

PR #674, opened 2024-09-19
feat: alby go app

reneaaron:
image

Would actually be quite easy to create a dynamic page with some fallback (for the "how to connect" section) that does the same thing for all available apps. Not sure if we should take this other step right away, wdyt?

### DIFF
--- a/frontend/src/screens/Home.tsx
+++ b/frontend/src/screens/Home.tsx
@@ -93,7 +93,7 @@ function Home() {
                     </div>
                   </CardTitle>
                   <CardDescription className="ml-4">
-                    A simple mobile wallet that works great with Alby Hub.
+                    The easiest Bitcoin mobile app that works great with Alby Hub.
                   </CardDescription>
                 </div>
               </div>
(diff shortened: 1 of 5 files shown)


frontend/src/screens/Home.tsx
```diff
@@ -77,6 +78,34 @@ function Home() {
             </Card>
           </ExternalLink>
         )}
+
+        <ExternalLink to="https://www.albygo.com">
```
rolznz, line 82: Should this go to `/internal-apps/alby-go` ?

frontend/src/screens/internal-apps/AlbyGo.tsx
```diff
@@ -0,0 +1,112 @@
...
+                <div className="text-sm font-normal text-muted-foreground">
+                  {app?.description}
+                </div>
+              </div>
+            </div>
+          </>
+        }
+        description=""
+        contentRight={
+          <Link to={`/apps/new?app=${app?.id}`}>
```
rolznz, line 41: Ideally we just show a qr code on the screen when you press this button rather than linking, so the user doesn't have to see all the wallet permissions options. But maybe we can do this as a second iteration

rolznz (approved):
tACK

PR #678, opened 2024-09-19
feat: transaction metadata (zaps, payerdata, comment)

### LINKED ISSUE #577: Parse invoice metadata to get LNURL payerdata and zap
```
{
  comment: "Hello", // lud-12
  payerData: {...}, // lud-18
  nostr: {...} // nip-57 zap request event
}
```

im-adithya:
Fixes #577

Just adds the backend, needs design input to handle zaps and payerdata on the frontend

### DIFF
--- a/api/models.go
+++ b/api/models.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+	"encoding/json"
 	"io"
 	"time"
 
@@ -230,6 +231,24 @@ type Metadata struct {
 	PayerData *PayerData  `json:"payerData,omitempty"`
 }
 
+func (m *Metadata) UnmarshalJSON(data []byte) error {
+	var t struct {
+		Comment   string      `json:"comment"`
+		Nostr     *NostrEvent `json:"nostr"`
+		PayerData *PayerData  `json:"payer_data"`
+	}
+
+	if err := json.Unmarshal(data, &t); err != nil {
+		return err
+	}
+
+	m.Comment = t.Comment
+	m.Nostr = t.Nostr
+	m.PayerData = t.PayerData
+
+	return nil
+}
+
 type NostrEvent struct {
 	Content   string     `json:"content"`
 	CreatedAt int64      `json:"createdAt"`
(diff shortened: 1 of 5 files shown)


api/models.go
```diff
@@ -218,11 +220,31 @@ type Transaction struct {
 	CreatedAt       string      `json:"createdAt"`
 	SettledAt       *string     `json:"settledAt"`
 	AppId           *uint       `json:"appId"`
-	Metadata        Metadata    `json:"metadata,omitempty"`
+	Metadata        *Metadata   `json:"metadata,omitempty"`
 	Boostagram      *Boostagram `json:"boostagram,omitempty"`
 }
 
-type Metadata = map[string]interface{}
+type Metadata struct {
```
rolznz, line 227: Transaction metadata can be anything. It _could_ contain these fields. With this change it probably breaks other types of metadata

api/models.go
```diff
@@ -218,11 +220,31 @@ type Transaction struct {
...
+	Content   string     `json:"content"`
+	CreatedAt int64      `json:"createdAt"`
+	ID        string     `json:"id"`
+	Kind      int        `json:"kind"`
+	PubKey    string     `json:"pubkey"`
+	Sig       string     `json:"sig"`
+	Tags      [][]string `json:"tags"`
+}
+
+type PayerData struct {
```
rolznz, line 243: PayerData can also contain anything. We just know about these fields

bumi:
what's the status on this one?

api/models.go
im-adithya: No it will not break other types, we just ignore them while converting to the api metadata (which we use in transaction list). If you pass in other fields and do a `list_transactions` call, you'll see the other fields too from the database.

api/models.go
im-adithya: Same as above^

api/models.go
im-adithya: But yeah having this here causes confusion as Metadata (and Boostagram) also exists in the db.Transaction type as `datatypes.JSON` but we again specify it differently here but not add it to `transactions.Transaction` struct (we only use these Metadata and Boostagram types in api > transactions.go)

im-adithya: Done!

im-adithya:
We just need the frontend, backend should be good now!

rolznz:
Closing as implemented in #803

Issue #380, opened 2024-07-31
NWC spec changes list


rolznz:
We've made some changes and are tracking them here to one day make PR(s):
- add `metadata` to nwc make_invoice https://github.com/getAlby/hub/pull/357
- NIP-47 notifications (open PR)

rolznz:
Merged!
- [x] NIP-47 notifications ([open PR](https://github.com/nostr-protocol/nips/pull/1164))
  - Add `fees_paid` to all payment responses

PR #682, opened 2024-09-20
feat: show pending and failed outgoing transactions

rolznz:
Fixes https://github.com/getAlby/hub/issues/426

Failed and unsettled outgoing transactions are now shown in the transaction list. Pending transactions have a pulse animation.

I also added an alert on the send page if the user has any pending payments, to avoid the chance they accidentally try pay again.

Note: Listing transactions is now ordered by `updated_at` rather than `settled_at, created_at`

image

image

image

image

image

image

image

TODOs:
- [x] fix ordering (settled at is an issue for unpaid payments)
- [ ] should transactions that have expired (unsettled after expiration of the invoice) be updated to failed?
- [ ] support params for NIP-47 so Alby Go can do the same

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,7 @@ FRONTEND_URL=http://localhost:5173
 #RELAY=ws://localhost:7447/v1
 #PORT=8080
 
+#LOG_DB_QUERIES=true
 
 # Alby OAuth configuration
 #ALBY_OAUTH_CLIENT_SECRET=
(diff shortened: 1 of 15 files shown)


rolznz:
I've also updated https://github.com/getAlby/hub/issues/380 which I think we need two things:
- splitting unpaid into outgoing and incoming
- including the transaction state in the transaction object (settled, pending, failed)

Issue #677, opened 2024-09-19
Prevent buying a public channel with Zeus LSP


rolznz:
Zeus LSP has no plan to support public channels.

When the user tries to check the checkbox, they could get a warning saying this LSP does not support public channels?

PR #686, opened 2024-09-21
feat: disable public channel checkbox for channel partners that do not support public channels

rolznz:
Fixes https://github.com/getAlby/hub/issues/677

The checkbox is disabled, with a message when you hover over it. I think this might be sufficient for now.


bumi (commented):
utACK

Issue #683, opened 2024-09-20
Use short sync interval while channels are opening


rolznz:
This could possibly fix the issue we saw where channels were unusable after being opened.

PR #688, opened 2024-09-21
fix: use short sync interval while LDK channel is waiting to become active

rolznz:
Fixes https://github.com/getAlby/hub/issues/683

### DIFF
--- a/lnclient/ldk/ldk.go
+++ b/lnclient/ldk/ldk.go
@@ -266,6 +266,15 @@ func NewLDKService(ctx context.Context, cfg config.Config, eventPublisher events
 			case <-time.After(MIN_SYNC_INTERVAL):
 				ls.syncing = true
 
+				channels := ls.node.ListChannels()
+				for _, channel := range channels {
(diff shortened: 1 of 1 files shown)


lnclient/ldk/ldk.go
```diff
@@ -266,6 +266,15 @@ func NewLDKService(ctx context.Context, cfg config.Config, eventPublisher events
 			case <-time.After(MIN_SYNC_INTERVAL):
 				ls.syncing = true
 
+				channels := ls.node.ListChannels()
+				for _, channel := range channels {
+					if channel.Confirmations != nil && channel.ConfirmationsRequired != nil && *channel.Confirmations < *channel.ConfirmationsRequired {
+						logger.Logger.WithField("channel_id", channel.UserChannelId).Debug("Using short sync time while opening channel")
+						ls.lastWalletSyncRequest = time.Now()
+						break
```
bumi, line 274: why break?

rolznz: To exit the for loop if we already found a pending channel as there is no need to iterate the rest of the channels. It's not really necessary I guess. Do you think it should be removed?

Issue #681, opened 2024-09-20
Change port for self-hosted Linux scripts


rolznz:
Currently we use 8080 which often conflicts with other services (even LND) which then the user has to check the service logs and manually edit the service and restart it before they can fix the issue.

Currently the service does not set a port at all (see https://github.com/getAlby/hub/blob/master/scripts/linux-x86_64/install.sh vs https://github.com/getAlby/hub/blob/master/scripts/pi-arm/install.sh)

Use 8029 in the install script

PR #690, opened 2024-09-21
fix: use a less-used port for the alby hub service so it doesn't conflict with other running apps

rolznz:
Fixes https://github.com/getAlby/hub/issues/681

This only applies to the Linux x86_64 service and introduces inconsistencies between ports.

### DIFF
--- a/scripts/linux-x86_64/README.md
+++ b/scripts/linux-x86_64/README.md
@@ -22,7 +22,7 @@ Optionally it can also create a systemd service for you.
 
 You can also do these quite simple steps manually, have a look in the install script for details.
 
-Alby Hub will run on localhost:8080 (standalone) or localhost:8029 (as a service) configurable using the `PORT` environment variable or by editing `Environment="PORT=8029"` in the albyhub.service file - See "Editing The Service" below)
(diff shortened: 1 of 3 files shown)


scripts/linux-x86_64/README.md
```diff
@@ -23,7 +22,7 @@ Optionally it can also create a systemd service for you.
 
 You can also do these quite simple steps manually, have a look in the install script for details.
 
-Alby Hub will run on localhost:8080 by default (configurable using the `PORT` environment variable or by adding `Environment="PORT=8081"` to the `[Service]` section of the albyhub.service file - See "Editing The Service" below)
+Alby Hub will run on localhost:8080 (standalone) or localhost:8029 (as a service) configurable using the `PORT` environment variable or by editing `Environment="PORT=8029"` in the albyhub.service file - See "Editing The Service" below)
```
bumi, line 25:
```suggestion
Alby Hub will run on localhost:8080 (standalone) or localhost:8029 (when run with a systemd service) configurable using the `PORT` environment variable or by editing `Environment="PORT=8029"` in the albyhub.service systemd config file - See "Editing The Service" below)
```

PR #693, opened 2024-09-23
feat: include LDK channel monitors in static channel backups

rolznz:
TODO:
- [x] rename "Key Backup" -> "Backup" + tell users to keep the seed safe and also save the data directory (on the cloud this is taken care for you)
- [x] update LDK-node dependency and undo changes to use local LDK-node version
- [ ] add new issue to derive an encryption key from seed instead of using unlock password to recover SCBs
- [ ] add new issue to improve backup pages - one for alby account vs no alby account
- [ ] test decrypting locally the encrypted saved data to getalby.com
- [ ] test LDK-node flow of starting with restored monitors (old channel state)
- [ ] make guide pages on how to do backups

### DIFF
--- a/events/models.go
+++ b/events/models.go
@@ -20,18 +20,18 @@ type Event struct {
 	Properties interface{} `json:"properties,omitempty"`
 }
 
-type ChannelBackupEvent struct {
-	NodeID   string                 `json:"node_id"`
-	Channels []ChannelBackupInfo    `json:"channels"`
-	Monitors []ChannelMonitorBackup `json:"monitors"`
+type StaticChannelsBackupEvent struct {
+	NodeID   string                        `json:"node_id"`
+	Channels []ChannelBackup               `json:"channels"`
+	Monitors []EncodedChannelMonitorBackup `json:"monitors"`
 }
 
-type ChannelMonitorBackup struct {
+type EncodedChannelMonitorBackup struct {
 	Key   string `json:"key"`
 	Value string `json:"value"`
 }
 
-type ChannelBackupInfo struct {
+type ChannelBackup struct {
 	ChannelID         string `json:"channel_id"`
 	PeerID            string `json:"peer_id"`
 	PeerSocketAddress string `json:"peer_socket_address"`
--- a/events/models.go
+++ b/events/models.go
@@ -21,6 +21,7 @@ type Event struct {
 }
 
 type ChannelBackupEvent struct {
+	NodeID   string                 `json:"node_id"`
 	Channels []ChannelBackupInfo    `json:"channels"`
 	Monitors []ChannelMonitorBackup `json:"monitors"`
 }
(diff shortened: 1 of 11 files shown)


events/models.go
```diff
@@ -21,7 +21,13 @@ type Event struct {
 }
 
 type ChannelBackupEvent struct {
-	Channels []ChannelBackupInfo `json:"channels"`
+	Channels []ChannelBackupInfo    `json:"channels"`
+	Monitors []ChannelMonitorBackup `json:"monitors"`
+}
+
+type ChannelMonitorBackup struct {
```
rolznz, line 28:
```suggestion
type EncodedChannelMonitorBackup struct {
```

frontend/src/screens/BackupMnemonic.tsx
```diff
@@ -124,7 +125,21 @@ export function BackupMnemonic() {
...
+                    You also need to make sure you do not forget your unlock
+                    password as this will be used to recover funds from
+                    channels.
+                  </b>
+                )}
+                {!info?.albyAccountConnected && (
+                  <b>
+                    Make sure to also backup your data directory as this is
+                    required to recover funds on your channels. You can also
+                    connect your Alby Account for automatic backups.
```
bumi, line 140: automatic encrypted backups
```suggestion
                    connect your Alby Account for automatic encrypted backups (you still need your seed and unlock password to decrypt those)
```

bumi: we should make a guides page with all the backup details what is needed in what case.

lnclient/ldk/ldk.go
```diff
@@ -224,6 +224,10 @@ func NewLDKService(ctx context.Context, cfg config.Config, eventPublisher events
 		"duration": math.Ceil(time.Since(syncStartTime).Seconds()),
 	}).Info("LDK node synced successfully")
 
+	// backup channels after successful startup
+	// TODO: consider removing after all hubs are updated past 1.9.1
+	ls.backupChannels()
```
bumi, line 229: don't you think it's good to keep this on bootup anyway? we write new backups anyway.

events/models.go
```diff
@@ -21,7 +21,13 @@ type Event struct {
 }
 
 type ChannelBackupEvent struct {
-	Channels []ChannelBackupInfo `json:"channels"`
+	Channels []ChannelBackupInfo    `json:"channels"`
+	Monitors []ChannelMonitorBackup `json:"monitors"`
```
bumi, line 25: can we add the node id here separately?

lnclient/ldk/ldk.go
rolznz: maybe it's good, yeah

lnclient/ldk/ldk.go
```diff
@@ -119,6 +119,24 @@ func NewLDKService(ctx context.Context, cfg config.Config, eventPublisher events
...
+			if err != nil {
+				logger.Logger.WithError(err).Error("Failed to decode LDK channel monitor hex")
+				continue
+			}
+			encodedMonitors = append(encodedMonitors, ldk_node.KeyValue{
+				Key:   monitor.Key,
+				Value: value,
+			})
+		}
+		builder.RestoreEncodedChannelMonitors(encodedMonitors)
```
rolznz, line 137: add a function to decode the monitors

PR #704, opened 2024-09-25
chore: make silhouttte svgs opaque

im-adithya:
Should be easier to adjust opacity now using css

<table>
<tr>
<td>
<h3>Light Mode</h3>
</td>
<td>
<h3>Dark Mode</h3>
</td>
</tr>

<tr>
<td>
Screenshot 2024-09-26 at 1 45 14 PM
</td>

<td>
Screenshot 2024-09-26 at 1 45 30 PM
</td>

</tr>
</table>


frontend/public/images/quotes/antonopoulos.svg
```diff
@@ -1 +1,4 @@
...
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" width="667" height="800" fill="none">
+  <path fill="gray"
```
rolznz, line 2: I think we should also use `fill=currentColor"` so we can specify it. Gray seems to not be so clear in light vs dark mode.

im-adithya: We should make it into an icon if we want thi currentColor to work, because we are currently passing this svg as an src to img tag , should I make this white instead of gray? It would look like this then:

Screenshot 2024-09-25 at 11 54 16 AM

Compared to what we have right now:
Screenshot 2024-09-25 at 11 56 59 AM

rolznz: @im-adithya this won't work on dark mode, right?

Actually I was wrong, if `gray` is completely half way between black and white it should be ok. We might just need to make the opacity stronger in light mode vs dark mode, since it seems less visible there.

> HTML / CSS defined gray at the midpoint of the 8-bit gray scale (128,128,128).

im-adithya: Ah yes it wouldn't, but actually the current one as is looks good both on Light Mode and Dark Mode

<table>
<tr>
<td>
<h3>Light Mode</h3>
</td>
<td>
<h3>Dark Mode</h3>
</td>
</tr>

<tr>
<td>
Screenshot 2024-09-25 at 12 58 06 PM
</td>

<td>
Screenshot 2024-09-25 at 12 58 43 PM
</td>

</tr>
</table>

rolznz: @im-adithya the light and dark mode are back to front, but the one with the black background is very hard to see the silhoutte. Can we use a different opacity on that one? e.g. `opacity-40 dark:opacity-20`

im-adithya: Done! Looks better now!


<table>
<tr>
<td>
<h3>Light Mode</h3>
</td>
<td>
<h3>Dark Mode</h3>
</td>
</tr>

<tr>
<td>
Screenshot 2024-09-26 at 1 45 14 PM
</td>

<td>
Screenshot 2024-09-26 at 1 45 30 PM
</td>

</tr>
</table>

rolznz (approved):
utACK

PR #705, opened 2024-09-25
chore: hide preimage in failed payment dialog

im-adithya:
<table>
<tr>
<td>
<h3>Before</h3>
</td>
<td>
<h3>After</h3>
</td>
</tr>
<td>
Screenshot 2024-09-25 at 10 22 53 AM
</td>
<td>
Screenshot 2024-09-25 at 10 23 58 AM
</td>
</tr>
</table>


im-adithya:
Should we also hide the fee?

rolznz:
> Should we also hide the fee?

I think so, a fee cannot be collected if the payment fails so it will always be zero

rolznz (approved):
utACK

PR #706, opened 2024-09-25
fix: reload balances, txs and channels after transfer is complete

### LINKED ISSUE #692: Bug: migrate shared funds on wallet page does not refresh the wallet balance and transaction list

im-adithya:
Fixes #692

Also reloading channels so the liquidity changes in Node screen

### DIFF
--- a/frontend/src/screens/channels/Channels.tsx
+++ b/frontend/src/screens/channels/Channels.tsx
@@ -270,11 +270,13 @@ export default function Channels() {
                 variant="outline"
                 channels={channels}
                 albyBalance={albyBalance}
-                onTransferComplete={async () => {
-                  await reloadAlbyBalance();
-                  await reloadBalances();
-                  await reloadChannels();
-                }}
+                onTransferComplete={() =>
+                  Promise.all([
+                    reloadAlbyBalance(),
+                    reloadBalances(),
+                    reloadChannels(),
+                  ])
+                }
               >
                 Migrate
               </TransferFundsButton>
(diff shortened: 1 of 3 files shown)


frontend/src/screens/channels/Channels.tsx
```diff
@@ -67,9 +67,9 @@ import { request } from "src/utils/request";
 
 export default function Channels() {
   useSyncWallet();
-  const { data: channels } = useChannels();
+  const { data: channels, mutate: reloadChannels } = useChannels();
```
rolznz, line 70: why do we need to reload the channels?

frontend/src/screens/wallet/index.tsx
```diff
@@ -61,7 +63,11 @@ function Wallet() {
               <TransferFundsButton
                 channels={channels}
                 albyBalance={albyBalance}
-                reloadAlbyBalance={reloadAlbyBalance}
+                onTransferComplete={async () => {
+                  await reloadAlbyBalance();
```
rolznz, line 67: You could use `Promise.all` here, and then there is no need to make the function async if it returns a single promise

frontend/src/screens/channels/Channels.tsx
im-adithya: The liquidity also changes right?

frontend/src/screens/wallet/index.tsx
im-adithya: Ah nice, thanks 🙌 I was also thinking how to change as this looks weird

frontend/src/screens/channels/Channels.tsx
rolznz: true, thanks!

rolznz (approved):
utACK

bumi (commented):
utACK

PR #708, opened 2024-09-25
chore: use slashed zero and sensitive

im-adithya:
Adds slashed-zeros everywhere we use amounts and sensitive where we reveal balances (onchain and lightning) and invoice amounts (i.e. only in transactions list)

### DIFF
--- a/frontend/src/components/BalanceCard.tsx
+++ b/frontend/src/components/BalanceCard.tsx
@@ -33,7 +33,7 @@ function BalanceCard({
         <BalanceCardIcon className="h-4 w-4 text-muted-foreground" />
       </CardHeader>
       <CardContent>
-        <div className="text-2xl font-bold balance sensitive">
+        <div className="text-2xl font-bold balance sensitive slashed-zero">
           {new Intl.NumberFormat().format(Math.floor(balance / 1000))} sats
         </div>
       </CardContent>
--- a/frontend/src/screens/apps/ShowApp.tsx
+++ b/frontend/src/screens/apps/ShowApp.tsx
@@ -222,7 +222,7 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
             <CardHeader>
               <CardTitle>Info</CardTitle>
             </CardHeader>
-            <CardContent>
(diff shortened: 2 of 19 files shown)


frontend/src/components/BalanceCard.tsx
```diff
@@ -33,7 +33,7 @@ function BalanceCard({
         <BalanceCardIcon className="h-4 w-4 text-muted-foreground" />
       </CardHeader>
       <CardContent>
-        <div className="text-2xl font-bold balance sensitive">
+        <div className="text-2xl font-bold balance sensitive slashed-zero">
```
rolznz, line 36: is `sensitive` required? we added posthog only temporarily for initial testing. It's already been removed.

im-adithya: Ah, so then should I remove sensitive from other places too?

rolznz: @bumi do you think we still need this classname?

bumi: does that hurt? 
tailwind likes classes. and actually that one is the most descriptive one of them :) 

https://imgflip.com/i/94m5ma

frontend/src/screens/apps/ShowApp.tsx
```diff
@@ -222,7 +222,7 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
             <CardHeader>
               <CardTitle>Info</CardTitle>
             </CardHeader>
-            <CardContent>
+            <CardContent className="slashed-zero">
```
rolznz, line 225: can you be more precise where you add it? this adds the classname to all child elements

im-adithya: I wanted the Dates and Pubkey etc to also have a slashed zero, so I applied it to the parent like we did in the DialogTrigger in TransactionItem
https://github.com/getAlby/hub/blob/71183a131a03919f98ee9db271353ab52f1021ba/frontend/src/components/TransactionItem.tsx#L164

rolznz (approved):
utACK

Issue #479, opened 2024-08-13
Manage isolated connection 


matjaz:
- add balance from main wallet
- withdraw balance to main wallet
- changing expiration

rolznz:
@matjaz I feel like the adding balance is the most important thing for a first iteration. Changing expiration is here: https://github.com/getAlby/hub/issues/598

I'm not sure about whether we should expose withdrawing balance, although this is a trusted setup anyway

PR #710, opened 2024-09-25
feat: increase isolated app balance

rolznz:
Closes https://github.com/getAlby/hub/issues/479

Adds ability to "Increase" the balance of an isolated app in 3 places (new isolated app connection, isolated app connection page, friends & family app)

Also renames the db service as this only had app-related functionality, and cleaned up some previous TODOs there.

On the UI it's just called "Increase". The user is warned that they must actually have enough real balance on their channels to make payments, as the isolated app balance could be larger than their channel balance.

TODO:
- [ ] Test Wails mode
- [ ] Create a new issue for followup improvement - to allow setting choosing a balance before creating apps (UI change only)

image

image

image

image

image

image

### DIFF
--- a/alby/alby_oauth_service.go
+++ b/alby/alby_oauth_service.go
@@ -20,6 +20,7 @@ import (
 	"golang.org/x/oauth2"
 	"gorm.io/gorm"
 
+	"github.com/getAlby/hub/apps"
 	"github.com/getAlby/hub/config"
 	"github.com/getAlby/hub/constants"
 	"github.com/getAlby/hub/db"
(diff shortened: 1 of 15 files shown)


bumi:
Is it possible to set this also similar to the budget also when creating the connection?

> print paper money using it

I don't think your comparison his correct here. there is no "paper money'. It allows a connection to spend a certain amount. and maybe the admin allows a connection to spend the same money that the admin already promised to another connection.

But we're not building a bank or a financial system here, so I think you should not try to compare it to that. 
It's an app to manage sats and that allows the owner of the sats to give some people the permission to spend those sats. 
comparing a wife-husband relationship to a broken financial system or to some bank is imo flawed. We need to think what the user wants and what makes sense for that setup.

rolznz:
@bumi 
> Is it possible to set this also similar to the budget also when creating the connection?

I think we can. Internally it'd still need to create and pay an invoice after the app is created, but the UX flow might be better that way, true.

bumi:
> I think we can. Internally it'd still need to create and pay an invoice after the app is created, but the UX flow might be better that way, true.

we just have to make sure that our API endpoints stay simple and unversal and don't get bloated with such things.

PR #713, opened 2024-09-25
fix: use string or number type for boostagram field flexibility

im-adithya:
## Description
Currently if someone sends a boostagram with wrong datatype for Episode, FeedId, ItemId or SenderId (i.e. number instead of string) we throw an error, this fix makes it flexible to accept both string/number in boostagram json


transactions/transactions_service.go
```diff
@@ -49,20 +49,44 @@ const (
...
+	Timestamp      int64          `json:"ts,omitempty"`
+	Message        string         `json:"message,omitempty"`
+	SenderId       StringOrNumber `json:"sender_id"`
+	SenderName     string         `json:"sender_name"`
+	Time           string         `json:"time"`
+	Action         string         `json:"action"`
+	ValueMsatTotal int64          `json:"value_msat_total"`
+}
+
+type StringOrNumber struct {
```
rolznz, line 68: Did you look at any other ways to do it, like Did you look at any other ways to do it? like https://stackoverflow.com/a/77624153/4562693 ? it seems `json.Number` might be a nice type so we don't need to implement it ourselves

im-adithya: That's nice! But these fields are unpredictable right? What if it's a uuid?

rolznz: Ah, good point! it won't work for a uuid I guess.

rolznz (approved):
utACK

PR #715, opened 2024-09-26
feat: generate lightning address for isolated app (WIP)

rolznz:
- [ ] communication in the UI
  - [ ] this is a paid service
  - [ ] generating a lightning address should be optional?
- [ ] update URL once the service is deployed

### DIFF
--- a/alby/alby_oauth_service.go
+++ b/alby/alby_oauth_service.go
@@ -20,6 +20,7 @@ import (
 	"golang.org/x/oauth2"
 	"gorm.io/gorm"
 
+	"github.com/getAlby/hub/apps"
 	"github.com/getAlby/hub/config"
 	"github.com/getAlby/hub/constants"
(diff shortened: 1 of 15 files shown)


bumi:
I somehow don't think that this a good architecture. that `CreateApp` has too much knowledge and does too many things. Why does `CreateApp` need to know when and how to create a lightning address? 

I wonder how we can separate this and maybe we should make this explicit and the user needs to actively create the LN address - also the user needs to agree to the costs of it actually.

rolznz:
Ideas: separate app for creating a lightning address? maybe this can be a good first iteration.

Enabling a feature after creating the friends & family app could be a second iteration?

rolznz:
Closing - decided instead to focus on improving the experience with getalby.com accounts - invites and possibly ability to manage multiple lightning addresses.

PR #722, opened 2024-10-03
fix: version comparision logic

im-adithya:
The current Hub won't show 1.10.1 as the new update in the Sidebar because we do a `"1.9.0" >= "1.10.1"` comparsion which results in `true` meaning the current version is wrongly shown as up to date. This fixes the comparision logic by adding the `compareVersions` utils function.

Screenshot 2024-10-03 at 4 12 02 PM

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -44,6 +44,7 @@
     "canvas-confetti": "^1.9.2",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.0",
+    "compare-versions": "^6.1.1",
     "date-fns": "^3.6.0",
     "dayjs": "^1.11.10",
     "embla-carousel-react": "^8.0.2",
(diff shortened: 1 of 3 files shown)


bumi:
comparing versions sucks, but to see if it's newer is something like this not enough?

```
parseInt("1.10.1".replaceAll(".","")) > parseInt("1.9.10".replaceAll("."))
```

reneaaron:
@bumi That doesn't work (1.10 would be smaller than 1.9.10 with your suggestion). 

Why not just use a package (that is heavily tested, etc) for that, e.g. https://www.npmjs.com/package/compare-versions

bumi:
I thought our versions are rather predictable with constant length. but we actually already have a minor with two digits. 

compare-versions sounds good as it does not has any dependencies. I did not read the exact solution here. what do you like better @im-adithya

im-adithya:
Going with René's suggestion to use compare-versions package 👍

Issue #703, opened 2024-09-25
Implement get_budget NWC method


rolznz:
See https://github.com/nostr-protocol/nips/pull/1504

jklein24:
Implemented by https://github.com/getAlby/hub/pull/726

PR #726, opened 2024-10-07
Add get_budget implementation

jklein24:
Closes https://github.com/getAlby/hub/issues/703


frontend/src/types.ts
```diff
@@ -56,6 +58,7 @@ export type ScopeIconMap = {
 
 export const scopeIconMap: ScopeIconMap = {
   get_balance: WalletMinimal,
+  get_budget: WalletMinimal,
```
jklein24, line 61: Let me know if you've got a good icon for this.

nip47/permissions/permissions.go
```diff
@@ -109,6 +109,8 @@ func scopeToRequestMethods(scope string) []string {
 		return []string{models.PAY_INVOICE_METHOD, models.PAY_KEYSEND_METHOD, models.MULTI_PAY_INVOICE_METHOD, models.MULTI_PAY_KEYSEND_METHOD}
 	case constants.GET_BALANCE_SCOPE:
 		return []string{models.GET_BALANCE_METHOD}
+	case constants.GET_BUDGET_SCOPE:
```
jklein24, line 112: Happy to instead include this in one of the other scopes if you think it makes sense.

constants/constants.go
```diff
@@ -22,6 +22,7 @@ const (
 const (
 	PAY_INVOICE_SCOPE       = "pay_invoice" // also covers pay_keysend and multi_* payment methods
 	GET_BALANCE_SCOPE       = "get_balance"
+	GET_BUDGET_SCOPE        = "get_budget"
```
rolznz, line 25: I am wondering if we should not have a scope for this one - instead it chould always be callable. Is there any benefit of having a scope? (also, an extra scope requires including an extra checkbox in the UI that the user has to understand when configuring permissions)

nip47/controllers/get_budget_controller.go
```diff
@@ -0,0 +1,51 @@
...
+	logger.Logger.WithFields(logrus.Fields{
+		"request_event_id": requestEventId,
+	}).Debug("Getting budget")
+
+	appPermission := db.AppPermission{}
+	controller.db.Where("app_id = ? AND scope = ?", app.ID, models.PAY_INVOICE_METHOD).First(&appPermission)
+
+	maxAmount := appPermission.MaxAmountSat
+	if maxAmount == 0 {
+		publishResponse(&models.Response{
```
rolznz, line 32: We actually did not catch this in your [proposal](https://github.com/nostr-protocol/nips/pull/1504). I am not sure an empty result makes sense. Maybe the connection should not support this method and return an error instead if the app tries to call this method when a budget is not set?

rolznz:
@jklein24 thanks for the PR. Awesome!

constants/constants.go
jklein24: Agreed having it always be callable makes sense. Is there another method which has that behavior right now that I could copy? If not, I'll just dig a bit deeper to make that work.

nip47/controllers/get_budget_controller.go
jklein24: Hmm that's a good alternative too. Maybe we should discuss on the NIP PR. There are some tradeoffs to think through for sure.

constants/constants.go
jklein24: Added an always granted scope with some tests for this. That way, if there are other "always allowed" methods, they can just be added there.

rolznz: I think the `get_info` method also bypasses the check here: https://github.com/getAlby/hub/blob/master/nip47/event_handler.go#L272

We need to come up with a better solution, but for now to see if it works you could try adding an `&& nip47Request.Method != models.GET_BUDGET_METHOD`

jklein24: Oh cool, should I just add get_info to my new ALWAYS_GRANTED scope instead?

rolznz: I'm not sure about the `ALWAYS_GRANTED` scope. This seems to have some side effects (for example, the new scope is now returned and the frontend does not know how to display it) 
image

Also the scope is now included in only new apps in the database (there is a new app_permissions row for always_granted), but only new ones. The older app connections will not have it.

image

And it seems a bit hacky to even call HasPermission explicitly with the always_granted scope which actually just always returns true anyway.

I know my original suggestion is not the best, but on the plus side it's isolated to a single line of code and simply skips the permission check for that request method.

@bumi @im-adithya any thoughts on this one?

jklein24: Ah, thanks for catching the UI issue, I hadn't noticed that! FWIW, my intent here was actually to not store this scope in the DB or show it in the UI at all. Instead, I was hoping to just have it be implicitly granted to all requests. I'll try to make that the case now, but can definitely switch to a specific GET_BUDGET_SCOPE if you prefer. I just think it's a bit weird to show that in the UI in general.

jklein24: Adjusted to avoid always_granted in the db and UI. Let me know if you'd still prefer an explicit budget scope though.

rolznz: Thanks! I also don't think we should have a `GET_BUDGET_SCOPE`. I was just wondering if a single line check like I suggested would be simpler than having this `ALWAYS_GRANTED` scope, and in the end we just bypass the logic anyway.

rolznz:
@jklein24 I have added some changes in a feedback PR here https://github.com/jklein24/alby-hub/pull/1

I've tested the flow, and everything looks good to me. Please let me know if you have any concerns with the changes!

rolznz:
JS SDK implementation: https://github.com/getAlby/js-sdk/pull/264

jklein24:
> @jklein24 I have added some changes in a feedback PR here [jklein24#1](https://github.com/jklein24/alby-hub/pull/1)
> 
> I've tested the flow, and everything looks good to me. Please let me know if you have any concerns with the changes!

LGTM, thanks!

rolznz (approved):
tACK

im-adithya (approved):
tACK

PR #732, opened 2024-10-09
fix: add get_info to isolated app permissions

im-adithya:
Adds `get_info` permission to isolated apps, previously created isolated app connections will remain unchanged


reneaaron (approved):
utACK

PR #733, opened 2024-10-09
chore: change alby hub icon in wails and pwa

im-adithya:
Changes icons and also adds an attempt to load the PWA icon in mobile


frontend/vite.config.ts
```diff
@@ -18,15 +18,11 @@ export default defineConfig(({ command }) => ({
         "icon-512.png",
         "icon-192.png",
       ],
+      useCredentials: true, // because the icons sit behind authentication
```
reneaaron, line 21:
```suggestion
      useCredentials: true, // because the manifest might sit behind authentication
```

PR #736, opened 2024-10-14
feat: app child key derived from wallet master key

### LINKED ISSUE #724: Separate wallet key for each connection
This will allow better privacy between connections, which currently can see metadata from events of other connections since it uses the same wallet key.

### LINKED ISSUE #744: rename App.NostrPubkey to App.AppPubKey

frnandu:
fixes #724 
fixes #744


db/migrations/202410141503_wallet_child_idx.go
```diff
@@ -0,0 +1,26 @@
+package migrations
```
rolznz, line 1: is this still needed? or can we use the app ID as the derivation path?

db/models.go
```diff
@@ -16,14 +16,15 @@ type UserConfig struct {
...
-	Metadata    datatypes.JSON
+	ID             uint
+	Name           string `validate:"required"`
+	Description    string
+	NostrPubkey    string `validate:"required"`
+	walletChildIdx uint   `validate:"required"`
+	CreatedAt      time.Time
+	UpdatedAt      time.Time
+	Isolated       bool
+	Metadata       datatypes.JSON
```
rolznz, line 27: How do we ensure the old app connections work using the original wallet service key? do we need an app version field?

nip47/event_handler.go
```diff
@@ -75,7 +76,7 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
 				Message: fmt.Sprintf("Failed to save nostr event: %s", err.Error()),
 			},
 		}
-		resp, err := svc.CreateResponse(event, nip47Response, nostr.Tags{}, ss)
+		resp, err := svc.CreateResponse(event, nip47Response, nostr.Tags{}, ss, svc.keys.GetNostrSecretKey())
```
rolznz, line 79: there is no need to pass this as an argument, is there? since it's already available in the `svc` and we're calling a method on the `svc`

rolznz: Ah, I see it depends if the app has its own key or not (it's an old app)

nip47/event_handler.go
```diff
@@ -159,8 +162,17 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
...
 	}).Debug("App found for nostr event")
 
+	appWalletPrivKeyBip32, err := svc.keys.GetBIP32ChildKey(uint32(app.ID))
+	if err != nil {
+		logger.Logger.WithFields(logrus.Fields{
+			"appId": app.ID,
+		}).WithError(err).Error("error deriving child key")
+		return
+	}
+	appWalletPrivKey, _ := nostr.GetPublicKey(hex.EncodeToString(appWalletPrivKeyBip32.Serialize()))
```
rolznz, line 172: is this right? the variable name is private but we call getPublicKey

nip47/event_handler.go
```diff
@@ -369,14 +381,16 @@ func (svc *nip47Service) CreateResponse(initialEvent *nostr.Event, content inter
 	allTags := nostr.Tags{[]string{"p", initialEvent.PubKey}, []string{"e", initialEvent.ID}}
 	allTags = append(allTags, tags...)
 
+	appWalletPubKey, _ := nostr.GetPublicKey(walletPrivKey)
```
rolznz, line 384: maybe `walletPrivKey` could be renamed to `appWalletPrivKey` to be consistent with the public one

nip47/notifications/nip47_notifier.go
```diff
@@ -108,7 +109,17 @@ func (notifier *Nip47Notifier) notifySubscriber(ctx context.Context, app *db.App
 		"appId":        app.ID,
 	}).Debug("Notifying subscriber")
 
-	ss, err := nip04.ComputeSharedSecret(app.NostrPubkey, notifier.keys.GetNostrSecretKey())
+	appWalletPrivKeyBIP32, err := notifier.keys.GetBIP32ChildKey(uint32(app.ID))
```
rolznz, line 112: this code is duplicated / slightly different in a few places - maybe we can move it to the `keys` module?

rolznz: Edit: it already is, but maybe we should add a method there to get the hex value directly

service/start.go
```diff
@@ -119,6 +119,82 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
 	return nil
 }
 
+func (svc *service) SubscribeToAppRequests(ctx context.Context, appWalletPubKey string) error {
```
rolznz, line 122: This seems duplicated from the existing code. Can we update the existing method to use this one for the original wallet service key?

service/start.go
```diff
@@ -119,6 +119,82 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+	relayUrl := svc.cfg.GetRelayUrl()
```
rolznz, line 123: Actually, all the relay connection code should not be per app, only setting up subscriptions after relay connection should be per app

frnandu: yep, been refactoring that all morning :)

db/models.go
frnandu: We should be able to compare the p tag on the request events to keys.GetNostrPublicKey, if it matches it means it is an old app connection still not using child keys, right?

nip47/notifications/nip47_notifier.go
frnandu: I think we need to actually check for the app much before to see if it is a child key or not, and if so use the child private key to do the nip04.ComputeSharedSecret for all needed places, which should resolve the duplication.

service/start.go
frnandu: yep, in progress major refactor :)

db/models.go
frnandu: Hmm, but when starting up the hub we actually don't know if a certain app from DB is an old version or new with child key, and thus can't decide if we need to open a separate subscription for that app or not.
So it might be actually needed some kind of flag/version on the app model.

frnandu: Maybe we add a field with wallet child pubKey which we only fill for new app connections created with child derived keys?
We would avoid then having to calculate each child derived key from master priv key (and from that the corresponding nostr pubKey) during the starting phase of the hub when we only need those apps pubKeys for subscribing to nwc requests on the relay.
Wdyt?

rolznz: @frnandu that sounds good to me. We then only need to derive the child private keys when creating events, right?

frnandu: I think also when handling events for `nip04.ComputeSharedSecret(event.PubKey, appWalletPrivKey)`

db/migrations/202410141503_wallet_child_idx.go
frnandu: refactored to use wallet_child_pubkey

db/models.go
frnandu: done

db/db_service.go
```diff
@@ -28,7 +28,17 @@ func NewDBService(db *gorm.DB, eventPublisher events.EventPublisher) *dbService
...
+func (svc *dbService) CreateApp(
+	name string,
+	pubkey string,
+	maxAmountSat uint64,
+	budgetRenewal string,
+	expiresAt *time.Time,
+	scopes []string,
+	isolated bool,
+	metadata map[string]interface{},
+	walletChildPrivKeyGeneratorFunc func(appId uint32) (string, error),
```
rolznz, line 40: Minor: rather than passing this method as an argument, the keys could be passed to the `db_service` in the `NewDBService` function

db/db_service.go
```diff
@@ -82,6 +92,21 @@ func (svc *dbService) CreateApp(name string, pubkey string, maxAmountSat uint64,
 			}
 		}
 
+		appWalletChildPrivKey, err := walletChildPrivKeyGeneratorFunc(uint32(app.ID))
```
rolznz, line 95: The pubkey can be passed in the original App declaration at line 72. I don't think there's any need to update the app after it's created inside the transaction

db/models.go
```diff
@@ -16,14 +16,15 @@ type UserConfig struct {
...
-	NostrPubkey string `validate:"required"`
-	CreatedAt   time.Time
-	UpdatedAt   time.Time
-	Isolated    bool
-	Metadata    datatypes.JSON
+	ID                uint
+	Name              string `validate:"required"`
+	Description       string
+	NostrPubkey       string `validate:"required"`
+	WalletChildPubkey string
```
rolznz, line 23: I'm wondering about the naming. Is `Child` needed or is it just an implementation detail?

I think actually it would be nice to rename the original `NostrPubkey` too, so we clearly differentiate the two. e.g. `AppPubkey` vs `WalletPubkey`

db/db_service.go
```diff
@@ -94,9 +119,28 @@ func (svc *dbService) CreateApp(name string, pubkey string, maxAmountSat uint64,
...
+	err := svc.db.Delete(app).Error
+	if err != nil {
+		return err
+	}
+	svc.eventPublisher.Publish(&events.Event{
+		Event: "app_deleted",
+		Properties: map[string]interface{}{
+			"name":              app.Name,
+			"id":                app.ID,
+			"walletChildPubkey": app.WalletChildPubkey,
```
rolznz, line 142: I'm not sure the wallet pubkey should be passed here. Instead, we can fetch it using the app ID if we need it (same with `app_created`)

service/keys/keys.go
```diff
@@ -12,6 +16,8 @@ type Keys interface {
 	GetNostrPublicKey() string
 	// Wallet Service Nostr secret key
 	GetNostrSecretKey() string
+	// GetBIP32ChildKey derives a BIP32 child key from the nostrSecretKey given a child key index
+	GetBIP32ChildKey(childIndex uint32) (string, error)
```
rolznz, line 20: Regarding naming, do we need to know it's BIP32? could it be `GetAppWalletKey(appId uint32)`?

service/keys/keys.go
```diff
@@ -51,3 +57,29 @@ func (keys *keys) GetNostrPublicKey() string {
 func (keys *keys) GetNostrSecretKey() string {
 	return keys.nostrSecretKey
 }
+
+func (keys *keys) GetBIP32ChildKey(childIndex uint32) (string, error) {
```
rolznz, line 61: Just to confirm, we are ok with the performance impact of this (twice for request/response) and once per notification

service/start.go
```diff
@@ -119,6 +115,132 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+			if err != nil {
+				logger.Logger.WithField("event", event).Error("Failed to publish nip47 info deletion")
+			}
+		}
+	}
+}
+
+func (svc *service) startAllExistingAppsWalletSubscriptions(ctx context.Context, relay *nostr.Relay) {
+	var apps []db.App
+	result := svc.db.Where("wallet_child_pubkey != ?", "").Find(&apps)
```
rolznz, line 191: is this query correct, shouldn't it be `wallet_child_pubkey IS NOT NIL`?

service/start.go
```diff
@@ -119,6 +115,132 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+		logger.Logger.WithError(result.Error).Error("Failed to fetch App records with non-empty WalletChildPubkey")
+		return
+	}
+
+	for _, app := range apps {
+		go func(app db.App) {
+			if app.WalletChildPubkey != "" {
+				err := svc.startAppWalletSubscription(ctx, relay, app.WalletChildPubkey, "")
+				if err != nil {
+					logger.Logger.WithError(err).Error("Failed to subscribe to wallet")
```
rolznz, line 202: Can we use `.WithFields` and include the app ID or something to identify what wallet the error was for?

service/start.go
```diff
@@ -119,6 +115,132 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
 }
 
+type createAppSubscriber struct {
+	events.EventSubscriber
+	svc   *service
+	relay *nostr.Relay
+}
+
+func (s *createAppSubscriber) ConsumeEvent(ctx context.Context, event *events.Event, globalProperties map[string]interface{}) {
+	switch event.Event {
```
rolznz, line 125: it seems kind of strange to have a switch since we only check one event kind.

Could we do `if event.Event != "app_created" return` instead?

service/start.go
```diff
@@ -119,6 +115,132 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+				if err != nil {
+					logger.Logger.WithError(err).Error("Failed to subscribe to wallet")
+				}
+				// TODO what should we do?
+			}()
+		}
+	}
+}
+
+type deleteAppSubscriber struct {
```
rolznz, line 158: I am wondering if we even need to do this. The next time Alby Hub is started or even next time Alby Hub connects to the relay, this should automatically be cleaned up?

service/start.go
```diff
@@ -119,6 +115,132 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+			go func() {
+				appWalletPrivKey, err := s.svc.keys.GetBIP32ChildKey(uint32(id))
+				if err != nil {
+					logger.Logger.WithError(err).Error("Failed to calculate app wallet priv key")
+				}
+				err = s.svc.startAppWalletSubscription(ctx, s.relay, walletChildPubkey, appWalletPrivKey)
+				if err != nil {
+					logger.Logger.WithError(err).Error("Failed to subscribe to wallet")
+				}
+				// TODO what should we do?
```
rolznz, line 152: I think this is fine to just log an error - if the relay is not connectable, the next time the relay connects the relay subscription will be setup again

nip47/event_handler.go
```diff
@@ -45,6 +45,14 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
 		}).Error("invalid event signature")
 		return
 	}
+	pTag := event.Tags.GetFirst([]string{"p"})
+	if pTag == nil {
```
rolznz, line 49: Why was this added? we can get it from the app, right?

the spec says `SHOULD` but here it is implemented as `MUST`

service/keys/keys.go
```diff
@@ -12,6 +16,8 @@ type Keys interface {
 	GetNostrPublicKey() string
 	// Wallet Service Nostr secret key
 	GetNostrSecretKey() string
+	// GetBIP32ChildKey derives a BIP32 child key from the nostrSecretKey given a child key index
```
rolznz, line 19: Should we add TODOs to above to remove the old method after a year so we can remove conditionals (and delete all apps that don't have a wallet pubkey)?

db/db_service.go
frnandu: I tried that but get:
```
package command-line-arguments
	imports github.com/getAlby/hub/http
	imports github.com/getAlby/hub/alby
	imports github.com/getAlby/hub/config
	imports github.com/getAlby/hub/db
	imports github.com/getAlby/hub/service/keys
	imports github.com/getAlby/hub/config: import cycle not allowed
```
Is is possible to easily fix this import cycle without too much refactoring?

db/db_service.go
frnandu: Since the key needs `app.ID`, which AFAIK will only be generated after `tx.Create` I think we need to do the update later, right?

db/models.go
frnandu: Will rename to WalletPubKey.
Renaming NostrPubkey let's maybe do it on a separate PR?

service/start.go
frnandu: I'm actually not sure, because if for some reason an app DB row is created but then it fails to update the column with the calculated pubkey it will have an empty string '' as value, which the current query filters, and `IS NOT NULL` does not.
Is there a way to make sure that when a new DB row is created without passing a value to that column it will have NULL value instead of ''?

service/start.go
```diff
@@ -119,6 +115,132 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+	var apps []db.App
+	result := svc.db.Where("wallet_child_pubkey != ?", "").Find(&apps)
+	if result.Error != nil {
+		logger.Logger.WithError(result.Error).Error("Failed to fetch App records with non-empty WalletChildPubkey")
+		return
+	}
+
+	for _, app := range apps {
+		go func(app db.App) {
+			if app.WalletChildPubkey != "" {
```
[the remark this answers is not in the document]
frnandu: If we would change the query above to `IS NOT NULL` maybe it is needed to make sure we don't start subscription for empty keys....?

service/start.go
frnandu: I think it's more hygienic to unsubscribe that nostr req and delete the 13194 event from the relay.
Without that the relay will keep events and subscriptions for deleted Apps that are no longer used.

nip47/event_handler.go
frnandu: My mistake, thanks!

service/keys/keys.go
frnandu: How do we make sure that people upgrading will not have any legacy apps anymore?

db/db_service.go
rolznz: Ah, true. Thanks!

db/db_service.go
rolznz: This might be fixed in https://github.com/getAlby/hub/pull/710 which I move the app management out of the db package. This will cause some conflicts here when it's merged, which I can then resolve.

service/keys/keys.go
rolznz: Yeah, it seems like it might be best to rethink this and store encrypted private keys in the app table. Then we don't need any conditional logic.

db/models.go
rolznz: Sounds good, thanks. A TODO could be good though so we don't forget.

service/start.go
rolznz: Ah, gorm uses the default value of the column type. If it was a `*string` instead of a `string` it would work. But maybe we will ensure all the apps have keys, so this is no longer an issue.

service/keys/keys.go
frnandu: I'll do some bench-marking just to be sure what is the impact.

service/keys/keys.go
frnandu: One issue I see with saving priv keys on DB for the legacy apps is that since all legacy apps will have to use the same (existing single) priv/pub wallet key (so not to break existing connections with clients), we still need conditional code to handle not deleting the single nip47 info event (kind 13194) from the relay when legacy app gets deleted, until it is the last legacy app in hub, right?

service/start.go
frnandu: Maybe some clients use that 13194 event to check if a NWC connection exists or/and can be used, and that gets broken in unpredictable ways.
Besides it will be inconsistent behavior if a request 23194 with command get_info returns invalid connection but the info event 13194 still returns info for a wallet key that doesn't exist anymore.

service/keys/keys.go
rolznz: I guess so, but only this one simple check. It's a big improvement over the backward compatibility issues we would have with not saving the encrypted private keys in the app table.

db/db_service.go
```diff
@@ -82,6 +92,21 @@ func (svc *dbService) CreateApp(name string, pubkey string, maxAmountSat uint64,
 			}
 		}
 
+		appWalletPrivKey, err := walletChildPrivKeyGeneratorFunc(uint32(app.ID))
```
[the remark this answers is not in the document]
frnandu: How would that cause a  conflict? 
Doesn't  PRIMARY KEY AUTO_INCREMENTED on a sqlite column ensure it will be unique and free of conflicts?

service/keys/keys.go
frnandu: The backward compatibility needed are only 2 simple checks (if WalletPubKey column is empty will use the single master key). Saving the encrypted private keys seems to me unnecessary (only for the sake of eliminating 2 simple checks) and more complicated to maintain.

db/db_service.go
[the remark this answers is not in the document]
bumi: I was thinking about some import again or some reset. but we create a random master key always 🤔

service/start.go
```diff
@@ -119,6 +115,127 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+	id, ok := properties["id"].(uint)
+	if !ok {
+		logger.Logger.WithField("event", event).Error("Failed to get app id")
+		return
+	}
+	walletPrivKey, err := s.svc.keys.GetAppWalletKey(uint32(id))
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to calculate app wallet priv key")
+	}
+	walletPubKey, _ := nostr.GetPublicKey(walletPrivKey)
```
rolznz, line 142: can the error be caught here?  rather than checking the walletPubKey is not "" a few lines below

service/keys/keys.go
frnandu: `GetAppWalletKey average execution time per call (10000 cycles): 2.622242ms`

service/start.go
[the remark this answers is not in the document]
frnandu: removed

service/start.go
```diff
@@ -119,6 +115,127 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+	var apps []db.App
+	result := svc.db.Where("wallet_pubkey != ?", "").Find(&apps)
+	if result.Error != nil {
+		logger.Logger.WithError(result.Error).Error("Failed to fetch App records with non-empty WalletPubkey")
+		return
+	}
+
+	for _, app := range apps {
+		go func(app db.App) {
+			if app.WalletPubkey != "" {
```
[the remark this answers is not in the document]
frnandu: removed

db/db_service.go
rolznz: fixed!

service/start.go
rolznz: @frnandu I think this would be a good change and actually to do the rename now (https://github.com/getAlby/hub/issues/744). I will make this change

nip47/publish_nip47_info.go
```diff
@@ -11,7 +12,7 @@ import (
 	"github.com/nbd-wtf/go-nostr"
 )
 
-func (svc *nip47Service) PublishNip47Info(ctx context.Context, relay nostrmodels.Relay, lnClient lnclient.LNClient) error {
+func (svc *nip47Service) PublishNip47Info(ctx context.Context, relay nostrmodels.Relay, appWalletPubKey string, appWalletPrivKey string, lnClient lnclient.LNClient) (*nostr.Event, error) {
```
rolznz, line 15: @frnandu I think for new apps the capabilities could be based on what permissions are set on the app (before we couldn't do this). I think @im-adithya mentioned this previously.

If we think this would be good to do, maybe this could be moved to a separate issue? we'd need to repost the event when the app permissions are updated. CC @bumi @reneaaron

frnandu: Yes, I can do that.

nip47/event_handler.go
```diff
@@ -102,7 +114,7 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
 				Message: "The public key does not have a wallet connected.",
 			},
 		}
-		resp, err := svc.CreateResponse(event, nip47Response, nostr.Tags{}, ss)
+		resp, err := svc.CreateResponse(event, nip47Response, nostr.Tags{}, ss, svc.keys.GetNostrSecretKey())
```
rolznz, line 117: I think this is wrong. @frnandu I don't think we should reply here actually. What do you think?

nip47/event_handler.go
rolznz: @frnandu  I think this is wrong. Actually I don't think we should reply in this case because we don't even know what key to use now

nip47/event_handler.go
```diff
@@ -134,7 +146,7 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
 				Message: fmt.Sprintf("Failed to save app to nostr event: %s", err.Error()),
 			},
 		}
-		resp, err := svc.CreateResponse(event, nip47Response, nostr.Tags{}, ss)
+		resp, err := svc.CreateResponse(event, nip47Response, nostr.Tags{}, ss, svc.keys.GetNostrSecretKey())
```
rolznz, line 149: This one is definitely wrong. If we can't save events to the DB I think we should log an error and return

nip47/notifications/nip47_notifier.go
```diff
@@ -108,7 +108,12 @@ func (notifier *Nip47Notifier) notifySubscriber(ctx context.Context, app *db.App
 		"appId":        app.ID,
 	}).Debug("Notifying subscriber")
 
-	ss, err := nip04.ComputeSharedSecret(app.NostrPubkey, notifier.keys.GetNostrSecretKey())
+	appWalletPrivKey := notifier.keys.GetNostrSecretKey()
+	if app.WalletPubkey != nil {
+		appWalletPrivKey, _ = notifier.keys.GetAppWalletKey(app.ID)
```
rolznz, line 113: can you add error handling here?

nip47/event_handler.go
```diff
@@ -373,14 +385,16 @@ func (svc *nip47Service) CreateResponse(initialEvent *nostr.Event, content inter
 	allTags := nostr.Tags{[]string{"p", initialEvent.PubKey}, []string{"e", initialEvent.ID}}
 	allTags = append(allTags, tags...)
 
+	appWalletPubKey, _ := nostr.GetPublicKey(appWalletPrivKey)
```
rolznz, line 388: Can we add error handling rather than `, _` - I think there are quite a few places in this PR

service/start.go
```diff
@@ -119,6 +116,52 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+	}
+
+	logger.Logger.Info("Subscribing to events for wallet ", appWalletPubKey)
+	sub, err := relay.Subscribe(ctx, svc.createFilters(appWalletPubKey))
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to subscribe to events")
+	}
+
+	// register a subscriber for "app_deleted" events, which handles nostr subscription cancel and nip47 info event deletion
+	svc.eventPublisher.RegisterSubscriber(&deleteAppConsumer{nostrSubscription: sub, walletPubkey: appWalletPubKey, svc: svc, relay: relay, infoEventId: infoEventId})
```
rolznz, line 155: Doesn't this need to be only done for new apps?

nip47/event_handler.go
frnandu: If we can response, maybe it's better for the client to get an error than get nothing, wdyt?
But yes, the key should be appWalletPrivKey. I'll fix it if you agree we should respond.

rolznz: I don't think there's any point because the client will not listen on a different key. I think just logging an error and returning is enough.

nip47/event_handler.go
rolznz: Let's just log an error and return?

frnandu: You're right, if there is no app with the key, we can't even encrypt the response, so just log the error.

nip47/event_handler.go
frnandu: I think at that point it already found the app so the client is using a correct key, and we can respond using the `appWalletPrivKey` for that wallet right?

service/start.go
```diff
@@ -119,6 +116,52 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+
+	err = svc.StartSubscription(sub.Context, sub)
```
rolznz, line 157: maybe this method should be moved to start.go

service/start.go
frnandu: Yes, but maybe we would want to cancel subscription and delete nip47 event once every legacy app has been deleted? Than we would need to handle that in the `ConsumeEvent` doing some db query count for `wallet_pubkey IS NULL` ?
Also, we can check if there are any legacy apps in the hub at start and do not subscribe nor create the nip47 event for the master key.
Let me know what you think, and I can add code for that.

nip47/event_handler.go
```diff
@@ -47,7 +47,6 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
 		return
 	}
 
-	ss, err := nip04.ComputeSharedSecret(event.PubKey, svc.keys.GetNostrSecretKey())
```
rolznz, line 50: Did you intend to remove this check? now on the right hand side we check the same error twice

rolznz: Maybe the second error check just needs to be removed

service/start.go
rolznz: I was wrong, the `if s.walletPubkey == walletPubKey {` check will always return false for legacy apps. Thanks!

nip47/notifications/nip47_notifier.go
```diff
@@ -108,7 +108,12 @@ func (notifier *Nip47Notifier) notifySubscriber(ctx context.Context, app *db.App
...
+	}
+
+	ss, err := nip04.ComputeSharedSecret(app.AppPubkey, appWalletPrivKey)
```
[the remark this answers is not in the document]
frnandu: `tests/create_app.go` now uses `AppsService` so it should test the new behavior with the new `walletPubKey`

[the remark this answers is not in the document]
rolznz: But we still have notifications for legacy apps too

nip47/event_handler.go
rolznz: oops, yeah I was on the wrong thread. Thanks

nip47/publish_nip47_info.go
rolznz: I created a separate issue: https://github.com/getAlby/hub/issues/766, resolving this for now.

service/create_app_consumer.go
```diff
@@ -0,0 +1,52 @@
...
+		return
+	}
+	id, ok := properties["id"].(uint)
+	if !ok {
+		logger.Logger.WithField("event", event).Error("Failed to get app id")
+		return
+	}
+	walletPrivKey, err := s.svc.keys.GetAppWalletKey(id)
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to calculate app wallet priv key")
```
rolznz, line 36: missing return, please check all!

service/create_app_consumer.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	walletPubKey, err := nostr.GetPublicKey(walletPrivKey)
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to calculate app wallet pub key")
```
rolznz, line 40: missing return

service/delete_app_consumer.go
```diff
@@ -0,0 +1,47 @@
...
+	properties, ok := event.Properties.(map[string]interface{})
+	if !ok {
+		logger.Logger.WithField("event", event).Error("Failed to cast event.Properties to map")
+		return
+	}
+	id, _ := properties["id"].(uint)
+
+	walletPrivKey, err := s.svc.keys.GetAppWalletKey(id)
+	if err != nil {
+		logger.Logger.WithError(err).WithField("id", id).Error("Failed to calculate app wallet priv key")
```
rolznz, line 35: missing return

service/start.go
```diff
@@ -89,22 +91,17 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
-			sub, err := relay.Subscribe(ctx, svc.createFilters(svc.keys.GetNostrPublicKey()))
-			if err != nil {
-				logger.Logger.WithError(err).Error("Failed to subscribe to events")
-				continue
-			}
-			err = svc.StartSubscription(sub.Context, sub)
+			// register a subscriber for events of "app_created" which handles creation of nostr subscription for new app
+			svc.eventPublisher.RegisterSubscriber(&createAppConsumer{svc: svc, relay: relay})
+
+			// legacy single wallet subscription - only subscribe once for all legacy apps
```
rolznz, line 102: change to only set this up if at least one legacy app exists

service/delete_app_consumer.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	walletPubKey, _ := nostr.GetPublicKey(walletPrivKey)
```
rolznz, line 37: add error handling

service/start.go
frnandu: This is consumer for `app_created` events, so one generally always is needed to handle app creation right?

service/start.go
```diff
@@ -126,25 +145,41 @@ func (svc *service) startAllExistingAppsWalletSubscriptions(ctx context.Context,
 
 	for _, app := range apps {
 		go func(app db.App) {
-			err := svc.startAppWalletSubscription(ctx, relay, *app.WalletPubkey, "")
+
+			// get nip47 event info for this app wallet key
+			nip47InfoEvent, err := svc.GetNip47Service().GetNip47Info(ctx, relay, *app.WalletPubkey)
```
rolznz, line 150: what is the benefit of fetching the info first?

service/start.go
```diff
@@ -99,13 +99,32 @@ func (svc *service) startNostr(ctx context.Context, encryptionKey string) error
...
+
+				// legacy single wallet subscription - only subscribe once for all legacy apps
+				// to ensure we do not get duplicate events
+				err = svc.startAppWalletSubscription(ctx, relay, svc.keys.GetNostrPublicKey(), nil)
+				if err != nil {
+					//err being non-nil means that we have an error on the websocket error channel. In this case we just try to reconnect.
+					logger.Logger.WithError(err).Error("Got an error from the relay while listening to subscription.")
+					continue
+				}
+			} else {
```
rolznz, line 125: This seems wrong. This is done inside a loop that will re-connect to the relay on errors. But here we only listen for context.Done. I think this will cause a deadlock if the relay disconnects, but no exit signal has happened, right?

Also having the `non-legacy` logic inside an `else` kind of feels wrong, like a code smell to me.

service/start.go
rolznz: Do we really need to do this? couldn't it cause more likelyhood of errors (and if an error occurs, the subscription is not created)

frnandu: You're right, we can do the fetching only when needed on the `deleteAppConsumer`.
We need to fetch the event since for deleting it we need it's `ID`

service/start.go
frnandu: replaced with `<-relay.Context().Done()` and handle `relay.ConnectionError` to continue in the loop

service/keys/keys.go
```diff
@@ -16,6 +18,8 @@ type Keys interface {
 	GetNostrPublicKey() string
 	// Wallet Service Nostr secret key (DEPRECATED)
 	GetNostrSecretKey() string
+	// Derives a BIP32 child key from appKey derived child dedicated for app wallet keys
+	GetAppWalletKey(childIndex uint) (string, error)
 	// Derives a child BIP-32 key from the app key (derived from the mnemonic)
 	DeriveKey(path []uint32) (*bip32.Key, error)
 }
```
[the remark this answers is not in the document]
frnandu: I don't think these are used anymore, or at least I can't find any usage besides in README.md.
I suggest removing this completely.

service/keys/keys.go
```diff
@@ -82,6 +103,16 @@ func (keys *keys) GetNostrSecretKey() string {
 	return keys.nostrSecretKey
 }
 
+func (keys *keys) GetAppWalletKey(appID uint) (string, error) {
+	path := []uint32{bip32.FirstHardenedChild + 1, uint32(appID)}
```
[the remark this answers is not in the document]
frnandu: `bip32.FirstHardenedChild + 0` is being used for SCB encryption, so `+1` is used for wallet keys.
But you're right that it should use also the first hardened child as the second level of derivation adding appID to that. Fixing, thanks!

nip47/event_handler.go
```diff
@@ -70,52 +61,50 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
...
 	err = svc.db.First(&app, &db.App{
-		NostrPubkey: event.PubKey,
+		AppPubkey: event.PubKey,
 	}).Error
 	if err != nil {
 		logger.Logger.WithFields(logrus.Fields{
-			"nostrPubkey": event.PubKey,
+			"appPubkey": event.PubKey,
 		}).WithError(err).Error("Failed to find app for nostr pubkey")
+		return
```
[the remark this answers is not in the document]
frnandu: If we don't have an app connection for that key anymore we have no way of encrypting the response, do we?
Actually we shouldn't get any requests for that key anyways, since the nostr subscription for that wallet pubKey should no longer exist and we don't get events from the relay for that anymore.

nip47/publish_nip47_info.go
```diff
@@ -11,7 +12,27 @@ import (
...
+
+	if len(events) == 0 {
+		return nil, nil
+	}
+
+	return events[0], nil
+}
+
+func (svc *nip47Service) PublishNip47Info(ctx context.Context, relay nostrmodels.Relay, appWalletPubKey string, appWalletPrivKey string, lnClient lnclient.LNClient) (*nostr.Event, error) {
+	// TODO: should the capabilities be based on the app permissions? (for non-legacy apps)
```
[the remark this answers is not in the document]
frnandu: Yes, will handle that on a separate issue/PR -> https://github.com/getAlby/hub/issues/766

api/api.go
```diff
@@ -80,7 +80,8 @@ func (api *api) CreateApp(createAppRequest *CreateAppRequest) (*CreateAppRespons
 		expiresAt,
```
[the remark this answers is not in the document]
rolznz: @im-adithya could you create an issue and link some code?

service/delete_app_consumer.go
```diff
@@ -0,0 +1,67 @@
...
+		return
+	}
+	walletPubKey, err := nostr.GetPublicKey(walletPrivKey)
+	if err != nil {
+		logger.Logger.WithError(err).WithField("id", id).Error("Failed to calculate app wallet pub key")
+		return
+	}
+	// Note: for legacy apps this check will always return false as the wallet pubkey
+	// generated by the id will not match the master key which is used for all legacy apps
+	if s.walletPubkey == walletPubKey {
```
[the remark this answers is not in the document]
frnandu: Isn't this listener ConsumeEvent called for each Subscriber added, meaning as many as each app connection wallet? And we want to only do something for the specific one matching the `walletPubkey` we saved in `deleteAppConsumer` struct rigth?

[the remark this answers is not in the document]
rolznz: I am not sure it's worth it to add this extra code. The old info event doesn't really do any harm, does it? (there will be others - e.g. if users abandon old hubs there will be many info events for apps that are no longer used and were never deleted)

nip47/nip47_service.go
```diff
@@ -29,8 +28,10 @@ type Nip47Service interface {
 	events.EventSubscriber
 	StartNotifier(ctx context.Context, relay *nostr.Relay, lnClient lnclient.LNClient)
 	HandleEvent(ctx context.Context, relay nostrmodels.Relay, event *nostr.Event, lnClient lnclient.LNClient)
-	PublishNip47Info(ctx context.Context, relay nostrmodels.Relay, lnClient lnclient.LNClient) error
-	CreateResponse(initialEvent *nostr.Event, content interface{}, tags nostr.Tags, ss []byte) (result *nostr.Event, err error)
+	GetNip47Info(ctx context.Context, relay *nostr.Relay, appWalletPubKey string) (*nostr.Event, error)
+	PublishNip47Info(ctx context.Context, relay nostrmodels.Relay, appWalletPubKey string, appWalletPrivKey string, lnClient lnclient.LNClient) (*nostr.Event, error)
+	PublishNip47InfoDeletion(ctx context.Context, relay nostrmodels.Relay, appWalletPubKey string, appWalletPrivKey string, infoEventId string) error
```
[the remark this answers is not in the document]
frnandu: Maybe we could, but since we already need/use `appWalletPubKey` in other places "outside" (before calling the service) and we have it already calculated , it would cause us to unnecessarily do multiple times `nostr.GetPublicKey(walletPrivKey)` which is not ideal IMO. Wdyt?

nip47/event_handler.go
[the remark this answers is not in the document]
im-adithya: Ah true ✅

nip47/nip47_service.go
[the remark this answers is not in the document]
im-adithya: True, just wanted to align things. But if we introduce pubkey for `CreateResponse` it would be additional code for legacy case. Best to leave it this way 👍

service/delete_app_consumer.go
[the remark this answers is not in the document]
im-adithya: > And we want to only do something for the specific one matching the walletPubkey we saved in deleteAppConsumer struct right?

Whoops sorry, it should be `s.walletPubKey == walletPubkey || (s.walletPubKey == s.svc.keys.GetNostrPublicKey() && legacyAppCount == 0)`

But yeah we can leave the legacy info event now as it does no harm 👍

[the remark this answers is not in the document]
im-adithya: One concern though, can we just pass in the `app.WalletPubkey` in the eventPublisher instead of deriving the key for every app when some app is deleted? So we can only do key derivation once per delete after the `s.walletPubkey == properties["walletPubkey"]` check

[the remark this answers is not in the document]
frnandu: > > And we want to only do something for the specific one matching the walletPubkey we saved in deleteAppConsumer struct right?
> 
> Whoops sorry, it should be `s.walletPubKey == walletPubkey || (s.walletPubKey == s.svc.keys.GetNostrPublicKey() && legacyAppCount == 0)`
> 
> But yeah we can leave the legacy info event now as it does no harm 👍

We would have to get an up-to-date `legacyAppCount` from DB to check that it was the last legacy app being deleted. So I think the legacy event info can stay undeleted, no harm.

service/keys/keys.go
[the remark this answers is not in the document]
im-adithya: > `bip32.FirstHardenedChild + 0` is being used for SCB encryption, so +1 is used for wallet keys.

Oh nice, didn't know that, thanks 🙌

service/delete_app_consumer.go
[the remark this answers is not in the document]
frnandu: > One concern though, can we just pass in the `app.WalletPubkey` in the eventPublisher instead of deriving the key for every app when some app is deleted? So we can only do key derivation once per delete after the `s.walletPubkey == properties["walletPubkey"]` check

I was doing that originally, but @rolznz suggested to be consistent and pass `app.ID` and derive there.
The derivation impact on app deletion for all existing subscribers/apps shouldn't be much of an issue, unless you have 1000's of apps, right?

im-adithya:
For some reason, if you do a fresh install (complete setup and go to Home page), the app isn't quitting on doing Ctrl+C

frnandu:
> For some reason, if you do a fresh install (complete setup and go to Home page), the app isn't quitting on doing Ctrl+C

Oh sorry, I broke that when trying to handle `relay.Context().Done()` when there is no subscription on legacy app wallet pubKey.
Don't have much experience with go contexts, but I think it should be fixed now by using a `select` and handle both `case <-ctx.Done():` and `case <-relay.Context().Done():`
Let me know if you have a chance to re-test this and confirm it is working both doing Ctrl-C and by restarting a local test relay (should reconnect and re-subscribe each app).
Thanks for catching this!

service/delete_app_consumer.go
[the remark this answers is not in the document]
rolznz: Hmm... good point. Couldn't the delete app consumer store the app ID? then it can do a simple check before any deriving.

rolznz:
> > For some reason, if you do a fresh install (complete setup and go to Home page), the app isn't quitting on doing Ctrl+C
> 
> Oh sorry, I broke that when trying to handle `relay.Context().Done()` when there is no subscription on legacy app wallet pubKey. Don't have much experience with go contexts, but I think it should be fixed now by using a `select` and handle both `<-ctx.Done()` and `<-relay.Context().Done()` Let me know if you have a chance to re-test this and confirm it is working both doing Ctrl-C and by restarting a local test relay (should reconnect and re-subscribe each app). Thanks for catching this!

@frnandu that sounds like a good solution. Thanks @im-adithya !

rolznz (approved):
tACK

service/delete_app_consumer.go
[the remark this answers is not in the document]
im-adithya: Yeah we can do that if required later 👍

service/start.go
```diff
@@ -113,6 +146,73 @@ func (svc *service) startNostr(ctx context.Context) error {
...
+
+	if sub.Relay.ConnectionError != nil {
+		logger.Logger.WithField("connectionError", sub.Relay.ConnectionError).Error("Relay error")
+		return sub.Relay.ConnectionError
+	}
+	logger.Logger.Info("Exiting subscription...")
+	return nil
+}
+
 func (svc *service) StartApp(encryptionKey string) error {
```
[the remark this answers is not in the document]
rolznz: @im-adithya we can't do that because launching the backend relies on the keys.

What about we change this check to also check if the node was run before, and only then start the app?

```
if appConfig.AutoUnlockPassword != "" {
		svc.StartApp(appConfig.AutoUnlockPassword)
	}
```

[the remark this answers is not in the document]
im-adithya: Nice that's better 🙌

PR #739, opened 2024-10-16
fix: add hostname check to decide http mode in windows

im-adithya:
## Description
Unlike Linux and Darwin, Windows doesn't allow wails:// protocol without having NSIS installer. So wails uses http://wails.localhost in Windows. This leads to the HTTP backup endpoint being requested in Windows which results in `405 Method Not Allowed`

### DIFF
--- a/frontend/src/components/ExternalLink.tsx
+++ b/frontend/src/components/ExternalLink.tsx
@@ -1,4 +1,5 @@
 import { Link } from "react-router-dom";
+import { isHttpMode } from "src/utils/isHttpMode";
 import { openLink } from "src/utils/openLink";
 
 type Props = {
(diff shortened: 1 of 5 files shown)


bumi:
good debugging. 
how is the HTTP endpoint being requested? what URL is exposed there?

frontend/src/components/ExternalLink.tsx
```diff
@@ -8,7 +8,9 @@ type Props = {
 };
 
 export default function ExternalLink({ to, className, children }: Props) {
-  const isHttpMode = window.location.protocol.startsWith("http");
+  const isHttpMode =
```
rolznz, line 11: can we extract this check now?

rolznz: Actually, I'm not sure we should do this and actually just use a NSIS installer, if that is not much extra work

rolznz: We decided to keep as-is. Since the check is more complicated now I think extracting it into `frontend/src/utils/isHttpMode.ts` would be good.

And we said we might want to do a hotfix. I will make a new milestone for this.

im-adithya: Done!

rolznz (approved):
utACK

PR #742, opened 2024-10-18
fix: use memory config

matjaz:


matjaz:
https://fly.io/docs/reference/configuration/#memory

Issue #752, opened 2024-10-23
Improve get balances endpoint


rolznz:
Currently the balances endpoint does not show LDK information about closing channels which:
- we do not handle balances properly in some edge cases, leading to https://github.com/getAlby/hub/issues/612
- we miss essential information to understand the state of currently closing channels (who we are closing with, are the funds locked or ready to be spent, does it need to be bumped?)

By improving this endpoint to have a internal property like the channels one (so different LNClients can add different info) we can then use the debug tools to debug when one of the above cases happens.

Issue #612, opened 2024-09-04
Bug: Doubled incoming balance after closing channel (LDK)


rolznz:
![image](https://github.com/user-attachments/assets/53cdc546-6d54-4844-bf42-23e173e42579)

jpl-btc:
This had a follow up -> The visualization bug part 2:
Finally those funds were settled and are in my Savings Balance.
However, it seems The Pending Closed Channels notification, continue to be there, and there is a ghost sum of exactly half of what appeared before on the incomings sats small grey section of the Savings Balance box.
![image](https://github.com/user-attachments/assets/c478fdd3-6d0c-483e-991e-12bd1260f8a2)

jpl-btc:
Part 3: After some blocks confirmations, the problems of part 2 disappeared. And now everything is as it should be.
![image](https://github.com/user-attachments/assets/4c714553-84f9-45e7-bb67-57cc6d304e51)

rolznz:
There is probably some overlap in `LightningBalances` and `PendingBalancesFromChannelClosures` https://github.com/getAlby/hub/blob/master/lnclient/ldk/ldk.go#L1000

If so, we could check the channel ID to ensure the balance from the same channel is not added twice

jpl-btc:
If this happens again I'll record it

rolznz:
This is worse than I thought because if the closing channel TX is not bumped, this could stay showing the wrong amount for many hours

PR #755, opened 2024-10-24
fix: do not show duplicated incoming funds from channel closures

rolznz:
Fixes https://github.com/getAlby/hub/issues/752
Fixes https://github.com/getAlby/hub/issues/612

There is overlap in balance from pending closed channels and total onchain balance.

They will show as incoming in the total onchain balance once the channel becomes sweepable. Therefore, this PR no longer manually includes balances from pending closed channels in the "incoming" amount on the savings balance card.

This PR also adds extra balance info that can be retrieved from debug tools -> get balances which is useful for debugging channels that are stuck (e.g. unconfirmed) or to know if the funds on a channel are still locked, and which counterparty the channel is with.

### DIFF
--- a/frontend/src/screens/channels/Channels.tsx
+++ b/frontend/src/screens/channels/Channels.tsx
@@ -318,15 +318,11 @@ export default function Channels() {
                   {new Intl.NumberFormat().format(balances.onchain.spendable)}{" "}
                   sats
                   {balances &&
-                    (balances.onchain.spendable !== balances.onchain.total ||
-                      balances.onchain.pendingBalancesFromChannelClosures >
-                        0) && (
+                    balances.onchain.spendable !== balances.onchain.total && (
                       <p className="text-xs text-muted-foreground animate-pulse">
                         +
                         {new Intl.NumberFormat().format(
-                          balances.onchain.total -
-                            balances.onchain.spendable +
-                            balances.onchain.pendingBalancesFromChannelClosures
+                          balances.onchain.total - balances.onchain.spendable
                         )}{" "}
                         sats incoming
                       </p>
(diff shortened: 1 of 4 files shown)


frontend/src/screens/channels/Channels.tsx
```diff
@@ -318,15 +318,11 @@ export default function Channels() {
...
-                      balances.onchain.pendingBalancesFromChannelClosures >
-                        0) && (
+                    balances.onchain.spendable !== balances.onchain.total && (
                       <p className="text-xs text-muted-foreground animate-pulse">
                         +
                         {new Intl.NumberFormat().format(
-                          balances.onchain.total -
-                            balances.onchain.spendable +
-                            balances.onchain.pendingBalancesFromChannelClosures
+                          balances.onchain.total - balances.onchain.spendable
```
im-adithya, line 325: The docs for `pending_balances_from_channel_closures` say:

> ...depending on the sync status of the wallets, swept balances listed here might or might not already be accounted for in [total_onchain_balance_sats](https://docs.rs/ldk-node/latest/ldk_node/struct.BalanceDetails.html#structfield.total_onchain_balance_sats)

I searched if there is any method which provides the sync status so it can be used in the get balances method but couldn't find any, do you know of such method?

Not sure if it's worth the effort, but from this change, my understanding is that if LDK and BDK are not in sync, then the user might not even see a dialog/incoming message after a channel is closed and is missing from the list like they do today for sometime (till they sync again)

rolznz: Nice find!

I think LDK and BDK are in sync, but it means there is some possible overlap in funds for a short period of time.

Because of this possibility that for a short period of time you will see funds from closed channels in two places, we now never sum the pending balances with the total balance, to avoid showing duplicated funds.

im-adithya (approved):

PR #768, opened 2024-10-30
feat: do not require unlock password to recover encrypted scb

rolznz:
This PR changes the how the Alby Account - specific static channel backups are encrypted to no longer require the unlock password, by deriving it from the seed. Therefore, the user only needs to backup their recovery phrase.

TODOs:
- [x] change encryption (no need to re-generate key)
- [x] should mnemonic exist for all backend types? _not solving in this PR_
- [x] review derivation logic
- [x] update derivation paths
  - [x] It's very unlikely, but possible, that a given index does not produce a valid private key. Error checking is skipped in this example for brevity but should be handled in real code. In such a case, a ErrInvalidPrivateKey is returned. - what to do here if we use a hardcoded index for the backup? wait for the issue to happen (this will occur on node startup if ever, since we also store a channel backup on node startup)?
- [x] review change to channel backup description
- [x] update copy on backup page
- [x] local E2E test


service/keys/keys.go
```diff
@@ -51,3 +88,8 @@ func (keys *keys) GetNostrPublicKey() string {
 func (keys *keys) GetNostrSecretKey() string {
 	return keys.nostrSecretKey
 }
+
+func (keys *keys) EncryptChannelBackupData(channelBackupData string) (string, error) {
+	// FIXME: this is not the right way to encrypt the data (we already have a key, no need to generate a new one)
+	return config.AesGcmEncrypt(channelBackupData, keys.encryptedChannelsBackupKey)
```
bumi, line 94: I think this should not be used there (or needs refactoring) the key should not go through the `DeriveKey` function which creates a key from a password string using argon2. 

we already have a key here, don't we?
or is that how one generally does this?

service/keys/keys.go
```diff
@@ -41,6 +48,36 @@ func (keys *keys) Init(cfg config.Config, encryptionKey string) error {
...
+			return err
+		}
+
+		ENCRYPTED_SCB_INDEX := uint32(0) // TODO: choose an index
+		encryptedChannelsBackupKey, err := appKey.NewChildKey(ENCRYPTED_SCB_INDEX)
+		if err != nil {
+			logger.Logger.WithError(err).Error("Failed to create seed from mnemonic")
+			return err
+		}
+		keys.encryptedChannelsBackupKey = hex.EncodeToString(encryptedChannelsBackupKey.Key)
```
bumi, line 78: do we need to store those on startup or is it enough to store the master key and then derive keys on the fly where ever it is needed. 

I am wondering if this `keys` part of the application needs to know about the `backup` part. 
ideally it's all a small piece that functions independently and knows as little as possible from other parts I think. With this it is universal/abstract and can be reused.

bumi, line 78: do we need to store those on startup or is it enough to store the master key and then derive keys on the fly where ever it is needed. 

I am wondering if this `keys` part of the application needs to know about the `backup` part. 
ideally it's all a small piece that functions independently and knows as little as possible from other parts I think. With this it is universal/abstract and can be reused.

rolznz: I updated it, by exposing a DeriveKey function which will derive child keys from the app key (which is derived from the master key). What do you think?

config/aesgcm.go
```diff
@@ -78,3 +79,60 @@ func AesGcmDecrypt(ciphertext string, password string) (string, error) {
 
 	return string(plaintext), nil
 }
+
+func AesGcmEncryptWithKey(plaintext string, key []byte) (string, error) {
```
bumi, line 83: we probably could use this function from the `AesGcmEncryptWithPassword` function? 
`AesGcmEncryptWithPassword` just also generates a key from the password, the rest is the same, isn't it

alby/alby_oauth_service_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/getAlby/hub/tests"
+	"github.com/stretchr/testify/assert"
+	"github.com/tyler-smith/go-bip32"
+	"github.com/tyler-smith/go-bip39"
+)
+
+func TestExistingEncryptedBackup(t *testing.T) {
+	defer tests.RemoveTestService()
+	svc, err := tests.CreateTestService()
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
rolznz: Thanks! - fixed in all tests (probably need to review some other assertions, but can be done in a follow-up PR)

alby/alby_oauth_service_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	mnemonic := "limit reward expect search tissue call visa fit thank cream brave jump"
+	unlockPassword := "123"
+	svc.Cfg.SetUpdate("Mnemonic", mnemonic, unlockPassword)
+	err = svc.Keys.Init(svc.Cfg, unlockPassword)
+	assert.Nil(t, err)
```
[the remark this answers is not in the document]
rolznz: Yes, fixed in all places. Thanks!

PR #773, opened 2024-11-02
chore: extend startup timeout

bumi:
startups can take some time. It should be faster but I think it is worse when a user submits the startup again.


rolznz (approved):
utACK

Issue #143, opened 2024-04-07
Add VSS for new users who use an Alby Account


rolznz:
https://lightningdevkit.org/blog/ldk-roadmap/#vss

Original ticket: https://github.com/getAlby/hub/issues/203

PR #779, opened 2024-11-05
feat: auth with alby account before node setup

rolznz:
Part of https://github.com/getAlby/hub/issues/143

This moves the "Connect Alby Account" step from after the node is launched to being after the user clicks "Get Started" or "Advanced Setup". Once the auth screen is passed, the setup flow returns to normal.

This allows us to know if the user is authenticated before setting up the node, which is required because the node will be launched differently if VSS is used.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path"
 
 	"github.com/getAlby/hub/db"
 	"github.com/getAlby/hub/logger"
@@ -117,19 +116,12 @@ func (cfg *config) init(env *AppConfig) error {
 }
 
 func (cfg *config) SetupCompleted() bool {
-	// TODO: remove AlbyUserIdentifier and hasLdkDir checks after 2025/01/01
-	// to give time for users to update to 1.6.0+
-	albyUserIdentifier, _ := cfg.Get("AlbyUserIdentifier", "")
 	nodeLastStartTime, _ := cfg.Get("NodeLastStartTime", "")
-	ldkDir, err := os.Stat(path.Join(cfg.GetEnv().Workdir, "ldk"))
-	hasLdkDir := err == nil && ldkDir != nil && ldkDir.IsDir()
 
 	logger.Logger.WithFields(logrus.Fields{
-		"has_ldk_dir":              hasLdkDir,
-		"has_alby_user_identifier": albyUserIdentifier != "",
 		"has_node_last_start_time": nodeLastStartTime != "",
 	}).Debug("Checking if setup is completed")
-	return albyUserIdentifier != "" || nodeLastStartTime != "" || hasLdkDir
+	return nodeLastStartTime != ""
 }
 
 func (cfg *config) GetJWTSecret() string {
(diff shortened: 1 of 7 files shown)


im-adithya:
Then it might be good to also mention this VSS benefit users get by conencting their Alby Account visibly here (using some color works I think):

Screenshot 2024-11-06 at 8 40 27 PM

frontend/src/components/redirects/HomeRedirect.tsx
```diff
@@ -13,31 +13,49 @@ export function HomeRedirect() {
...
+            to = "/alby/auth";
+          }
         } else {
-          to = "/alby/auth";
+          to = "/unlock";
         }
+      } else if (info.setupCompleted && !info.running) {
+        to = "/start";
+      } else if (info.albyAccountConnected) {
+        to = "/welcome";
```
im-adithya, line 44: If we are adding this, then we should also skip showing `ConnectAlbyAccount` screen if it is connected already. Can be done by adding a small `if (!info.albyAccountConnected)` check at the top in `navigateToAuthPage`

rolznz:
> Then it might be good to also mention this VSS benefit users get by conencting their Alby Account visibly here (using some color works I think):

Ah, right. @reneaaron had an item for backups and we took it out, but it would be good to add back in. Maybe could go in https://github.com/getAlby/hub/pull/780 (I will add a TODO item)

frontend/src/components/redirects/HomeRedirect.tsx
rolznz: Note: This particular fix was put in for if the user logs in, then navigates back. Because of how the OAuth redirect works, the back button actually takes them to `/`, which then redirects them to the `intro` again. It is possibly a bit of an edge case but I thought the UX was just so bad there (and is hard to fix due to the OAuth redirect).

im-adithya (approved):
LGTM

rolznz:
> LGTM (not sure if we want to skip connect alby screen if already conected)

@im-adithya I think it would be good to fix, thanks for testing that.

frontend/src/components/redirects/HomeRedirect.tsx
rolznz: Done, and added comments to both

im-adithya: 🙌

config/config.go
```diff
@@ -117,19 +117,17 @@ func (cfg *config) init(env *AppConfig) error {
 }
 
 func (cfg *config) SetupCompleted() bool {
-	// TODO: remove AlbyUserIdentifier and hasLdkDir checks after 2025/01/01
+	// TODO: remove hasLdkDir check after 2025/01/01
```
im-adithya, line 120: Just curious, what changed after v1.6.0? Why can't we remove both checks at once?

rolznz: In 1.6.0 we made the Alby Account optional, and the existence of the Alby Account was used to check if the user had setup their node (as it was the last thing that happened after setting up the node)

When we made the Alby Account optional, we instead introduced the `NodeLastStartTime` config entry which we use to detect if they ever went through the setup.

So for backward compatibility, we had these old checks. But now, the Alby Account can be connected before the node is setup, so I had to remove that one. I was going to remove both, but I decided to keep the LDK directory check in as this is more critical to ensure the user does not go through the setup again (and wipe their old data)

PR #780, opened 2024-11-06
feat: enable VSS for new hubs with alby subscription

rolznz:
Closes https://github.com/getAlby/hub/issues/143

This allows a LDK node to be launched with a remote VSS store rather than local sqlite store. This only works for brand new hubs, and is currently only enabled for paid subscribers. The VSS code could be altered to allow users to provide their own server, but I think this is lower priority right now and doubt many people would use it.

Right now VSS only applies if:
- The node has never been started (brand new setup)
- The user has a paid subscription
- The LDK_VSS_URL env variable is set 

TODOs:
- [x] Update Alby vss token generation endpoint (not sure what it is yet)
  - [x] Should we pass a hash of a derived key from the seed when generating a token so users can run multiple hubs with the same Alby Account? (VSS does not support using the same user_token for multiple seeds)
- [x] end to end test (local)
- [x] disable disconnect account for users with VSS data while VSS data cannot be migrated


lnclient/ldk/ldk.go
```diff
@@ -1772,3 +1786,18 @@ func forceCloseChannelsFromStaticChannelsBackup(node *ldk_node.Node, staticChann
...
+	if err != nil {
+		return "", err
+	}
+
+	pubkeyBytes, err := key.PublicKey().Serialize()
+	if err != nil {
+		return "", err
+	}
+
+	return hex.EncodeToString(pubkeyBytes[0:3]), nil
```
bumi, line 1802: why such a short identifier? and not the whole?

service/start.go
```diff
@@ -280,10 +280,47 @@ func (svc *service) launchLNBackend(ctx context.Context, encryptionKey string) e
...
+				return err
+			}
+			if albyUserIdentifier != "" {
+				me, err := svc.albyOAuthSvc.GetMe(ctx)
+				if err != nil {
+					logger.Logger.WithError(err).Error("Failed to fetch alby user")
+					return err
+				}
+				// only activate VSS for Alby paid subscribers
+				if me.Subscription.Buzz {
```
bumi, line 302: can we extract this check somewhere to keep the start function small here. 
and potentially also easier to run it independently of Alby (if somebody wants to self-host a VSS?)

service/start.go
```diff
@@ -280,10 +280,47 @@ func (svc *service) launchLNBackend(ctx context.Context, encryptionKey string) e
...
+
+		vssToken := ""
+		vssEnabled, _ := svc.cfg.Get("LdkVssEnabled", "")
+		if vssEnabled == "true" {
+			vssNodeIdentifier, err := ldk.GetVssNodeIdentifier(svc.keys)
+			if err != nil {
+				logger.Logger.WithError(err).Error("Failed to get VSS node identifier")
+				return err
+			}
+			vssToken, err = svc.albyOAuthSvc.GetVssAuthToken(ctx, vssNodeIdentifier)
```
bumi, line 316: is this somehow possible to make it generic here that people could potentially self-host VSS?

lnclient/ldk/ldk.go
rolznz: It's a 6-digit hex number e.g. `0488b2`, which should be a one in over a million chance of having a collision if the same user runs with a different seed. I am not sure there is any need to include the whole derived pubkey on this specific path and makes the identifier very long. What do you think? It is only used to ensure the user can run multiple nodes with the same alby account.

service/start.go
rolznz: We decided for now to keep it as-is until we have a concrete usecase (running VSS for a single user may not be worthwhile for most users)

PR #788, opened 2024-11-11
fix: simplify friends & family app

### LINKED ISSUE #756: Simplify Friends & Family App
Currently we show many options in an accordion that a user has to choose from.

Instead, we could show the QR + copy button (the content of the Alby Go option right now) and show optional guides below (Possibly in an "Advanced" or "View specific guides" expandable section or something, so that users are not overwhelmed with choices by default.

Suggested by @jpl-btc

reneaaron:
Fixes #756 

 - Unfold the first option
 - Remove duplicate copy


rolznz (approved):
utACK

Issue #781, opened 2024-11-07
Restrict migration of funds from shared wallet to node to 1M sats per transaction


rolznz:
Currently we have some users with much more than that, which makes it less likely the payment will succeed. Some users also might not want to transfer all their funds when testing Alby Hub (currently the first channel flow does repeatedly migrate funds to drain the wallet though).

PR #794, opened 2024-11-13
fix: limit funds transferred from shared wallet to hub node

rolznz:
Fixes https://github.com/getAlby/hub/issues/781

- renamed "migrate" to "transfer" to match getalby.com/node page 
- when transferring funds on first channel open, will not repeatedly drain funds (except when user only has small amount of funds left over)
- transfer funds endpoint now caps the amount to transfer to 1M sats.


bumi:
lgtm! great

Issue #759, opened 2024-10-26
BuzzPay PoS incorrectly passes name property


rolznz:
When you copy the URL it starts with `https://pos.albylabs.com/?name=Test%20123%20123#/wallet/nostr%2` which does not work since we use a hash router.

This should also be updated to the new PoS URL format (see https://github.com/getAlby/pos/pull/29)

PR #797, opened 2024-11-13
fix: use buzzpay new URL format

rolznz:
Fixes https://github.com/getAlby/hub/issues/759

`name` param is not read correctly and needs fixing on the PoS side.

### DIFF
--- a/frontend/src/screens/internal-apps/BuzzPay.tsx
+++ b/frontend/src/screens/internal-apps/BuzzPay.tsx
@@ -51,7 +51,7 @@ export function BuzzPay() {
         });
 
         setPosUrl(
-          `https://pos.albylabs.com/?name=${encodeURIComponent(name)}#/wallet/${encodeURIComponent(createAppResponse.pairingUri)}/new`
+          `https://pos.albylabs.com/#/?nwc=${btoa(createAppResponse.pairingUri)}&name=${encodeURIComponent(name)}`
         );
 
         await reloadApps();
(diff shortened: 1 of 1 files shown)


bumi:
I am not sure if I can fix the query param on the POS side. maybe I ask some AI :) 

should we use `name` or `label` - I think internally we call it label there.

rolznz:
@bumi added here: https://github.com/getAlby/pos/commit/86363cf55eae67bbe74664be3643885edd36873d also to the share link: https://github.com/getAlby/pos/commit/08ae31c6fbb4054974d70d33c2d2838101bade89

frontend/src/screens/internal-apps/BuzzPay.tsx
```diff
@@ -51,7 +51,7 @@ export function BuzzPay() {
         });
 
         setPosUrl(
-          `https://pos.albylabs.com/?name=${encodeURIComponent(name)}#/wallet/${encodeURIComponent(createAppResponse.pairingUri)}/new`
+          `https://pos.albylabs.com/#/?nwc=${btoa(createAppResponse.pairingUri)}&name=${encodeURIComponent(name)}`
```
rolznz, line 54:
```suggestion
          `https://pos.albylabs.com/#/?nwc=${btoa(createAppResponse.pairingUri)}&label=${encodeURIComponent(name)}`
```

rolznz:
> should we use name or label - I think internally we call it label there.

It supports both, but I agree it makes more sense as label. I've updated it.

bumi:
🚀

Issue #782, opened 2024-11-07
Add encrypted backups benefit to ConnectAlbyAccount page


rolznz:
This is a great feature of the Alby Account that people should know about.

PR #798, opened 2024-11-14
feat: add encrypted backups card to connect alby account page

rolznz:
Fixes https://github.com/getAlby/hub/issues/782

A temporary addition without a major redesign.

image

### DIFF
--- a/frontend/src/screens/ConnectAlbyAccount.tsx
+++ b/frontend/src/screens/ConnectAlbyAccount.tsx
@@ -59,7 +59,7 @@ export function ConnectAlbyAccount({ connectUrl }: ConnectAlbyAccountProps) {
           <Card className="w-full">
             <CardHeader className="flex flex-col justify-center items-center text-center p-4">
               <LifeBuoy className="w-6 h-6" />
-              <CardTitle className="text-sm">Real Support</CardTitle>
+              <CardTitle className="text-sm">Support</CardTitle>
               <CardDescription className="text-xs">
                 Human support via live chat when you need a helping hand
               </CardDescription>
--- a/frontend/src/screens/ConnectAlbyAccount.tsx
+++ b/frontend/src/screens/ConnectAlbyAccount.tsx
@@ -1,9 +1,9 @@
 import {
   CreditCard,
+  DatabaseBackup,
   Headphones,
   LifeBuoy,
   Mail,
-  PartyPopper,
   Zap,
 } from "lucide-react";
 import Container from "src/components/Container";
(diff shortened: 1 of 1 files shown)


frontend/src/screens/ConnectAlbyAccount.tsx
```diff
@@ -47,10 +47,19 @@ export function ConnectAlbyAccount({ connectUrl }: ConnectAlbyAccountProps) {
               </CardDescription>
             </CardHeader>
           </Card>
+          <Card className="w-full">
```
rolznz, line 50: This will only work for LDK

rolznz: Also, we do not know at startup if they will choose LDK, because this screen is shown before picking the backend type.

rolznz: I don't know how to do this without a*

* Encrypted backups are only available for the Alby Hub embedded node. Read more. (guide link)

Thoughts? @bumi @reneaaron

bumi: I was actually thinking of adding some "security" page. Which describes all the details from the importance of the password, the seed, and also these backup options. 
So we can link there and people have to acknowledge it similar to a ToS.

frontend/src/screens/ConnectAlbyAccount.tsx
```diff
@@ -47,10 +47,19 @@ export function ConnectAlbyAccount({ connectUrl }: ConnectAlbyAccountProps) {
...
+              <CardDescription className="text-xs">
+                Ensures you can always recover funds from lightning channels
+              </CardDescription>
+            </CardHeader>
+          </Card>
           <Card className="w-full">
             <CardHeader className="flex flex-col justify-center items-center text-center p-4">
               <LifeBuoy className="w-6 h-6" />
-              <CardTitle className="text-sm">Support</CardTitle>
+              <CardTitle className="text-sm">Real Support</CardTitle>
```
bumi, line 62: Not sure about the word "real" here. is the other one "fake"? :)

bumi (approved):

frontend/src/screens/ConnectAlbyAccount.tsx
rolznz: @bumi that sounds great. I think we also want to advertise backups as a feature worth having an Alby Account though, right?

frontend/src/screens/ConnectAlbyAccount.tsx
rolznz: True, I was thinking about chatbots. But I will undo this change.

frontend/src/screens/ConnectAlbyAccount.tsx
rolznz: I will move your idea to a new issue.

bumi: yeah, I think this change is good. I don't think we need the disclaimer here when we talk more about security and backups in detail and link to that.

frontend/src/screens/ConnectAlbyAccount.tsx
rolznz, line 62:
```suggestion
              <CardTitle className="text-sm">Support</CardTitle>
```

Issue #638, opened 2024-09-11
Include invoice description in permission denied event


rolznz:
When I received an email notifying me that I didn't have enough budget to make a payment, I didn't know what payment it was for (because it was automated).

Including the invoice description could help.

PR #805, opened 2024-11-19
feat: include invoice description in nwc payment permission denied event message

rolznz:
Fixes https://github.com/getAlby/hub/issues/638


transactions/transactions_service.go
```diff
@@ -795,12 +794,17 @@ func (svc *transactionsService) validateCanPay(tx *gorm.DB, appId *uint, amount
 			balance := queries.GetIsolatedBalance(tx, appPermission.AppId)
 
 			if amountWithFeeReserve > balance {
+				message := NewInsufficientBalanceError().Error()
+				if description != "" {
+					message += " Invoice description: " + description
```
bumi, line 799: do we need that "invoice description" addition?

rolznz: I guess not. I have removed it, thanks

Issue #723, opened 2024-10-04
Replace Bitcoin Connect with custom invoice payment component


rolznz:
Bitcoin Connect does not properly fit this usecase. We can just show a QR and copy button (similar to the onchain deposit component we have) . Also, we should clearly add a link to a guide if a user does not know how to pay the invoice.

In the guide we can include different ways to topup (e.g. pay with lightning from an exchange)

rolznz:
@jpl-btc can you help us prepare a guide for this?

rolznz:
@jpl-btc created a guide: https://guides.getalby.com/user-guide/alby-account-and-browser-extension/alby-hub/wallet/open-your-first-channel

rolznz:
idea: same style as the onchain QR on the deposit page, but directly below the QR add a clear "Learn about ways to pay" link which goes to the above guide.

jpl-btc:
The guide, while published, is still in draft form and has room for improvement.

I'd like to propose two ideas:

1. Add a "Help" button on both pages of the flow to assist users if they encounter issues. This button would redirect users to the "Open Your First Channel" guide. Each button will have context-specific copy that suits the user’s current environment, even though both link to the same guide: "What is this page, Learn more!" | "Unsure of this QR code? Find out!" -> the buttons should be visible for them to "work" and be clicked, perhaps a little bit of gray on the backround of the text, inside the button. something aligned with the current theme of course.

2. Move all bottom buttons closer to the channel creation steps. On the first page, these buttons should be nearer to the "Next" button (they’re barely visible now). On the second page, they should be closer to the QR code... especially the new button "Unsure of this QR code? Find out!" that links to the Open your first channel guide.

Here is a screenshot with the idea;
![image](https://github.com/user-attachments/assets/d4b79d8d-caad-4469-9686-14d381476d64)

rolznz:
@jpl-btc looks good (I think the "Unsure about this QR code" should look more like a clickable link ).You said you'd try work on this, right? it would be good to look at the onchain deposit page to get an idea of how the onchain QR code is implemented (since that is without bitcoin connect)

PR #819, opened 2024-11-21
feat: redesign first channel payment UI

rolznz:
Fixes https://github.com/getAlby/hub/issues/723 

Includes a link to @jpl-btc 's guide for people who do not know how to pay the invoice.

image

Bitcoin Connect is used in other places, however this is the most important place to improve. Bitcoin Connect should be removed in a follow-up PR.


frontend/src/screens/channels/first/FirstChannel.tsx
```diff
@@ -110,7 +110,7 @@ export function FirstChannel() {
             {new Intl.NumberFormat().format(channelSize)} sats of incoming
```
MoritzKa, line 110: Why is in the screenshot provided of this PR 100 sats instead of 100,000 sats in this phrase.

rolznz: Ah, when I was developing I mocked some values as the Megalith Mutinynet LSP service was down.

Issue #800, opened 2024-11-18
Auto refresh tls.cert from LND_CERT_FILE when it expires


frnandu:
If `tls.cert` gets refreshed with a new one when it expires we don't auto refresh from `LND_CERT_FILE`, and payments stop working.

rolznz:
@frnandu can you help with this?

PR #820, opened 2024-11-21
fix: Always update LND certificate

rolznz:
Fixes https://github.com/getAlby/hub/issues/800


bumi:
should we do this for the others (macaroon, address), too?

Issue #824, opened 2024-11-26
Panic in go-nostr Relay code causes Alby Hub to ungracefully exit


rolznz:
```
/go/pkg/mod/github.com/nbd-wtf/go-nostr@v0.35.0/relay.go:176 +0x206
created by github.com/nbd-wtf/go-nostr.(*Relay).ConnectWithTLS in goroutine 18918
/go/pkg/mod/github.com/nbd-wtf/go-nostr@v0.35.0/relay.go:180 +0xfd
github.com/nbd-wtf/go-nostr.(*Relay).ConnectWithTLS.func2()
goroutine 31429 [running]:
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xacb05d]
panic: runtime error: invalid memory address or nil pointer dereference
```

rolznz:
We need to catch panics in`start.go` by moving the relay connection request into a method with recovery (see `closeRelay` as an example in the same file.)
```
relay, err = nostr.RelayConnect(ctx, relayUrl, nostr.WithNoticeHandler(svc.noticeHandler))
```

PR #827, opened 2024-11-26
fix: recover from panics when connecting to relay

rolznz:
Fixes https://github.com/getAlby/hub/issues/824

### DIFF
--- a/service/start.go
+++ b/service/start.go
@@ -73,7 +73,7 @@ func (svc *service) startNostr(ctx context.Context) error {
 				"iteration": i,
 			}).Info("Connecting to the relay")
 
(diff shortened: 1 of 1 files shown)


bumi:
do you know what causes the panic?

rolznz:
The issue is that the connection is nil at https://github.com/nbd-wtf/go-nostr/blob/v0.35.0/relay.go#L180C38-L180C55 which has already been fixed in a newer version of go-nostr.

I will try to update to the latest version instead (we have some dependency issues with LND)

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/828

PR #829, opened 2024-11-27
feat: app-wide received payment notifications

rolznz:
Followup of https://github.com/getAlby/hub/pull/815 that will show a notification if you receive payments via lightning address.

I am not sure it's really needed though, or should just go on the receive page so that the rest of the app does not necessarily poll to get the latest transaction


rolznz:
@stackingsaunter approves this PR :grin:

PR #830, opened 2024-11-27
feat: show VSS state in settings

rolznz:
also adds some missing settings headers on other settings pages, and removes useless probing functions from the UI

image


bumi:
That's good. 
Because it only works with the Alby account maybe we can mention it there? 


```
Versioned Storage Service (VSS) provides a secure, encrypted server-side storage of essential lightning and onchain data. This service is enabled by your Alby account and provides additional backup security which allows you to recover your lightning data with your recovery phrase alone, without having to close your channels.
```

rolznz:
@bumi sounds good, thanks! I originally thought it would be good to let non-connected users know what they are missing out on, but I think it would be better to be done somehow in https://github.com/getAlby/hub/issues/782

rolznz:
image

MoritzKa:
Lago has a smart way of presenting premium features by showing the benefits
chrome_E8sR4pWfcT
For instance: Show VSS in the list of "Settings" for every LDK node Alby Hub users regardless if they connected an Alby account and explain the benefits on the page. Is that the case here?

Can we replace VSS in the left menu with: "Wallet Recovery" ?

PR #832, opened 2024-11-28
fix: re-add pending balance for LightningBalanceClaimableAwaitingConfirmations

rolznz:
(if the user force closes from their side this will show as pending funds from closed channels until the timelock ends)


bumi (commented):
sounds good

Issue #821, opened 2024-11-21
Allow disconnecting from Alby Account for users in blocked countries


rolznz:
Currently users will be blocked from accessing Alby Hub if they connected their account.
1. Allow disconnect on self-hosted auth page
2. Do not show Alby Account option in onboarding flow

There will be an endpoint to check if the user is blocked (TBC)

Once completed, also add an issue for the Alby Extension

PR #833, opened 2024-11-28
Feat: disconnect alby account in auth code form

rolznz:
Fixes https://github.com/getAlby/hub/issues/821

Uses the same functionality as the Alby Account settings page. This will only show for users using the default OAuth client (self-hosted hubs)

image

image


rolznz:
@bumi is there an API we can use to check if the account is blocked? then we can also only show this button in that case

bumi:
I think that's good, it gives control to the user

PR #834, opened 2024-11-29
feat: use nip44 and versioning

### LINKED ISSUE #737: Implement versioning and NIP-44 support
See https://github.com/nostr-protocol/nips/pull/1531

Notes:
- Alby Hub should indicate support for 1.0 and 0.0, and maintain support for 0.0 for the unforeseen future.
- This should be implemented in the Alby JS SDK (and therefore at least Alby Go + Alby Extension) to ensure the full flow is working.
- See how [NIP-47 notifications](https://github.com/nostr-protocol/nips/pull/1164) fits in with versioning.
- Add unit tests to ensure we maintain support for both versions and return the correct error code for unsupported versions.

im-adithya:
Fixes #737

Ready for review (use https://github.com/getAlby/js-sdk/pull/273)

- Notifications are published twice (with nip-04 and nip-44)~(so nip-44 doesn't work currently as the relays block 23917 kind)~ (Fixed by Fernando)
- ~Tests are to be added/modified~ (Done)


frnandu:
spec PR needs to adapt to the notifications (legacy and new) https://github.com/nostr-protocol/nips/pull/1531

nip47/event_handler.go
```diff
@@ -410,3 +462,40 @@ func (svc *nip47Service) publishResponseEvent(ctx context.Context, relay nostrmo
 
 	return nil
 }
+
+func (svc *nip47Service) isVersionSupported(version string, isNip04Encrypted bool) (bool, error) {
```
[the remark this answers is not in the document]
im-adithya: > why do we need isNip04Encrypted?

It is for these cases:

https://github.com/getAlby/hub/pull/834/files#diff-f539cac2183c47476e54f7827bf43aa74c6013b21e18d7ec52be6aab733fe960R493

where encryption is done using nip-04 but the version says `1.0+`. If we use versions then we can't communicate via nip44 with someone using nip04 to tell what the error is (this case should be mentioned in the spec)

nip47/event_handler.go
```diff
@@ -93,12 +96,21 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
 		}
 	}
 
-	ss, err := nip04.ComputeSharedSecret(app.AppPubkey, appWalletPrivKey)
+	var ss []byte
+	var ck [32]byte
+	isNip04Encrypted := strings.Contains(event.Content, "?iv=")
```
rolznz, line 101: I think we should check the version tag instead

nip47/event_handler.go
```diff
@@ -93,12 +96,21 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
...
+	if isNip04Encrypted {
+		ss, err = nip04.ComputeSharedSecret(app.AppPubkey, appWalletPrivKey)
+	} else {
+		ck, err = nip44.GenerateConversationKey(app.AppPubkey, appWalletPrivKey)
```
rolznz, line 105: I am wondering if we can better abstract the differences between NIP-04 and NIP-44, so the event handler does not need to know about it.

e.g. the event handler interacts with this abstraction and just passes the version number. Then the abstraction will either use NIP-04 or NIP-44, and all the conditionals will be isolated there.

And then we do not need to e.g. pass `ss` and `ck` separately, but have an encryption object that knows how to encrypt or decrypt based on the version it has?

nip47/event_handler.go
rolznz, line 466: Can you add a few tests for this method?

nip47/notifications/nip47_notifier.go
```diff
@@ -98,38 +99,40 @@ func (notifier *Nip47Notifier) notifySubscribers(ctx context.Context, notificati
...
-			"notification": notification,
-			"appId":        app.ID,
-		}).WithError(err).Error("Failed to compute shared secret")
-		return
+		go notifier.notifySubscriber(ctx, &app, notification, tags, appWalletPubKey, appWalletPrivKey, false)
+		go notifier.notifySubscriber(ctx, &app, notification, tags, appWalletPubKey, appWalletPrivKey, true)
 	}
+}
+
+func (notifier *Nip47Notifier) notifySubscriber(ctx context.Context, app *db.App, notification *Notification, tags nostr.Tags, appWalletPubKey, appWalletPrivKey string, legacy bool) {
```
rolznz, line 129: rather than passing `legacy bool` we could pass the version number, and then use this abstraction idea I mentioned to remove the large conditional.

nip47/notifications/nip47_notifier.go
```diff
@@ -139,39 +142,65 @@ func (notifier *Nip47Notifier) notifySubscriber(ctx context.Context, app *db.App
...
 	event := &nostr.Event{
 		PubKey:    appWalletPubKey,
 		CreatedAt: nostr.Now(),
-		Kind:      models.NOTIFICATION_KIND,
 		Tags:      allTags,
 		Content:   msg,
 	}
+	if legacy {
+		event.Kind = models.LEGACY_NOTIFICATION_KIND
+	} else {
```
rolznz, line 194: rather than the `else` you could put the `Kind` back in the original declaration, and only overwrite it if we use `0.0`

rolznz:
@im-adithya it would be good to add some tests, especially for any changes in the `event_handler`. It's critical that this is working and no vulnerabilities are added.

nip47/event_handler.go
im-adithya: Yeah was thinking of the same, should we make a encrypt/decrypt function that takes in event (which has version and pubkey), privkey and content which conditionally does nip04/44?

im-adithya:
Yup, already started adding tests, will push soon!

nip47/event_handler.go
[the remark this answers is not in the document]
rolznz: I don't think we should complicate our code a lot just to be able to give a better error message if people do not follow the spec and send completely the wrong encryption type.

nip47/event_handler.go
im-adithya: Done!

nip47/event_handler.go
im-adithya: Done, added `nip47/cipher/cipher.go`

nip47/notifications/nip47_notifier.go
im-adithya: Done

nip47/notifications/nip47_notifier.go
im-adithya: Done

im-adithya:
Feedback is addressed, only have to agree on the `isVersionSupported` function in cipher (currently it is a bit too complicated but is future-proof)

Also have to add tests.

rolznz:
@im-adithya if you also add some tests for isVersionSupported to ensure it passes the versions we support maybe it is ok for now and we don't need to simplify it (since you already wrote it)

rolznz:
I wonder if it's good to add the NIP-47 version to the app table, and default to 0.0 for existing entries. We do not republish info events for apps on startup. Later we might want to show which connections EDIT: did not signal they support NIP-44 so that the user can remove them and re-add them? CC @reneaaron @frnandu @bumi

tests/create_app.go
```diff
@@ -65,6 +65,6 @@ func CreateLegacyApp(svc *TestService, senderPrivkey string) (app *db.App, ss []
 		},
 	})
 
-	ss, err = nip04.ComputeSharedSecret(svc.Keys.GetNostrPublicKey(), senderPrivkey)
-	return app, ss, nil
+	nip47Cipher, err = cipher.NewNip47Cipher("1.0", svc.Keys.GetNostrPublicKey(), senderPrivkey)
```
im-adithya, line 68: Should legacy app return a cipher with v0? Or should we add `CreateAppV0` and `CreateAppV1` 😵‍💫

nip47/notifications/nip47_notifier_test.go
```diff
@@ -175,10 +176,10 @@ func doTestSendNotificationPaymentSent(t *testing.T, svc *tests.TestService, app
 	notifier := NewNip47Notifier(relay, svc.DB, svc.Cfg, svc.Keys, permissionsSvc, transactionsSvc, svc.LNClient)
 	notifier.ConsumeEvent(ctx, receivedEvent)
 
-	assert.NotNil(t, relay.PublishedEvent)
-	assert.NotEmpty(t, relay.PublishedEvent.Content)
+	assert.NotNil(t, relay.PublishedEvents[1])
+	assert.NotEmpty(t, relay.PublishedEvents[1].Content)
```
im-adithya, line 180: `[1]` because `[0]` is of kind `23196` and it can't be decrypted by v1.0 NIP-44 cipher

nip47/event_handler_test.go
```diff
@@ -55,13 +55,13 @@ func TestCreateResponse(t *testing.T) {
...
 
-	res, err := nip47svc.CreateResponse(reqEvent, nip47Response, nostr.Tags{}, ss, svc.Keys.GetNostrSecretKey())
+	res, err := nip47svc.CreateResponse(reqEvent, nip47Response, nostr.Tags{}, nip47Cipher, svc.Keys.GetNostrSecretKey())
 	assert.NoError(t, err)
 	assert.Equal(t, reqPubkey, res.Tags.GetFirst([]string{"p"}).Value())
 	assert.Equal(t, reqEvent.ID, res.Tags.GetFirst([]string{"e"}).Value())
 	assert.Equal(t, svc.Keys.GetNostrPublicKey(), res.PubKey)
 
-	decrypted, err := nip04.Decrypt(res.Content, ss)
+	decrypted, err := nip47Cipher.Decrypt(res.Content)
```
im-adithya, line 64: Here we only check 1.0 NIP-44 request event but not backwards compatibility with 0.0 NIP-04 request event, because we don't have access to the appWalletPrivKey here (although we can derive it here like we do in HandleEvent, but I feel it would look ugly then).

One solution I have in mind is to add a `LegacyEncrypt` and `LegacyDecrypt` methods to the cipher which can then help us decrypt it here using NIP-04

rolznz: I don't understand what you mean, we should have a separate test that creates a legacy request using NIP-04 that does not pass a version, and we should expect that one should use version 0.0 for the cipher

nip47/publish_nip47_info.go
```diff
@@ -37,7 +38,7 @@ func (svc *nip47Service) GetNip47Info(ctx context.Context, relay *nostr.Relay, a
 func (svc *nip47Service) PublishNip47Info(ctx context.Context, relay nostrmodels.Relay, appWalletPubKey string, appWalletPrivKey string, lnClient lnclient.LNClient) (*nostr.Event, error) {
 	var capabilities []string
 	var permitsNotifications bool
-	tags := nostr.Tags{[]string{}}
+	tags := nostr.Tags{[]string{"v", cipher.SUPPORTED_VERSIONS}}
```
im-adithya, line 41: I had to do this to avoid an import cycle (`cipher` imports `models` which imports `transactions` which imports `tests` which imports `cipher` again)

> `models` which imports `transactions`

I'm adding a commit which fixes the imports as this part doesn't make sense

nip47/event_handler_test.go
im-adithya: Yes, but the way the tests are written is such that we first use `CreateAppWithPrivateKey` to get the app along with a shared secret (which is now replaced with a cipher), hence the cipher version is set to "1.0" for all.

I think an even better way is to pass the version too to `CreateAppWithPrivateKey` which would make sense as you have to specify the app's nwc version during it's creation. Will update this, thanks 🙌

nip47/notifications/nip47_notifier_test.go
```diff
@@ -82,10 +83,17 @@ func doTestSendNotificationPaymentReceived(t *testing.T, svc *tests.TestService,
 	notifier := NewNip47Notifier(relay, svc.DB, svc.Cfg, svc.Keys, permissionsSvc, transactionsSvc, svc.LNClient)
 	notifier.ConsumeEvent(ctx, receivedEvent)
 
-	assert.NotNil(t, relay.PublishedEvent)
-	assert.NotEmpty(t, relay.PublishedEvent.Content)
+	var publishedEvent *nostr.Event
+	if cipher.Version == "0.0" {
```
im-adithya, line 87: Is this okay? cc @rolznz

Doesn't feel good to me

nip47/controllers/list_transactions_controller.go
```diff
@@ -67,13 +68,13 @@ func (controller *nip47Controller) HandleListTransactionsEvent(ctx context.Conte
 		return
 	}
 
-	transactions := []models.Transaction{}
+	txs := []models.Transaction{}
```
rolznz, line 71: can you undo this change? the transactions service should not know about nip47.

nip47/event_handler_legacy_test.go
rolznz, line 1: I think you're right, we shouldn't use "legacy" and now we are making it worse.

I suggest we have clearly separated tests:
- event_handler_nip44_test
- event_handler_nip04_test
- event_handler_no_wallet_pubkey_nip04_test
- event_handler_no_wallet_pubkey_nip44_test

rolznz: This is a critical part of our code, so I think it's worth adding more tests and clearly identifying what is being tested.

nip47/event_handler_test.go
rolznz: Cool. I think for legacy tests we should also support passing the version. This way we can test the 4 combinations.

nip47/event_handler_test.go
```diff
@@ -112,17 +125,22 @@ func TestHandleResponse_WithPermission(t *testing.T) {
 		Tags:      nostr.Tags{},
 		Content:   msg,
 	}
+
+	if version != "0.0" {
```
rolznz, line 129: I am worried about adding `ifs` in our tests like this for such a critical part. What about we do my suggestion of testing the 4 combinations separately?

rolznz: that or pass the tags in as an argument to this function

nip47/event_handler_test.go
```diff
@@ -148,13 +166,29 @@ func TestHandleResponse_DuplicateRequest(t *testing.T) {
 	defer tests.RemoveTestService()
 	svc, err := tests.CreateTestService()
 	require.NoError(t, err)
+
+	doTestHandleResponse_DuplicateRequest(t, svc, "1.0")
+}
+
+func TestHandleResponse_DuplicateRequest_LegacyVersion(t *testing.T) {
```
rolznz, line 173: `Legacy` is being mixed here. Let's be specific on what we are testing

nip47/notifications/nip47_notifier_test.go
rolznz: Yeah, we should try avoid `if`s in the test as much as possible. You could pass the index as an argument to this method, or the event kind, and get the event from the array matching the kind

nip47/notifications/nip47_notifier_test.go
```diff
@@ -112,22 +120,32 @@ func TestSendNotification_PaymentReceived(t *testing.T) {
...
 
-	app, ss, err := tests.CreateLegacyApp(svc, nostr.GeneratePrivateKey())
+	app, cipher, err := tests.CreateAppWithPrivateKey(svc, "", "0.0")
 	assert.NoError(t, err)
-	doTestSendNotificationPaymentReceived(t, svc, app, ss)
+	doTestSendNotificationPaymentReceived(t, svc, app, cipher)
 }
 
-func doTestSendNotificationPaymentSent(t *testing.T, svc *tests.TestService, app *db.App, ss []byte) {
+func TestSendNotification_LegacyApp_PaymentReceived(t *testing.T) {
```
rolznz, line 138: It would be good to change `LegacyApp` to `NoWalletPubkey` and do tests for both NIP04 and NIP44

tests/create_app.go
rolznz: Can you rename the function to CreateAppWithoutWalletPubkey and add a `nip47Version` argument?

db/models.go
```diff
@@ -25,6 +25,7 @@ type App struct {
 	UpdatedAt    time.Time
 	Isolated     bool
 	Metadata     datatypes.JSON
+	// TODO: add BroadcastedNip47Versions *string. For new apps, should be "1.0 0.0"
```
rolznz, line 28: @im-adithya what do you think of this?

So we know what version tag we used when we posted the info event on app creation / update.

nip47/controllers/list_transactions_controller.go
im-adithya: I did this for this reason: https://github.com/getAlby/hub/pull/834#discussion_r1901552774

Is there any other way around it?

nip47/notifications/nip47_notifier_test.go
im-adithya: > NoWalletPubkey

I got confused reading this comment thinking "how is it related to `The public key does not have a wallet connected` test", can we have a different name?

im-adithya: Does this sound good to you:
- `WithNip04Encryption`
- `WithNip44Encryption`
- `SharedWalletPubkey_WithNip04Encryption`
- `SharedWalletPubkey_WithNip44Encryption`

db/models.go
im-adithya: Yup this sounds good to me, +1 for adding this column.

But will this column have highest supported version ("1.0") or all the supported ones like "1.0 0.0"?

nip47/cipher/cipher.go
```diff
@@ -0,0 +1,112 @@
...
+		payload, err = nip44.Decrypt(content, c.conversationKey)
+		if err != nil {
+			return "", err
+		}
+	}
+	return payload, nil
+}
+
+func isVersionSupported(version string) (bool, error) {
+	versionParts := strings.Split(version, ".")
```
rolznz, line 81: I chatted with @bumi and the versioning is a real exception (and hopefully all apps update to nip44 so we can remove the v0.0 code)

So we think for now it's better just to check `version == "1.0" || version == "0.0"`

db/models.go
rolznz: I chatted with Bumi and it might just be ok to always re-publish the app info events on startup. Then we do not need to change the database.

CC @frnandu @reneaaron

tests/create_app.go
im-adithya: Done

nip47/controllers/list_transactions_controller.go
im-adithya: I undid the change, supported versions are fetched from cipher package now

nip47/event_handler_legacy_test.go
im-adithya: Added all tests, this is the format:
`_Nip04_`
`_Nip44_`
`_SharedWalletPubkey_Nip04_`
`_SharedWalletPubkey_Nip44_`

db/models.go
im-adithya: Added code to republish info event

nip47/cipher/cipher.go
im-adithya: Done!

nip47/event_handler_legacy_test.go
```diff
@@ -1,192 +1,88 @@
 package nip47
 
 import (
-	"context"
```
rolznz, line 4: can we rename this file to `event_handler_shared_wallet_pubkey_test`?

nip47/event_handler_test.go
```diff
@@ -39,7 +39,7 @@ func TestCreateResponse(t *testing.T) {
 
 	reqEvent.ID = "12345"
 
-	ss, err := nip04.ComputeSharedSecret(reqPubkey, svc.Keys.GetNostrSecretKey())
+	nip47Cipher, err := cipher.NewNip47Cipher("1.0", reqPubkey, svc.Keys.GetNostrSecretKey())
```
rolznz, line 42: Can we add another test method for `TestCreateResponse_Nip04` similar to this one?

nip47/event_handler_test.go
rolznz: @im-adithya can you address this and add a new parameter to the function for `tags`?

nip47/event_handler_test.go
```diff
@@ -182,35 +213,53 @@ func TestHandleResponse_DuplicateRequest(t *testing.T) {
...
 
 	nip47svc.HandleEvent(context.TODO(), relay, reqEvent, svc.LNClient)
 
-	assert.NotNil(t, relay.PublishedEvent)
-	assert.NotEmpty(t, relay.PublishedEvent.Content)
+	assert.NotNil(t, relay.PublishedEvents[0])
+	assert.NotEmpty(t, relay.PublishedEvents[0].Content)
 
-	relay.PublishedEvent = nil
+	relay.PublishedEvents = nil
```
rolznz, line 231: shouldn't this be an empty array?

nip47/event_handler_test.go
```diff
@@ -379,30 +525,35 @@ func TestHandleResponse_IncorrectPubkey(t *testing.T) {
...
-		Tags:      nostr.Tags{},
+		Tags:      nostr.Tags{[]string{"v", requestVersion}},
 		Content:   msg,
 	}
-	err = reqEvent.Sign(reqPrivateKey2)
-	assert.NoError(t, err)
 
-	// set a different pubkey (this will not pass validation)
-	reqEvent.PubKey = reqPubkey
+	if requestVersion != "0.0" {
```
rolznz, line 546: This looks wrong, and duplicated

nip47/event_handler_test.go
```diff
@@ -253,121 +307,213 @@ func TestHandleResponse_NoPermission(t *testing.T) {
...
+	// it shouldn't return anything for an invalid app key
+	assert.Nil(t, relay.PublishedEvents)
 }
 
-func TestHandleResponse_IncorrectPubkey(t *testing.T) {
+func TestHandleResponse_IncorrectVersions(t *testing.T) {
 	defer tests.RemoveTestService()
 	svc, err := tests.CreateTestService()
 	require.NoError(t, err)
+
```
rolznz, line 504:
```suggestion
// version specifies what cipher will use. If "1.0" is passed, cipher must be NIP-44, otherwise cipher MUST be NIP-04
```

nip47/event_handler_test.go
```diff
@@ -253,121 +307,213 @@ func TestHandleResponse_NoPermission(t *testing.T) {
...
+	doTestHandleResponse_IncorrectVersion(t, svc, "0.0", "1.0")
+	doTestHandleResponse_IncorrectVersion(t, svc, "1.0", "0.0")
```
rolznz, line 506: would it be worth adding a third test where NIP-44 is used but no version tag is passed?

service/start.go
```diff
@@ -162,7 +162,15 @@ func (svc *service) startAllExistingAppsWalletSubscriptions(ctx context.Context,
 
 	for _, app := range apps {
 		go func(app db.App) {
-			err := svc.startAppWalletSubscription(ctx, relay, *app.WalletPubkey)
+			// republish info event for all existing apps
+			walletPrivKey, err := svc.keys.GetAppWalletKey(app.ID)
```
rolznz, line 166: Don't we need a check `if app.WalletPubkey != nil {`?

service/start.go
```diff
@@ -162,7 +162,15 @@ func (svc *service) startAllExistingAppsWalletSubscriptions(ctx context.Context,
...
+			_, err = svc.GetNip47Service().PublishNip47Info(ctx, relay, *app.WalletPubkey, walletPrivKey, svc.lnClient)
+			if err != nil {
+				logger.Logger.WithError(err).WithFields(logrus.Fields{
```
rolznz, line 169: @frnandu it's ok if we just log an error and proceed here, right?

service/start.go
im-adithya: In the above SQL query we filter by the apps where `wallet_pubkey IS NOT NULL` right?

rolznz: Ah, right! thanks.

nip47/event_handler_test.go
im-adithya: Whoops sorry, might have missed while changing, what would you prefer though? Passing version tag everytime reduces a conditional but then we don't check the case where version is not specified, should we pass version all the time and have a separate test where we check if a request event without any version tag is considered as "0.0"?

rolznz: I guess we could do `if requestVersion != "" {` and then pass `""` for the third test. And instead of removing the line on this comment, we don't set the tags a few lines above

rolznz: > have a separate test where we check if a request event without any version tag is considered as "0.0"?
I think this would be good too

nip47/event_handler_test.go
im-adithya: True, will add it!

nip47/event_handler_legacy_test.go
im-adithya: Done!

nip47/event_handler_test.go
rolznz: Ah ok. Maybe it's fine as-is then. Thanks!

nip47/event_handler_test.go
im-adithya: It is of type `[]*nostr.Event`

nip47/event_handler_test.go
im-adithya: > I guess we could do if `requestVersion != "" {` and then pass `""` for the third test

I like this better since then we can check this for all other tests as well, what should we name it though 🤔 

`_Nip04_Unspecified`?

nip47/event_handler_test.go
im-adithya: Will also change it here to check `version != ""`

nip47/event_handler_test.go
im-adithya: Done!

rolznz (approved):
tACK

PR #835, opened 2024-11-29
feat: return transaction state and allow fetching unpaid transactions by state

rolznz:
This allows us to show **failed** and pending payments in Alby Go and other NWC clients.

The method of fetching `unpaid_outgoing: true` is not optimal, but this is a simple solution until we have a better idea and do not need to propose this as an official change to the spec.

### DIFF
--- a/nip47/controllers/list_transactions_controller.go
+++ b/nip47/controllers/list_transactions_controller.go
@@ -11,12 +11,14 @@ import (
 )
 
 type listTransactionsParams struct {
-	From   uint64 `json:"from,omitempty"`
-	Until  uint64 `json:"until,omitempty"`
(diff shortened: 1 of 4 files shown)


rolznz:
@bumi knowing the state of a payment is quite useful in my opinion. I had to re-login to Alby Hub on my phone after a payment got stuck in Alby Go - and I didn't know if it had failed or was stuck (we don't show either), to ensure I didn't try to double-pay.

I think it is definitely a good addition to the spec, I just think there might be a better way to design the request filtering than the way I have suggested.

bumi:
ok, great then...

rolznz:
Changes not proposed yet are listed at https://github.com/getAlby/hub/issues/380

PR #849, opened 2024-12-05
feat: fire channel open and closure events in lnd

### LINKED ISSUE #806: Fire channel closure events for LND
Currently we do not track channel closures, so users do not get notified like they do with LDK.

im-adithya:
Fixes #806

Adds a channel event susbcription to listen to channel closures.

`closureReason` can be the following:

- `"COOPERATIVE_CLOSE"`
- `"LOCAL_FORCE_CLOSE"`
- `"BREACH_CLOSE"`
- `"REMOTE_FORCE_CLOSE"`
- `"FUNDING_CANCELED"`
- `"ABANDONED"`

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -18,6 +18,7 @@ import (
 	decodepay "github.com/nbd-wtf/ln-decodepay"
 	"google.golang.org/grpc/status"
 
+	"github.com/getAlby/hub/config"
 	"github.com/getAlby/hub/events"
 	"github.com/getAlby/hub/lnclient"
 	"github.com/getAlby/hub/lnclient/lnd/wrapper"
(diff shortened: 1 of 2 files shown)


im-adithya:
We also can add a ChannelActive update publisher in the same PR but I'm not sure if it's the same as ChannelReady in LDK. Is ChannelReady triggered when peer connection is lost (making the channel inactive) and then connected again?

rolznz:
@im-adithya channel_ready in LDK is only triggered once, when the newly-opened channel is ready to be used

rolznz:
@im-adithya it would be good to also add the channel_ready event in this PR. Is there something equivalent in LND?

im-adithya:
> channel_ready in LDK is only triggered once

Ah then LND doesn't have any similar method, I'll see if there's any alternate way of doing this

rolznz:
@im-adithya possibly we can subscribe to the `ChannelActive` event but only fire our event if it's the first time it becomes active. To ensure this, we could check if the current number of block confirmations is smaller than or equal to 6. (there is a chance the channel goes offline and on again, maybe there's a chance we get a duplicate event. But I think this should be rare?)

im-adithya:
Sorry my bad, there actually is an open channel event as well 🤦 added now!

rolznz:
@im-adithya does that event run when the channel is opened or when it is opened AND the channel has the required confirmations?

im-adithya:
It only runs when the channel is ready to use (i.e when the min required block confirmations are done)

rolznz (approved):
utACK

Issue #75, opened 2024-06-07
Support paying 0-amount invoices


rolznz:
Currently we cannot pay default (0-amount) invoices generated by WoS, Phoenix or CashApp, Blink, etc.

- NIP-47 pay invoice supports 0-amount invoices
- Pay 0-amount invoices created by multiple wallets

I think the easiest way to start is just implement it for NWC and review what apps make sense to add the UI to (probably just Alby Hub and Alby mobile, maybe later the extension)

PR #860, opened 2024-12-08
feat: pay 0-amount invoices

rolznz:
Fixes https://github.com/getAlby/hub/issues/75

Currently only implemented in LDK. Includes a new screen for entering the amount (the same as the LNURL one, just without the ability to set a comment)

- [x] Update to use https://github.com/getAlby/js-lightning-tools/pull/181 (currently I have this branch linked locally)
- [x] Add support in LND LNClient - or could move to a separate issue

Tested by running 2 nodes locally and generating a 0-amount invoice from one and paying with the other. NOTE: there is no plan to add support for creating 0-amount invoices.


bumi:
you know how much I like zero-amount invoices :D 

will this work with the supported backends or will this need more `if` ?

http/http_service.go
```diff
@@ -438,7 +438,14 @@ func (httpSvc *HttpService) balancesHandler(c echo.Context) error {
 func (httpSvc *HttpService) sendPaymentHandler(c echo.Context) error {
 	ctx := c.Request().Context()
 
-	paymentResponse, err := httpSvc.api.SendPayment(ctx, c.Param("invoice"))
+	var payInvoiceRequest api.PayInvoiceRequest
```
im-adithya, line 441: Why not push the invoice to the `PayInvoiceRequest` struct?

im-adithya (approved):
Small nit regarding payload struct, otherwise tACK 👌

http/http_service.go
im-adithya: Hmmm, true should be good 👍

frontend/package.json
```diff
@@ -20,8 +20,8 @@
     "prepare": "cd .. && husky frontend/.husky"
   },
   "dependencies": {
-    "@getalby/lightning-tools": "^5.1.1",
     "@getalby/bitcoin-connect-react": "^3.6.2",
+    "@getalby/lightning-tools": "5.1.2",
```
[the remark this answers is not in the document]
im-adithya: Ah true, sorry for this!

SoulReaver-SS:
"NOTE: there is no plan to add support for creating 0-amount invoices." why?
That's where the meat is w/ 0 amount invoices, ability to generate them to receive flexible amounts in payment.

rolznz:
@SoulReaver-SS we believe lightning addresses are a better solution than 0-amount invoices

PR #869, opened 2024-12-09
fix: remove breaks in relay loop that could cause permanent disconnection from relay

rolznz:


service/start.go
```diff
@@ -123,7 +120,6 @@ func (svc *service) startNostr(ctx context.Context) error {
 					logger.Logger.WithError(err).Error("Got an error from the relay while listening to subscription.")
 					continue
 				}
-				break
```
im-adithya, line 126: I think this break is here because it will be only hit when `err == nil` (which we assumed that it happens when ctx is done). And for that to happen, `StartSubscription` should return nil err, which can only happen when subscription context ends and
- NOT because of relay connection error
- NOT because of main context being cancelled (because otherwise breaking should be valid as the app is anyways terminated)

Is it possible? How to reproduce this?

im-adithya: I think it happens when `sub.Relay.Write` returns an error ([here](https://github.com/nbd-wtf/go-nostr/blob/master/subscription.go#L159)) and we also have this: 
https://github.com/getAlby/hub/blob/a8101aac1307af3f5a825bd335a4f4ef6fe149e5/service/start.go#L136

So it should be possible. But in any case the fix makes sense as it is always good to check instead of breaking out by assuming things :+1:

frnandu: I tried to reproduce this, and besides invalid filters that the relay rejects and closed the subscription but not the socket (which should not happen when using our properly configured relay) the only possible thing that might be happening is some sort of network socket read/write timeouts which causes the subscription to be closed but no connection error on the relay level. I did not reproduce such situation, so this is not 100% confirmed.
This would previously be treated as a program exit and so the break would cancel the loop.
+1 for removing them and making the reconnect loop more reliable.

frnandu (approved):
I tried to reproduce this, and besides invalid filters that the relay rejects and closes the subscription but not the socket (which should not happen when using our properly configured relay) the only possible thing that might be happening is some sort of network socket read/write timeouts which causes the subscription to be closed but no connection error on the relay level. I did not reproduce such situation, so this is not 100% confirmed.
This would previously be treated as a program exit and so the break would cancel the loop.
+1 for removing them and making the reconnect loop more reliable.

service/start.go
```diff
@@ -123,7 +120,6 @@ func (svc *service) startNostr(ctx context.Context) error {
...
 			}
 			select {
 			case <-ctx.Done():
```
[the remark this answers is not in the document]
rolznz: I would like to keep it easy to understand by only having one exit point. The more places we can have breaks the more chance one of them doesn't work and incorrectly exits the loop.

rolznz:
@frnandu thanks. It definitely happens related to networking issue, more likely in self-hosted hubs.

service/start.go
rolznz: At first I think it would be good to log and see it happens. If we can find a way to reproduce it, then we can submit an issue or fix in go-nostr.

PR #877, opened 2024-12-11
fix: allow self-signed certificates and to skip during setup

### LINKED ISSUE #405: Bug: Self-signed LND certificates error in Windows Desktop
From https://primal.net/e/note16l86pvcfcqdkrpm7h0re58gn9420js2tszndxqh08xfgvqe2lxpqph82r0

> (1) There seems to be an issue when LND uses self-signed certificates, which is the default behavior. AlbyHub complains that the certificate isn't trusted, and this happens even if the self-signed certificate is added to the Trusted Root Authorities folder in the Windows Certs console. What I ended up doing was creating a dedicated DNS entry for LND (lnd.mydomain.com), and then getting a free SSL cert from one of the countless free providers, including that DNS name on the certificate. I forced LND to use that certificate instead of the self-signed one and then referenced it in the AlbyHub setup. That seemed to clear up the error.

im-adithya:
Fixes #405

## Description

EDIT: We decided that it's good to always verify if a certificate is passed by the user. If users raise concerns about using their own self-signed certificate (as it fails the verification), we'll think of adding a checkbox saying "I'm using a self signed certificate" which when checked can skip the cert verification process.

As we anyways ask the user to provide the certificate, it doesn't make any sense to verify the server certificate and host name, the same is also true for the case when the TLS certificate is not provided (as the user agrees to the risk it poses). So this adds `InsecureSkipVerify: true` in both cases while generating transport credentials.

(And yes, this PR also adds the possibility to continue without adding a TLS certificate)

## Screenshot
<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

### DIFF
--- a/lnclient/lnd/wrapper/lnd.go
+++ b/lnclient/lnd/wrapper/lnd.go
@@ -46,11 +46,16 @@ func NewLNDclient(lndOptions LNDoptions) (result *LNDWrapper, err error) {
 		if err != nil {
 			return nil, err
 		}
-		cp.AppendCertsFromPEM(cert)
-		creds = credentials.NewClientTLSFromCert(cp, "")
+		if !cp.AppendCertsFromPEM(cert) {
+			return nil, errors.New("failed to append certificate")
+		}
+		creds = credentials.NewTLS(&tls.Config{
+			ClientCAs:          cp,
+			InsecureSkipVerify: true,
+		})
 		// if a path to a cert file is provided
 	} else {
-		creds = credentials.NewTLS(&tls.Config{})
+		creds = credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})
 	}
 	opts := []grpc.DialOption{
 		grpc.WithTransportCredentials(creds),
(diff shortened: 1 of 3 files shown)


lnclient/lnd/wrapper/lnd.go
```diff
@@ -46,11 +46,16 @@ func NewLNDclient(lndOptions LNDoptions) (result *LNDWrapper, err error) {
...
 			return nil, err
 		}
-		cp.AppendCertsFromPEM(cert)
-		creds = credentials.NewClientTLSFromCert(cp, "")
+		if !cp.AppendCertsFromPEM(cert) {
+			return nil, errors.New("failed to append certificate")
+		}
+		creds = credentials.NewTLS(&tls.Config{
+			ClientCAs:          cp,
+			InsecureSkipVerify: true,
```
rolznz, line 54: I think we should verify the cert if it's provided. But everything else looks good!

rolznz (approved):
utACK

rolznz:
@im-adithya nice improvements!

PR #882, opened 2024-12-13
feat: pay 0 amount invoices in LND

im-adithya:
Adds 0 amount invoices for LND

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -20,8 +20,8 @@
     "prepare": "cd .. && husky frontend/.husky"
   },
   "dependencies": {
-    "@getalby/lightning-tools": "^5.1.2",
     "@getalby/bitcoin-connect-react": "^3.6.2",
+    "@getalby/lightning-tools": "5.1.2",
     "@radix-ui/react-accordion": "^1.2.0",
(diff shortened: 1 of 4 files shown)


frontend/package.json
```diff
@@ -20,8 +20,8 @@
     "prepare": "cd .. && husky frontend/.husky"
   },
   "dependencies": {
-    "@getalby/lightning-tools": "^5.1.2",
     "@getalby/bitcoin-connect-react": "^3.6.2",
+    "@getalby/lightning-tools": "5.1.2",
```
rolznz, line 24: Why this change?

im-adithya: I did yarn add @getalby/lightning-tools, and it corrected this, it should be in alphabetical order right?

rolznz: it changed it to the exact version, and this was already in the PR (You probably should have just did yarn install?), I think the change should be undone

im-adithya: Yeah, but isn't it the convention? To have the packages in sorted order? https://eslint.org/docs/latest/contribute/package-json-conventions#order

rolznz: The order is not the only change, you also changed the version from "^5.1.2" to "5.1.2"

PR #902, opened 2024-12-17
feat: cleanup old static channel backup files

rdmitr:


lnclient/ldk/ldk.go
```diff
@@ -1588,6 +1587,43 @@ func (ls *LDKService) saveStaticChannelBackupToDisk(event *events.StaticChannels
...
+		if strings.HasPrefix(name, "20") && filepath.Ext(name) == ".json" && !file.IsDir() {
+			backupFiles = append(backupFiles, file)
+		}
+	}
+
+	if len(backupFiles) <= keepBackupFiles {
+		return
+	}
+
+	// ReadDir() returns file entries listed alphabetically, so we can just
```
rolznz, line 1614: This is a bit scary to me, can we explicitly sort it in an external function that we can add unit tests for to ensure we only delete the oldest files?

bumi:
how much space will the backup take? 
If we create a SCB only if a channel is created and no longer on startup, do we then even need to clean it up in the app?

rolznz:
> how much space will the backup take?
If we create a SCB only if a channel is created and no longer on startup, do we then even need to clean it up in the app?

Hmm, maybe you are right and this is enough. My hub was a bit of an exception, currently I have 250MB of backups across 17 files

bumi:
Do we have an estimate of the average file we expect? 
If somebody has a lot of channels and activity maybe we can expect them to delete those files? 
and on average it just fills up a bit. I don't want to know what crap data is laying around on every computer where companies like Apple and Microsoft do not manage to clean up properly.

lnclient/ldk/ldk.go
rolznz: @rdmitr I think actually Bumi is right, maybe we convert this to draft for now and just stop saving new backups on every app start as a first step. I will create a separate PR for this.

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/943

Issue #530, opened 2024-08-22
Change OAuth Service to consume whitelisted events


rolznz:
Currently it blacklists certain events, which is easy to miss and process unintended events.

PR #903, opened 2024-12-18
chore: use whitelist for alby events

rolznz:
Fixes https://github.com/getAlby/hub/issues/530

### DIFF
--- a/alby/alby_oauth_service.go
+++ b/alby/alby_oauth_service.go
@@ -646,38 +646,15 @@ func (svc *albyOAuthService) ConsumeEvent(ctx context.Context, event *events.Eve
 		return
 	}
 
-	// ensure we do not send unintended events to Alby API
-	eventWhitelist := []string{
-		"nwc_backup_channels",
-		"nwc_payment_received",
-		"nwc_payment_sent",
-		"nwc_payment_failed",
-		"nwc_app_created",
-		"nwc_app_deleted",
-		"nwc_unlocked",
-		"nwc_node_sync_failed",
-		"nwc_outgoing_liquidity_required",
-		"nwc_incoming_liquidity_required",
-		"nwc_budget_warning",
-		"nwc_channel_ready",
-		"nwc_channel_closed",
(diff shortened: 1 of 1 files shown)


alby/alby_oauth_service.go
```diff
@@ -645,11 +646,38 @@ func (svc *albyOAuthService) ConsumeEvent(ctx context.Context, event *events.Eve
...
+
+	if !slices.Contains(eventWhitelist, event.Event) {
+		logger.Logger.WithField("event", event).Debug("Skipped sending non-whitelisted event to alby events API")
+		return
+	}
 
 	// TODO: rename this config option to be specific to the alby API
 	if !svc.cfg.GetEnv().LogEvents {
-		logger.Logger.WithField("event", event).Debug("Skipped sending to alby events API")
+		logger.Logger.WithField("event", event).Debug("Skipped sending to alby events API (alby event logging disabled)")
```
bumi, line 680: nit: first check if it's enabled and if it's enabled we do the whitelist check?

alby/alby_oauth_service.go
```diff
@@ -645,11 +646,38 @@ func (svc *albyOAuthService) ConsumeEvent(ctx context.Context, event *events.Eve
 		return
 	}
 
-	// TODO: we should have a whitelist rather than a blacklist, so new events are not automatically sent
+	// ensure we do not send unintended events to Alby API
+	eventWhitelist := []string{
```
bumi, line 650: can we move this outside of the function into some constant?

alby/alby_oauth_service.go
rolznz: It's just a bit harder to test without enabling events in dev mode. But I will switch it

alby/alby_oauth_service.go
rolznz: I moved it to a separate function (unfortunately in go there is no constant for arrays)

im-adithya (approved):
utACK

PR #907, opened 2024-12-18
feat: nwc create_connection command (WIP)

rolznz:
This is a bit of an exploration on how new app connections could be created via nwc (e.g. for NWA). e.g. in Alby Go.

Things I noticed:
- initial implementation is quite minor, most of the code is already written.
- [x] use methods instead of scopes since scopes do not exist in the NIP-47 spec.
- [x] UI change required to have the new superuser scope
- [ ] budgets and other restrictions do not make sense if this scope exists, since the superuser scope can create new connections without restrictions
- [ ] Alby Go would need a custom detail page to enable the superuser scope
- [ ] Does there need to be a way in the hub to explicitly confirm a superuser scope to create an app (e.g. re-entering unlock password?)
- [x] add missing tests
- [x] implement in JS SDK
- [ ] implement "NWA" in bitcoin connect
- [ ] implement "NWA" in Alby Go

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -883,6 +883,8 @@ func (api *api) GetWalletCapabilities(ctx context.Context) (*WalletCapabilitiesR
 	if len(notificationTypes) > 0 {
 		scopes = append(scopes, constants.NOTIFICATIONS_SCOPE)
 	}
+	// add always-supported capabilities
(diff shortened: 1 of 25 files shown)


rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/1016

Issue #818, opened 2024-11-21
Reset router does not work with VSS enabled


rolznz:
We directly try to write to the sqlite DB rather than writing to the ldk-node KV store which does nothing if we don't even use sqlite

rolznz:
@rdmitr any chance you can implement the alby hub side of this too?

PR #912, opened 2024-12-20
feat: implement router reset through the LDK-node Builder

rdmitr:
Fixes https://github.com/getAlby/hub/issues/818


rolznz (approved):
tACK (VSS and non-VSS)

PR #922, opened 2024-12-24
feat: PostgreSQL support

rdmitr:
TODOs:
- [x] fix timestamps
- [x] github workflows should run tests on both DB types
- [ ] how do existing cloud users migrate from sqlite to postgres? -> run migration from getalby.com as part of update process? - _move to seperate issue_
- [ ] what changes do we need to make in the backup & restore process?
  - [ ] cloud -> self-hosted -> _move to separate issue_
  - [ ] self-hosted -> cloud -> _move to separate issue_
- [x] disable migrations for now if postgres is enabled
- [x] Roland's testing
  - [x] brand new postgres DB
  - [x] brand new sqlite DB
  - [x] existing sqlite DB
  - [x] migrate existing sqlite DB (minor testing)
  - [x] migrate postgres DB to sqlite


db/db.go
```diff
@@ -2,25 +2,81 @@ package db
...
 func NewDB(uri string, logDBQueries bool) (*gorm.DB, error) {
+	return NewDBWithConfig(&Config{
+		URI:        uri,
+		LogQueries: logDBQueries,
+		DriverName: "",
+	})
+}
 
-	config := &gorm.Config{
+func NewDBWithConfig(cfg *Config) (*gorm.DB, error) {
```
rolznz, line 32: shouldn't this function always be used? how does the migration code work by using `NewDB()`? the URI cannot be specified?

rolznz:
@rdmitr do we need to update our github workflows to also test in Postgres mode? ideally we test in both modes

db/db.go
rolznz: oops, sorry - I misread this

db/db.go
```diff
@@ -2,25 +2,81 @@ package db
...
 	}
-	if logDBQueries {
-		config.Logger = gorm_logger.Default.LogMode(gorm_logger.Info)
+	if cfg.LogQueries {
+		gormConfig.Logger = gorm_logger.Default.LogMode(gorm_logger.Info)
+	}
+
+	var ret *gorm.DB
+
+	if IsPostgresURI(cfg.URI) {
```
rolznz, line 42: Leaving a comment to look into the timestamp issue. I am not sure what the problem is yet.

db/migrations/202405302121_store_decrypted_request.go
```diff
@@ -16,7 +16,7 @@ var _202405302121_store_decrypted_request = &gormigrate.Migration{
 			return err
 		}
 
-		err = tx.Exec("CREATE INDEX `idx_request_events_method` ON `request_events`(`method`);").Error
+		err = tx.Exec("CREATE INDEX \"idx_request_events_method\" ON \"request_events\"(\"method\");").Error
```
rolznz, line 19: @rdmitr would it be better to be consistent in the format here?

The migration above uses **`** and **'**

This migration escapes using **\**

db/migrations/202406061259_delete_content.go
```diff
@@ -19,8 +20,11 @@ var _202406061259_delete_content = &gormigrate.Migration{
 			return err
 		}
 
-		if err := tx.Exec("VACUUM").Error; err != nil {
-			return err
+		// Cannot run when testing with txdb: VACUUM must be run outside of transaction.
```
rolznz, line 23: I think we probably don't need this migration any more. Could we just comment out the content?

db/migrations/202406071726_vacuum.go
```diff
@@ -15,8 +16,11 @@ import (
 var _202406071726_vacuum = &gormigrate.Migration{
 	ID: "202406071726_vacuum",
 	Migrate: func(tx *gorm.DB) error {
-		if err := tx.Exec("VACUUM").Error; err != nil {
-			return err
+		// Cannot run when testing with txdb: VACUUM must be run outside of transaction.
+		if !testing.Testing() {
```
rolznz, line 20: I think we probably don't need this migration any more. Could we just comment out the content?

db/migrations/20241221234500_fix_types.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+var _20241221234500_fix_types = &gormigrate.Migration{
+	ID: "20241221234500_fix_types",
+	Migrate: func(tx *gorm.DB) error {
+		if tx.Dialector.Name() == "postgres" {
```
rolznz, line 11: Needs review

db/migrations/20241221234500_fix_types.go
```diff
@@ -0,0 +1,27 @@
...
+			err := tx.Exec(`
+CREATE SEQUENCE user_configs_id_seq;
```
rolznz, line 13: add comment: `// due to not explicitly setting the user_configs table to use auto increment in previous migrations`

db/migrations/20241221234500_fix_types.go
```diff
@@ -0,0 +1,27 @@
...
+ALTER TABLE user_configs ALTER COLUMN id SET DEFAULT nextval('user_configs_id_seq');
+ALTER SEQUENCE user_configs_id_seq OWNED BY user_configs.id;
+ALTER TABLE user_configs ALTER COLUMN encrypted SET DATA TYPE BOOLEAN USING encrypted::integer::boolean;
```
rolznz, line 16: can this be undone? what is the type without this?

db/db.go
```diff
@@ -91,3 +154,19 @@ func Stop(db *gorm.DB) error {
...
 	return nil
 }
+
+func IsPostgresURI(uri string) bool {
+	return strings.HasPrefix(uri, "postgresql://")
+}
+
+var txSerializer = make(chan struct{}, 1)
+
+func RunTransaction(db *gorm.DB, txFunc func(tx *gorm.DB) error) error {
```
rolznz, line 164: I saw only some instances of `db.Transaction` have been updated to use this. Is that intentional?

db/migrations/20241221234500_fix_types.go
rolznz: edit: resolved

db/db.go
rolznz: This feels a bit hacky if it is done just for tests, is there a way we can isolate the changes to the test DB instead?

cmd/migrate/main.go
```diff
@@ -0,0 +1,124 @@
+package main
```
rolznz, line 1: just thinking of the filename `migrate` we have different types of migrations. Could we be more clear e.g. `db_migrate` or something?

cmd/migrate/main.go
```diff
@@ -0,0 +1,124 @@
...
+
+func migrateDB(from, to *gorm.DB) error {
+	tx := to.Begin()
+	defer tx.Rollback()
+
+	if err := tx.Error; err != nil {
+		return fmt.Errorf("failed to start transaction: %w", err)
+	}
+
+	// Table migration order matters: referenced tables must be migrated
```
rolznz, line 68: can we fail the migration if the number of tables in the DB are not expected (e.g. we made a new table and forgot to update this migrate command?)

cmd/migrate/main.go
```diff
@@ -0,0 +1,124 @@
...
+	var data []T
+	if err := from.Find(&data).Error; err != nil {
+		return fmt.Errorf("failed to fetch data: %w", err)
+	}
+
+	if len(data) == 0 {
+		return nil
+	}
+
+	if err := to.Create(data).Error; err != nil {
```
rolznz, line 119: I got an error when I tried to create a new app after migrating from sqlite to postgres. It says duplicate ID not allowed.

My current apps:

 id | name | description |                            app_pubkey                            |          
created_at          |          updated_at           | isolated | metadata |                          wa
llet_pubkey                           
----+------+-------------+------------------------------------------------------------------+------------------------------+-------------------------------+----------+----------+------------------------------------------------------------------
  1 | test |             | a0aa432130518b7bf921e6d55cce725f172693196fa1b8b8b103f2bcd2225c87 | 2025-01-10 15:29:08.26065+07 | 2025-01-10 15:29:08.268216+07 | f        | {}       | 19c16a92888d3ebbb594fa7fa814886b9bfd014395ca9e09f98165010ec01dd9
(1 row)

db/migrations/202406071726_vacuum.go
rolznz: maybe commenting out the code inside the migration is fine for now

db/db.go
rolznz: I think this is ok. We can run the sqlite tests locally by default and check all of them in the GH action

cmd/migrate/main.go
rolznz: Thanks, that worked!

rolznz (approved):
tACK

Issue #675, opened 2024-09-19
Implement improved flow of opening channels


stackingsaunter:
Flow was tested and improved after usability test among 4 users, two rather beginners with low technical/node/channels knowledge and 2 with a bit more. 

[Figma and mobile design](https://www.figma.com/design/2IyilDewDrdNP3G1NixPYd/%F0%9F%8C%93-Alby-Hub?node-id=5347-99256&t=0u6SyGHmjhrEkyT8-4)

### Intermediary page
All users from usability test claimed they initiated channel opening through main "Open Channel" CTA. This page is what appears when user clicks it, and describes two types of channels

![image](https://github.com/user-attachments/assets/a3aa18a1-e41e-4641-9e40-cd63477d2efb)

### Payment methods page
When chosen to open a channel with only spending balance, user views a choice of payment methods that lead to 3 different flows. This also shows when user clicks "Top Up" on Node page

![image](https://github.com/user-attachments/assets/df57bae3-1c42-4fb4-8851-ebd7493154ee)

### Open Channel with On-chain
![image](https://github.com/user-attachments/assets/62c65f0c-5ddb-42bf-9dce-1a5d5f0ffcba)

![image](https://github.com/user-attachments/assets/6f3ad295-6de7-4300-ae97-743bea8ae48b)

![image](https://github.com/user-attachments/assets/4a778e2d-f2ed-4adc-9ec9-7ab6f570e57a)

![image](https://github.com/user-attachments/assets/17b364af-adce-45f5-8ffd-8d37422299d8)

![image](https://github.com/user-attachments/assets/6bb5212c-4736-4a9e-9b5b-0f5b68ec29ce)

![image](https://github.com/user-attachments/assets/f88f84eb-fbff-4ed9-8cf3-04cf3bfbaae8)

---

### Open Channel with Lightning
This page also appears when users clicks "Increase" on Node page

![image](https://github.com/user-attachments/assets/196cd1cc-129b-4a08-a240-2f0416f8f79d)

![image](https://github.com/user-attachments/assets/57917218-368c-443f-8fac-e44c4762c4e2)

![image](https://github.com/user-attachments/assets/c1c368f7-f9db-4bd5-8d2a-320899a0c77c)

![image](https://github.com/user-attachments/assets/5b5709bf-8398-4452-aecc-09a4ef1a8977)

![image](https://github.com/user-attachments/assets/cef71174-eec7-44a7-aa6d-c861f80032e4)

[...]

PR #925, opened 2024-12-27
feat: node page revamp for web and mobile

pavanjoshi914:
node page revamp for desktop and mobile
add backend endpoint to fetch bitcoin rates
show usd equivalent of balances

image
image

Part of #675


frontend/src/screens/channels/Channels.tsx
```diff
@@ -69,6 +70,7 @@ export default function Channels() {
   const { data: channels, mutate: reloadChannels } = useChannels();
   const { data: nodeConnectionInfo } = useNodeConnectionInfo();
   const { data: balances, mutate: reloadBalances } = useBalances();
+  const { data: bitcoinRate } = useBitcoinRate("usd");
```
rolznz, line 73: Does it make sense to pass the currency from the frontend? should it be something a user can change in the settings page?

http/alby_http_service.go
```diff
@@ -29,6 +29,7 @@ func NewAlbyHttpService(svc service.Service, albyOAuthSvc alby.AlbyOAuthService,
 func (albyHttpSvc *AlbyHttpService) RegisterSharedRoutes(restrictedGroup *echo.Group, e *echo.Echo) {
 	e.GET("/api/alby/callback", albyHttpSvc.albyCallbackHandler)
 	e.GET("/api/alby/info", albyHttpSvc.albyInfoHandler)
+	e.GET("/api/alby/btc_rate", albyHttpSvc.albyBitcoinRateHandler)
```
rolznz, line 32: can you please try to be consistent with the other routes? why `btc_rate`?

alby/alby_oauth_service.go
```diff
@@ -1051,6 +1051,34 @@ func (svc *albyOAuthService) GetChannelPeerSuggestions(ctx context.Context) ([]C
...
+	}
+	setDefaultRequestHeaders(req)
+
+	res, err := client.Do(req)
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to fetch Bitcoin rate from API")
+		return nil, err
+	}
+
+	var rate = &BitcoinRate{}
```
rolznz, line 1072: can you also check the http response is OK here? check other methods. If you copied this from somewhere can you please also update it there too.

frontend/src/screens/channels/Channels.tsx
```diff
@@ -328,6 +330,17 @@ export default function Channels() {
                     sats
                   </div>
                 )}
+                {balances && bitcoinRate && (
```
rolznz, line 333: should we create a component for displaying a sat + fiat value? I see this is duplicated, and then we could also move the hook to fetch the rates to this component

frontend/src/screens/channels/Channels.tsx
pavanjoshi914: hmm. nice idea. we can fetch supported currencies and show it in settings and use it to show balances + fait equivalent of transactions as well.  maybe we do it in separate pr?

http/alby_http_service.go
pavanjoshi914: i changed endpoint name to rates

frontend/src/screens/channels/Channels.tsx
rolznz: I think so, but in that case can you remove the query parameter passing that is in this PR?

frontend/src/screens/channels/Channels.tsx
pavanjoshi914: resolved

frontend/src/components/channels/ChannelsCards.tsx
```diff
@@ -90,25 +83,10 @@ export function ChannelsCards({ channels, nodes }: ChannelsCardsProps) {
...
-                            <TooltipTrigger>
-                              <div className="flex flex-row gap-2 items-center">
-                                <p className="text-muted-foreground font-medium">
-                                  Reserve
-                                </p>
-                                <InfoIcon className="h-4 w-4 shrink-0" />
-                              </div>
-                            </TooltipTrigger>
-                            <TooltipContent className="w-[400px]">
-                              Funds each participant sets aside to discourage
```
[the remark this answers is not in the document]
rolznz: @stackingsaunter did you intend to do this? we want to provide context about what things mean

http/alby_http_service.go
```diff
@@ -85,6 +86,18 @@ func (albyHttpSvc *AlbyHttpService) albyInfoHandler(c echo.Context) error {
 	return c.JSON(http.StatusOK, info)
 }
 
+func (albyHttpSvc *AlbyHttpService) albyBitcoinRateHandler(c echo.Context) error {
+	currency := c.QueryParam("currency")
```
rolznz, line 90: can you please remove the currency query param

wails/wails_handlers.go
```diff
@@ -92,6 +92,28 @@ func (app *WailsApp) WailsRequestRouter(route string, method string, body string
 		}
 	}
 
+	btcRateRegex := regexp.MustCompile(
```
rolznz, line 95: all this code can be removed, we do not need to pass the currency from the frontend

alby/alby_oauth_service.go
```diff
@@ -1051,6 +1051,42 @@ func (svc *albyOAuthService) GetChannelPeerSuggestions(ctx context.Context) ([]C
...
+
+	res, err := client.Do(req)
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to fetch Bitcoin rate from API")
+		return nil, err
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
```
rolznz, line 1074: can you please check the others API calls to see what they do? I think this is a fragile check. It could be another 2XX or 3XX code (e.g. if the result was not modified)

frontend/src/components/channels/ChannelsCards.tsx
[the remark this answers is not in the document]
rolznz: Confirmed with saunter that he removed them initially because tooltips didn't work on mobile. This has been fixed.

@pavanjoshi914 please add the same tooltips as on the channels table.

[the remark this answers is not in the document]
pavanjoshi914: added tooltips for mobile view as well

rolznz (approved):
tACK

PR #930, opened 2024-12-30
feat: add boltz.exchange swap out option

reneaaron:
PoC

### DIFF
--- a/frontend/src/components/channels/ChannelDropdownMenu.tsx
+++ b/frontend/src/components/channels/ChannelDropdownMenu.tsx
@@ -1,4 +1,5 @@
 import {
+  ArrowRightLeft,
   ExternalLinkIcon,
   HandCoins,
   MoreHorizontal,
@@ -20,6 +21,7 @@ import {
   DropdownMenuTrigger,
 } from "src/components/ui/dropdown-menu.tsx";
 import { Channel } from "src/types";
+import { openLink } from "src/utils/openLink";
 
 type ChannelDropdownMenuProps = {
   alias: string;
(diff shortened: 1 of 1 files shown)


frontend/src/components/channels/ChannelDropdownMenu.tsx
```diff
@@ -76,6 +78,23 @@ export function ChannelDropdownMenu({
               </DropdownMenuItem>
             </AlertDialogTrigger>
           )}
+          {channel.localSpendableBalance > 0 && (
+            <DropdownMenuItem
+              className="flex flex-row items-center gap-2 cursor-pointer"
+              onClick={() => {
+                const amount = channel.localSpendableBalance / 1000;
```
rolznz, line 85: We currently cannot choose what channel to pay from, so I think this option needs to go in the advanced dropdown for the node page itself, rather than per channel.

We can probably not spend the whole amount. Would it make sense to allow the user to input an amount, and default to ~90% of the value of the largest channel? that way if it fails, users can still try again with a different amount.

frontend/src/components/channels/ChannelDropdownMenu.tsx
```diff
@@ -76,6 +78,23 @@ export function ChannelDropdownMenu({
...
+                // TODO: Fetch new onchain address
```
rolznz, line 86: Is it ok to always use hub addresses? or the user can choose between a hub one and an external one (or they can pay to an external one after, but this requires an additional transaction)

reneaaron: I would suggest implementing the simplest scenario and iterating afterwards based on user feedback. 

If you want to do advanced scenarios you can always resort to use a swap provider manually for now. (e.g. to withdraw to cold storage)

frontend/src/components/channels/ChannelDropdownMenu.tsx
reneaaron: > We currently cannot choose what channel to pay from, so I think this option needs to go in the advanced dropdown for the node page itself, rather than per channel.

> We can probably not spend the whole amount. Would it make sense to allow the user to input an amount, and default to ~90% of the value of the largest channel? that way if it fails, users can still try again with a different amount.

Oh, okay. Thanks for this feedback, much appreciated :muscle:

frontend/src/components/channels/ChannelDropdownMenu.tsx
rolznz: Sounds good!

frontend/src/screens/channels/Channels.tsx
```diff
@@ -114,86 +150,152 @@ export default function Channels() {
...
+                    </p>
+                  </div>
+                </div>
+                <DialogFooter>
+                  <Button
+                    type="submit"
+                    onClick={async () => {
+                      await getNewAddress();
+                      openLink(
+                        `https://boltz.exchange/?sendAsset=LN&receiveAsset=BTC&sendAmount=${swapOutAmount}&destination=${onchainAddress}`
```
rolznz, line 290:
```suggestion
                        `https://boltz.exchange/?sendAsset=LN&receiveAsset=BTC&sendAmount=${swapOutAmount}&destination=${onchainAddress}&ref=alby`
```

rolznz (approved):
tACK

Issue #911, opened 2024-12-20
Include counterparty node IDs in pending channel closures message on node page


rolznz:
Currently the user has to go to Settings -> Debug Tools -> Get Balances to find this, amongst a lot of JSON (less technical users will definitely have issues).

It would be even better if we had the TXID, but that would require saving it e.g. https://github.com/getAlby/hub/issues/848 or LDK making it always available.

PR #932, opened 2024-12-30
feat: balance details for pending channel closures

rdmitr:
Fixes https://github.com/getAlby/hub/issues/911


rolznz (approved):
tACK

Node names are shown as unknown since I tested on regtest/mutinynet

@im-adithya  I couldn't find a way to get the LND channel ID but we currently do not use it.

image

image

Issue #402, opened 2024-08-05
Plugins (Ideas tracking issue)


rolznz:
What are we trying to solve?
- Easily run custom apps without requiring a separate web server?
- ?

What would the plugins be able to access / modify in Alby Hub?
- a NWC connection?
- custom frontend pages?

Plugin ideas:
- Uncle Jim (generate subaccounts)
- ZapSplitter (receive a payment and forward it to N addresses)
- ?

rolznz:
Automation: https://feedback.getalby.com/-alby-hub-request-a-feature/posts/bitcoin-automation

PR #933, opened 2025-01-02
feat: ZapPlanner custom app

rolznz:
TODOs:
- [x] add card on home page if the user has no zapplanner apps
- [x] ability to provide an optional name and message through dialog
- [x] create subscription via api and save subscription ID to the app metadata
- [x] change to sat pricing, not fiat
- [ ] review method of app detail page interacting with zapplanner (should this be proxied through alby hub rather than adding a CSRF exception?)

image


frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -0,0 +1,183 @@
...
+      if (apps?.some((existingApp) => existingApp.name === name)) {
+        throw new Error("A connection with the same name already exists.");
+      }
+
+      const amountSats = 5000;
+
+      const createAppRequest: CreateAppRequest = {
+        name: `ZapPlanner - ${name}`,
+        scopes: [
+          "get_balance",
```
rolznz, line 66: remove unnecessary scopes

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -0,0 +1,183 @@
...
+      };
+
+      const createAppResponse = await createApp(createAppRequest);
+      /*toast({
+        title: "Created subscription",
+        description: "The first payment is scheduled for 1st of October.",
+      });*/
+      const comment = encodeURIComponent("ZapPlanner payment from Alby Hub"); // TODO: allow customization
+      const payerData = ""; //encodeURIComponent(JSON.stringify({}));
+      // TODO: consider fiat conversion
```
rolznz, line 90: remove this TODO - we also need fiat budgets which are unlikely to be added soon

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -0,0 +1,183 @@
...
+    (app) => app.metadata?.app_store_app_id === "zapplanner"
+  );
+
+  return (
+    <div className="grid gap-5">
+      <AppHeader
+        title="ZapPlanner"
+        description="Schedule automatic recurring lightning payments"
+      />
+      <h2 className="font-semibold text-xl">Inspiration</h2>
```
rolznz, line 114: remove this header and content

bumi:
that's super cool!

can we add a quick note somewhere explaining on how it works? 

"ZapPlanner is a tool to securely schedule recurring payments. A new special app connection with a strict budget is created for each scheduled payment. This allows you to securely setup recurring payments and be in full control"

rolznz:
@bumi @reneaaron I will merge it as is for now.

I am not really happy with the CSRF exception here, but I discussed with Rene and this was a way to do a quick proof of concept. The other option is to proxy the API calls through the backend, but this requires some thought architecturally as it means an app detail page also requires adding routes, and then is much more complicated and possibly more in the direction of proper [plugins](https://github.com/getAlby/hub/issues/402). 

image

PR #944, opened 2025-01-06
feat: enable multi-path payments in LND

im-adithya:
:warning: This requires #937 for `CalculateFeeReserveMsat` method

## Description

This adds SendPaymentaV2 method to make use of the MaxParts parameter to make multi path payments in LND backend


lnclient/lnd/lnd.go
```diff
@@ -313,34 +313,54 @@ func (svc *LNDService) LookupInvoice(ctx context.Context, paymentHash string) (t
...
+
 func (svc *LNDService) SendPaymentSync(ctx context.Context, payReq string, amount *uint64) (*lnclient.PayInvoiceResponse, error) {
-	sendRequest := &lnrpc.SendRequest{PaymentRequest: payReq}
+	const MAX_PARTIAL_PAYMENTS = 21
+	const SEND_PAYMENT_TIMEOUT = 60
+	sendRequest := &routerrpc.SendPaymentRequest{
+		PaymentRequest: payReq,
+		MaxParts:       MAX_PARTIAL_PAYMENTS,
+		TimeoutSeconds: SEND_PAYMENT_TIMEOUT,
+		FeeLimitSat:    math.MaxInt64,
```
im-adithya, line 336: This is set to max currently because leaving this would assume that we want 0-fee routes and would result in no route errors. This should be changed with the value from `CalculateFeeReserveMsat`

lnclient/lnd/lnd.go
```diff
@@ -372,15 +392,16 @@ func (svc *LNDService) SendKeysend(ctx context.Context, amount uint64, destinati
...
 	const KEYSEND_CUSTOM_RECORD = 5482373484
 	destCustomRecords[KEYSEND_CUSTOM_RECORD] = preImageBytes
-	sendPaymentRequest := &lnrpc.SendRequest{
+	sendPaymentRequest := &routerrpc.SendPaymentRequest{
 		Dest:              destBytes,
 		AmtMsat:           int64(amount),
 		PaymentHash:       paymentHashBytes,
 		DestFeatures:      []lnrpc.FeatureBit{lnrpc.FeatureBit_TLV_ONION_REQ},
 		DestCustomRecords: destCustomRecords,
+		FeeLimitSat:       math.MaxInt64,
```
im-adithya, line 401: Same as above

lnclient/lnd/lnd.go
```diff
@@ -313,34 +313,54 @@ func (svc *LNDService) LookupInvoice(ctx context.Context, paymentHash string) (t
...
+
+		if payment.Status != lnrpc.Payment_IN_FLIGHT {
+			return payment, nil
+		}
+	}
+}
+
 func (svc *LNDService) SendPaymentSync(ctx context.Context, payReq string, amount *uint64) (*lnclient.PayInvoiceResponse, error) {
-	sendRequest := &lnrpc.SendRequest{PaymentRequest: payReq}
+	const MAX_PARTIAL_PAYMENTS = 21
```
rolznz, line 330: is there a good default provided by LND? this seems way too high to me

lnclient/lnd/lnd.go
rolznz: hmm, I thought LND set a reasonable default, but anyway good to point this out, please make sure to update both places as soon as we merge the other PR

lnclient/lnd/lnd.go
```diff
@@ -313,34 +313,54 @@ func (svc *LNDService) LookupInvoice(ctx context.Context, paymentHash string) (t
...
+	const SEND_PAYMENT_TIMEOUT = 60
```
rolznz, line 331: note: LDK is set to 50 seconds currently, since the JS SDK at least uses a 60 second timeout for NWC requests (although all this should be re-reviewed)

im-adithya: I took 60 from here: https://github.com/getAlby/hub/blob/master/lnclient/ldk/ldk.go#L465, should this be changed?

lnclient/lnd/lnd.go
im-adithya: Yeah there should've been a default for this :/

> please make sure to update both places as soon as we merge the other PR

Will do 🫡

lnclient/lnd/lnd.go
im-adithya: Sadly no, a good reference can be Polar, where it is set to 16 (I think setting it to a large number would cause more number of in flight HTLCs which can be bad)

https://github.com/jamaljsr/polar/blob/ba6b1cf082732fa8aa12bb6db46c2c57b1a7a089/src/lib/lightning/lnd/lndService.ts#L184

rolznz: I doubt a MPP payment with 16 parts would succeed. The more parts and more hops the more chance of failures. I asked Zeus

lnclient/lnd/lnd.go
rolznz: True, we should probably set that to 50 seconds too. Could you update it?

lnclient/lnd/lnd.go
im-adithya: Atleast in LND, the way it works is it tries to make it in full, if not then it keeps splitting till the min amount depending on the max parts, so we should be good. From https://lightning.engineering/posts/2020-05-07-mpp/

> Initially, LND will try to complete the payment in one part. If that’s not possible, say because channel capacities aren’t sufficient or because other constraints like the fee limit are hit, an attempt is made for half the payment amount. After that HTLC has been launched, the software searches for a second path to pay the remaining amount (the other half). If further failures happen, amounts are repeatedly split until a minimum amount limit is reached or the payment succeeds.

lnclient/lnd/lnd.go
im-adithya: Done!

lnclient/lnd/lnd.go
im-adithya: I have a feeling (from the blog) that this should be good, reduced to 16 now, let me know what you think

rolznz: Thanks for checking!

rolznz: Evan from Zeus also recommended 16, cool.

rolznz (approved):
utACK

Issue #720, opened 2024-10-03
Improve Auto unlock feature


rolznz:
There is an `AUTO_UNLOCK_PASSWORD` field which can be provided to automatically start the node on Alby Hub startup. However, this requires Umbrel and Start9 to allow the user to edit this env variable and is not enabled by default, which should be.

How can we improve this for self-hosted hubs?

rolznz:
Fundamentally, maybe Start9 and Umbrel should not even have unlock passwords. This needs more planning.

I still think this is an issue though e.g. with powercuts or restarts, it's harder for users to get Alby Hub back up and running.

jpl-btc:
After months of using Alby Hub on Start9 and Umbrel, I’ve rarely needed to enter the unlock password. Once the app is opened, the OS keeps it active, so re-entering the password is only required after a system reboot when I did restart Umbrel, logging in felt natural since all other apps also required re-authentication.

In summary, the unlock password hasn’t been a hassle, it’s only prompted after a reboot, which aligns with expected security behavior. This is based solely on my own experience, and I have not heard complains about this from umbrel or Start9 users so far

rolznz:
I personally have issues with this due to semi-frequent powercuts. But this can be fixed with a backup power supply.

riccardobl:
Hi, I run Alby Hub with Docker on a home server that occasionally needs to reboot. I used this env variable to make Alby Hub come back online automatically, just like the other services, so I don’t have to worry about unlocking it manually. 

In my specific setup this is a spending node with few sats in it, i don't need to protect it from being stolen or things like that, so this security feature (that is very important otherwise) is not needed for me. 

I think this setup might be fairly common. I wouldn’t recommend enabling auto-unlock as the default, but it may be worth mentioning this thing and the related env variable in the Docker setup guide. It’s a bit unexpected behavior for a Docker container, and I only discovered it after receiving Alby’s email about unlocking the hub (by the way, thanks for doing these extra checks on our nodes :+1: )

PR #949, opened 2025-01-07
feat: auto unlock

rolznz:
Closes https://github.com/getAlby/hub/issues/720

Users can optionally save their unlock password to their user_config as plaintext to enable auto start of the node after a machine restart or alby hub update. The unlock password is still required to access the UI.

if the AUTO_UNLOCK_PASSWORD environment variable is set it will overwrite the entry in user_configs.

This feature is only enabled for non-Alby Cloud environments by checking if the default OAuth client is used. Because we do not control other hosting environments (e.g. nodana, and others) there is an additional warning before enabling the feature.

This feature **must be disabled** before backing up the node for migration to ensure that the unlock password is not unintentionally moved to another machine. It also must be disabled before changing the unlock password just for simplicity. (Checks are added to both functions)

image

image

### DIFF
--- a/frontend/src/screens/settings/AutoUnlock.tsx
+++ b/frontend/src/screens/settings/AutoUnlock.tsx
@@ -1,0 +1,129 @@
+import { AlertTriangle } from "lucide-react";
+import React from "react";
+
+import Container from "src/components/Container";
+import Loading from "src/components/Loading";
+import SettingsHeader from "src/components/SettingsHeader";
+import { Alert, AlertDescription, AlertTitle } from "src/components/ui/alert";
+import { Input } from "src/components/ui/input";
+import { Label } from "src/components/ui/label";
+import { LoadingButton } from "src/components/ui/loading-button";
+import { useToast } from "src/components/ui/use-toast";
+
+import { useInfo } from "src/hooks/useInfo";
+import { request } from "src/utils/request";
+
+export function AutoUnlock() {
+  const { toast } = useToast();
+  const { data: info, mutate: refetchInfo } = useInfo();
+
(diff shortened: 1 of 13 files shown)


bumi:
I am wondering if we somehow can unify the two ways we now have to set the auto unlock password. 
We have the env variable and we have the DB. I am wondering if we should handle the env var similar to the macaroon for example and save it to the DB and then always use the DB value.

bumi:
some more inspiration: https://docs.lightning.engineering/lightning-network-tools/lnd/wallet#more-secure-example-with-password-manager-and-using-a-named-pipe

frontend/src/screens/settings/AutoUnlock.tsx
```diff
@@ -0,0 +1,129 @@
...
+            </form>
+          </>
+        )}
+        {info.autoUnlockPasswordEnabled && (
+          <>
+            <Alert>
+              <AlertTriangle className="h-4 w-4" />
+              <AlertTitle>Security Warning</AlertTitle>
+              <AlertDescription>
+                Your unlock password is currently saved in plaintext on the
```
bumi, line 111: "currently" makes it sound like it's our fault that this is insecure

I would say something like: 

In some situations it can be impractical to manually unlock the wallet every time Alby Hub is started. In those cases you can save the unlock password in plain text to tell the wallet to auto-unlock itself. 
Attention: everyone who has access to your machine could read that password and take your funds. Use this only in a secure environment.

rolznz: Updated and simplified to have this as a constant alert rather than two variants

Issue #890, opened 2024-12-16
Cashu: handle pending proofs


rolznz:
Currently it's possible if a payment got stuck to be unable to make any more payments.

1. Upgrade gonuts dependency: https://github.com/getAlby/hub/pull/776

2. Handle pending proofs: https://github.com/elnosh/gonuts/issues/88

3. More clearly mark that this LNClient option is experimental in the onboarding

4. Consider a way for users from <0.3 to recover their funds as recommended by elnosh. But it must be isolated to the cashu LNClient. Maybe there could be some check at startup that if it discovers there are pending proofs it can reset the wallet?

itstomekk:
There is one quite active Cashu discord server, great guys, I posted it there

callebtc:
I would like to help but I don't understand what the issue is. Is it something related to Cashu?

rolznz:
@callebtc it's not an issue with Cashu. Alby Hub currently does not handle pending proofs properly, leading to users unable to make payments once they experience a payment that gets stuck. The gonuts library has improved support for pending proofs. I've added the steps I think need to be done to this issue.

PR #962, opened 2025-01-08
feat: update gonuts dependency

elnosh:
Closes https://github.com/getAlby/hub/issues/890
Closes https://github.com/getAlby/hub/issues/651

I have made changes for new gonuts version, however not pointing to v0.3.0 but instead to latest commit in main for now since I have changed some things since v0.3.0.

This should handle pending proofs that are stuck in payments and hence not try to use them in other payments.

Another change for this new version is that it does not have an `Invoice` type anymore and instead `MintQuote` and `MeltQuote`s so it has to aggregate those 2 when listing transactions. 

Regarding point 4 in https://github.com/getAlby/hub/issues/890, I think doing a restore if it finds any pending proofs during startup could be undesirable since the restore does not restore previous transaction history. So any time there are pending proofs and the hub starts, it will wipe out the transaction history. Maybe a cashu-backend specific button could be added in Settings > debug tools where if the user experienced the issue of a unusable wallet due to stuck payments, they could trigger a manual restore? 

Regarding this issue https://github.com/getAlby/hub/issues/651, I know it's not ideal not having the preimage for incoming payments and one is expected but maybe a dummy/placeholder one could be used so that the hub can mark it as settled and show up in the transaction history and don't go into that loop where it would always check that invoice even though it already got paid.

Any thoughts or comments on the changes and suggestions made here are appreciated :)


rolznz:
Thanks for picking this up @elnosh !

> Regarding point 4 in https://github.com/getAlby/hub/issues/890, I think doing a restore if it finds any pending proofs during startup could be undesirable since the restore does not restore previous transaction history. So any time there are pending proofs and the hub starts, it will wipe out the transaction history. Maybe a cashu-backend specific button could be added in Settings > debug tools where if the user experienced the issue of a unusable wallet due to stuck payments, they could trigger a manual restore?

That sounds better, but I would prefer to not leak individual LNClient-specific code to other parts of the app. It might be best to have a generic "Execute Node Command" button in the debug tools where you can type some text, and then the Cashu implementation can parse it and do what it needs. We originally planned to do something like this but did not yet add it. I will see what we can do and get back to you. CC @rdmitr 

> Regarding this issue https://github.com/getAlby/hub/issues/651, I know it's not ideal not having the preimage for incoming payments and one is expected but maybe a dummy/placeholder one could be used so that the hub can mark it as settled and show up in the transaction history and don't go into that loop where it would always check that invoice even though it already got paid.

This sounds like a good workaround to me!

lnclient/cashu/cashu.go
```diff
@@ -82,30 +65,24 @@ func (cs *CashuService) SendPaymentSync(ctx context.Context, invoice string, amo
...
-	transaction, err := cs.cashuInvoiceToTransaction(cashuInvoice)
-	if err != nil {
-		logger.Logger.WithField("invoice", invoice).WithError(err).Error("Failed to convert invoice to transaction")
-		return nil, err
+	if meltResponse == nil || meltResponse.Preimage == "" {
+		return nil, errors.New("no preimage in melt response")
 	}
-
-	fee := uint64(transaction.FeesPaid)
+	totalPaid := balanceBeforeMelt - cs.wallet.GetBalance()
```
rolznz, line 84: minor: could this generate an incorrect fee if there were two simultaneous payments? (but I don't think we need to change anything here)

rolznz:
@elnosh for now temporarily we could put the reset logic in this function just to test it, and not block this PR:
```
func (cs *CashuService) ResetRouter(key string) error {
	return nil
}
```

It's not so good because it's unrelated to resetting the wallet, but it is executed from the dev tools page.

lnclient/cashu/cashu.go
elnosh: yeah thanks for catching that, if another transaction happens in between it could generate incorrect fee. A better way could be to use the `Change` from the melt response and then do `FeeReserve` - `ChangeAmount`.

elnosh:
made some  other changes:
- saving mnemonic from cashu wallet
- test cashu wallet restore from the ResetRouter in the debug tools
- added dummy preimage for mint quotes so that those get marked as settled.

rolznz:
I found an old cashu wallet for Alby Hub with mint https://8333.space:3338 which is operational (I can pay with my old balance) and I can send payments, but when I pay I get "no preimage in melt response". My balance then does not update after paying. Is this expected?

I managed to restore the proofs at https://wallet.cashu.me/ - nice! afterwards, I could still pay with my hub (I thought the tokens would have become invalid).

rolznz:
Starting a brand new hub with a new cashu wallet seems to work well.

Restoring my sats on the old wallet with the incorrect balance changed it from 33 to 11 sats.

With minibits I get the balance issue.
I deposited 10 sats, then paid 1 sat. Now my balance is 12. Maybe just a calculation bug?

lnclient/cashu/cashu.go
```diff
@@ -43,37 +45,31 @@ func NewCashuService(workDir string, mintUrl string) (result lnclient.LNClient,
...
-		if isFirstSetup {
-			// delete wallet to ensure user gets a fresh start when they try a different mint
-			// otherwise the wallet freezes on next startup
-			fileErr := os.RemoveAll(newpath)
-			if fileErr != nil {
-				logger.Logger.WithError(fileErr).Error("Failed to remove broken wallet directory")
-			}
+	cashuWalletMnemonic := cashuWallet.Mnemonic()
+	mnemonic, _ := cfg.Get("Mnemonic", encryptionKey)
+	if mnemonic != cashuWalletMnemonic {
```
rolznz, line 61: this will overwrite an existing mnemonic (there shouldn't be one). But should the check be `if mnemonic == ""` ?

rolznz: Actually, if the mnemonic is different and not an empty string, we should probably return an error rather than overwriting it.

frontend/src/screens/settings/DebugTools.tsx
```diff
@@ -304,7 +304,7 @@ export default function DebugTools() {
               Get Network Graph
             </Button>
           </AlertDialogTrigger>
-          {info?.backendType === "LDK" && (
+          {(info?.backendType === "LDK" || info?.backendType === "CASHU") && (
```
rolznz, line 307:
```suggestion
          {(info?.backendType === "LDK" || info?.backendType === "CASHU" /* TODO: move to generic node command */) && (
```

elnosh:
> https://8333.space:3338 which is operational (I can pay with my old balance) and I can send payments, but when I pay I get "no preimage in melt response". My balance then does not update after paying. Is this expected?

I've been testing with some mints (including that one) and I think that happens if you try to pay an invoice that was generated from that same mint. Looking at the response from the melt request, the payment is being settled and hence returning PAID but with an empty preimage and then making this check fail. 
```	
if meltResponse == nil || meltResponse.Preimage == "" {
	return nil, errors.New("no preimage in melt response")
}
```
I created issue for that here: https://github.com/cashubtc/nutshell/issues/625 
In other cases where I'm trying to pay a different lightning wallet or invoice from different mint, the preimage is there. Instead of checking the preimage we could check the melt state for "PAID" but not sure if you would want the empty preimage in those specific cases when it is returned empty.

elnosh:
> Restoring my sats on the old wallet with the incorrect balance changed it from 33 to 11 sats.
> 
11 should be the correct balance given some previous payments?

> With minibits I get the balance issue. I deposited 10 sats, then paid 1 sat. Now my balance is 12. Maybe just a calculation bug?
Sorry, I didn't understand this. Could you explain a bit further? You deposited 10 sats to an empty wallet set with minibits mint and got this?

lnclient/cashu/cashu.go
elnosh: I was first checking if it was empty but while testing it seems a different (from cashu wallet's) mnemonic was being created and set so that's why I put that check. I'm assuming the different one was set in first startup.

rolznz:
> Sorry, I didn't understand this. Could you explain a bit further? You deposited 10 sats to an empty wallet set with minibits mint and got this?

Yes, I deposited 10 sats into an empty wallet with minibits mint.

Then I did a 1 sat payment to rolznz@getalby.com. I guess the fee was 1 sat.

Instead of decreasing the balance to 8 sats (expected) it increased to 12 sats.

lnclient/cashu/cashu.go
rolznz: So what does this mean for existing hub users? their old mnemonic will be overwritten, right?

rolznz:
> 11 should be the correct balance given some previous payments?
I think so, yes. It just seems something is broken when we make payments, that the balance increases rather than decreases, or it does not omit already used cashu tokens

elnosh:
> Then I did a 1 sat payment to [rolznz@getalby.com](mailto:rolznz@getalby.com). I guess the fee was 1 sat.
> Instead of decreasing the balance to 8 sats (expected) it increased to 12 sats.


> It seems more than this - because in all my test payments I paid to my LDK hub node. I did not pay to the same mint.

I haven't been able to replicate those issues :( 

I've setup one with minibits mint and have been making payments fine (with preimage in the response). Sent to this address [rolznz@getalby.com](mailto:rolznz@getalby.com) and some other nodes and mints and balance has been updating correctly.

lnclient/cashu/cashu.go
elnosh: looking at this: https://github.com/getAlby/hub/blob/master/service/keys/keys.go#L63-L75 it seems one is created anyways even if the mnemonic is set to false for any specific backend. So mnemonic would never be empty. Maybe this mnemonic should not be overwritten (?). I'm not sure what it's used for. Should I put the cashu wallet mnemonic in some other key?

rolznz: Damn, I see. Ideally we use the one that Alby Hub generates as that's the one we ask the user to back up as part of the getting started steps, and then we pass that mnemonic to gonuts. Is this possible?

If gonuts always overwrites the mnemonic it is inconsistent with the rest of our LNClients and will cause other issues later on.

rolznz:
image

I cannot figure out why mine is like this, I sent you my cashu wallet in case that helps you

lnclient/cashu/cashu.go
elnosh: I noticed that "Backup" section in settings does not currently show up for cashu backend since mnemonic is set to false. 

> If gonuts always overwrites the mnemonic it is inconsistent with the rest of our LNClients and will cause other issues later on.

Ok, sure. 

Instead, maybe it can just be another command with that "Node Command" API you guys are planning? Send command to cashu wallet and that would return the mnemonic.

rolznz: @elnosh I guess that's acceptable. Then we also need to undo the change to display that the cashu connector. I will add a comment. Thanks!

The commands PR is in progress and should be hopefully ready early next week: https://github.com/getAlby/hub/pull/1007

frontend/src/lib/backendType.ts
```diff
@@ -33,7 +33,7 @@ export const backendTypeConfigs: Record<BackendType, BackendTypeConfig> = {
     hasNodeBackup: false,
   },
   CASHU: {
-    hasMnemonic: false,
+    hasMnemonic: true,
```
rolznz, line 36: let's undo this one since the mnemonic is incompatible

frontend/src/screens/settings/DebugTools.tsx
rolznz: We probably can wait for the generic node command feature to be merged and then undo this actually, then merge this PR

lnclient/cashu/cashu.go
```diff
@@ -9,19 +9,24 @@ import (
...
 )
 
 type CashuService struct {
-	wallet *wallet.Wallet
+	wallet  *wallet.Wallet
+	workDir string
 }
 
-func NewCashuService(workDir string, mintUrl string) (result lnclient.LNClient, err error) {
+func NewCashuService(cfg config.Config, workDir, encryptionKey, mintUrl string) (result lnclient.LNClient, err error) {
```
rolznz, line 29: undo these changes as we will now return the mnemonic from a node command

lnclient/cashu/cashu.go
rolznz: To move `cashuWallet.Mnemonic()` to a node command

lnclient/cashu/cashu.go
```diff
@@ -231,6 +220,25 @@ func (cs *CashuService) RedeemOnchainFunds(ctx context.Context, toAddress string
 }
 
 func (cs *CashuService) ResetRouter(key string) error {
+	mnemonic := cs.wallet.Mnemonic()
```
rolznz, line 223: for now we could check if `key == "ALL"` do the restore, otherwise get the mnemonic: `cashuWallet.Mnemonic()`

rolznz:
Added some improvements here: https://github.com/elnosh/hub/pull/1

rolznz:
Just waiting on https://github.com/getAlby/hub/pull/1007 to be merged.

rolznz (approved):
tACK

PR #968, opened 2025-01-09
fix: nil dereference in GetNodeStatus of cashu backend

elnosh:
app crashes if tries `Get Node Status` for cashu backend. 

Currently `GetNodeStatus` returns nil and then it tries to dereference that here https://github.com/getAlby/hub/blob/master/wails/wails_handlers.go#L573 causing it to crash. 

Returning an empty object rather than nil.

### DIFF
--- a/wails/wails_handlers.go
+++ b/wails/wails_handlers.go
@@ -582,6 +582,9 @@ func (app *WailsApp) WailsRequestRouter(route string, method string, body string
 		if err != nil {
 			return WailsRequestRouterResponse{Body: nil, Error: err.Error()}
 		}
+		if nodeStatus == nil {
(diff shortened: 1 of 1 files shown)


rolznz (approved):
utACK

rolznz:
Edit: actually there are multiple LNClients which return nil here and the method allows returning nil. I think we should actually check in the wails handler.

```
if (nodeStatus == nil) {
  return WailsRequestRouterResponse{Body: nil, Error: ""}
}
return WailsRequestRouterResponse{Body: *nodeStatus, Error: ""}
```

elnosh:
made change to do nil check in the handler

rolznz:
Thanks @elnosh !

Issue #969, opened 2025-01-10
ZapPlanner - OpenSats invalid address


sudocarlos:
The pre populated address for OpenSats in ZapPlanner is wrong. The correct address is opensats@vlt.ge per the lud16 in their nostr profile: https://njump.me/support@opensats.org

Also, can you add Brink there (the org that gives us Bitcoin Optech newsletter and funds Bitcoin Core devs)? Schmidty says the address is bitcoinbrink@zbd.gg

https://njump.me/nevent1qqs2jsjaxdhc4arwx4gal3s2lf3dqng07j7xgcxwq577f24ggjaex7qzyq2rn2759xq3vh4vengyd02kt2k3nu33f2f6mx7sntvruv6zhmye7qcyqqqqqqgpzemhxue69uhhyetvv9ujumn0wd68ytnzv9hxgqg5waehxw309aex2mrp0yhxgctdw4eju6t0qyxhwumn8ghj7mn0wvhxcmmvqn0g3x

PR #970, opened 2025-01-10
chore: add brink as zapplanner recipient, fix opensats lightning address

rolznz:
Fixes https://github.com/getAlby/hub/issues/969


reneaaron (approved):
utACK

Issue #957, opened 2025-01-08
LDK: connect to peers on startup in separate goroutine


rolznz:
Currently if connecting to any peers fails, it delays startup until the connection times out (+10 seconds).

Also double-check if the peers we have currently are actually accepting new connections.

Issue #175, opened 2024-03-15
Add healthcheck endpoint 


bumi:
The healthcheck should report if the nwc wallet is fully functioning and can be used internally (on the NWC UI) to show the status in the UI and externally to monitor the app.

+ connected to the relay
+ relay responds to some message
+ node is running
+ node can probe a route

rolznz:
@rdmitr I added some extra properties for the Alby Account status which could be used as part of this task https://github.com/getAlby/hub/pull/939

rolznz:
Current idea: the healthcheck should fail if any of the following are true:
- Alby Account status is failing / has incidents (above)
- One or more channels are offline
- Relay is currently not connected
- Cannot do a basic request from the node (not sure what is best, the probe command does not give a useful response)

PR #971, opened 2025-01-10
feat: health check endpoint

rdmitr:
Closes https://github.com/getAlby/hub/issues/957
Closes https://github.com/getAlby/hub/issues/175

image

image

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1027,12 +1027,12 @@ func (api *api) GetLogOutput(ctx context.Context, logType string, getLogRequest 
 func (api *api) Health(ctx context.Context) (*HealthResponse, error) {
 	var alarms []HealthAlarm
 
-	oauthStatus, err := api.albyOAuthSvc.GetInfo(ctx)
+	albyInfo, err := api.albyOAuthSvc.GetInfo(ctx)
 	if err != nil {
 		return nil, err
 	}
-	if !oauthStatus.Healthy {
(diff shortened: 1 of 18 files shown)


api/api.go
```diff
@@ -1024,6 +1024,65 @@ func (api *api) GetLogOutput(ctx context.Context, logType string, getLogRequest
 	return &GetLogOutputResponse{Log: string(logData)}, nil
 }
 
+func (api *api) Health(ctx context.Context) (*HealthResponse, error) {
+	var alarms []HealthAlarm
+
+	oauthStatus, err := api.albyOAuthSvc.GetInfo(ctx)
```
rolznz, line 1030: minor: could this field be renamed? it's the full info object right?

rolznz:
@rdmitr I removed the peers offline check as it does mean the node is not functional, I think the channels offline one is sufficient.

I also removed some of the rawDetails (especially because most of them are unmapped, and maybe we do not need such an advanced UI).

rolznz:
@reneaaron I did a first effort at the most basic UI, maybe we could huddle to review.

If there are no alarms, the health check will be green.

Things I think are not so great:
- It's just a tooltip - you cannot click it and probably users will not know what to do next.
- The Alby services one is vague and could be improved by reading the incidents, but probably we do not want to spend too much time on services being down.

rolznz:
Merging for now, we can do a follow-up iteration with @stackingsaunter 's design when it is ready

PR #978, opened 2025-01-12
chore: linux update script improvements

rolznz:
- add missing newlines
- use pwd to get current absolute directory, remove /albyhub suffix
- add file check to test if user is running in correct directory

### DIFF
--- a/scripts/linux-x86_64/update.sh
+++ b/scripts/linux-x86_64/update.sh
@@ -10,11 +10,12 @@ echo "You will have to unlock Alby Hub after the update."
 echo ""
 echo "Make sure you have your unlock password available and a backup of your seed."
 
-read -p "Do you want continue? (y/n): " -n 1 -r
+read -p "Do you want continue? (y/n):" -n 1 -r
 if [[ ! $REPLY =~ ^[Yy]$ ]]
 then
   exit
 fi
+echo ""
 
 sudo systemctl list-units --type=service --all | grep -Fq albyhub.service
 if [[ $? -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


scripts/linux-x86_64/update.sh
```diff
@@ -28,10 +29,17 @@ then
   pkill -f albyhub
 fi
 
-SCRIPT_DIR=$(dirname "$0")
-read -p "Absolute install directory path (default: $SCRIPT_DIR/albyhub): " USER_INSTALL_DIR
+SCRIPT_DIR=$(pwd)
```
bumi, line 32: this is wrong. `pwd` gives you the current directory. so for example:

I am in `/home/foo` (`cd /home/foo`) the `update.sh` is in `/home/bar/update.sh`.
If I now run `/home/bar/update.sh` the `/home/foo` directory will be used. which is wrong. 

you need to have the directory of the script here.


maybe we can also remove this and always expect that update.sh is in the install directory.  (as we write it also during installation)

bumi (commented):
this would need to be changed in the other install scripts, too.

scripts/linux-x86_64/update.sh
rolznz: Ah ok, thanks. I was wrong, and your SCRIPT_DIR change works in all console environments I tested.

So:
1. change `SCRIPT_DIR` to `SCRIPT_DIR=$(dirname "$(readlink -f "$0")");`
2. keep removal of `/albyhub` in `INSTALL_DIR` (to me it seems wrong it is appended)

bumi: not sure if I can follow.

it needs to be in sync with what we do here: https://github.com/getAlby/hub/blob/master/scripts/linux-x86_64/install.sh

rolznz: I don't think any changes need to be made in the install script.

We just need to make sure the update script runs in the correct folder.
1. The SCRIPT_DIR change above ensures that it picks the folder of the executing update script, as an absolute path rather than a relative one.
2. I think this has always been a bug and is just ignores that it cannot CD into a non-existent albyhub folder.

+ we have the small addition to check if we are in the right folder before proceeding

rolznz: I will apply the fix only to this install script for now. The other 2 install scripts are at least specified for RPI (which I do not this is right, at least the arm64 one) and expect the script is run from within the albyhub directory.

I will make a new issue to review that and reference this PR.

Issue #785, opened 2024-11-08
Add metadata to get_info NWC command


rolznz:
This allows us to return data not necessarily related to NWC e.g. keysend custom key info for the friends & family app.

PR #981, opened 2025-01-13
feat: get info metadata

rolznz:
Closes https://github.com/getAlby/hub/issues/785

See example output: https://github.com/getAlby/js-sdk/pull/287

Merges the existing app metadata with two general fields (the id and name of the app)

Allows keysend receive support to be added in getalby.com for accounts that are connected to an Alby Hub subwallet

### DIFF
--- a/nip47/controllers/get_info_controller.go
+++ b/nip47/controllers/get_info_controller.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 
 	"github.com/getAlby/hub/constants"
 	"github.com/getAlby/hub/db"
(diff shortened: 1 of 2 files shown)


nip47/controllers/get_info_controller.go
```diff
@@ -70,6 +72,27 @@ func (controller *nip47Controller) HandleGetInfoEvent(ctx context.Context, nip47
 		responsePayload.Network = network
 		responsePayload.BlockHeight = info.BlockHeight
 		responsePayload.BlockHash = info.BlockHash
+
+		if app != nil {
+			metadata := map[string]interface{}{}
+			if app.Metadata != nil {
+				jsonErr := json.Unmarshal(app.Metadata, &metadata)
```
bumi, line 79: this means we also return the metadata that the apps are setting?

rolznz: Yes, do you see any issues with this?

rolznz:
Merging for now

PR #989, opened 2025-01-13
Update "+ Submit your app" link

itstomekk:
Instead of Jotform (that is used for /discover) let's utilize that wiki for contributors: https://github.com/getAlby/hub/wiki/How-to:-submit-new-app-to-Hub's-Store


rolznz:
Should we put this guide in gitbook? do we need a wiki too?

MoritzKa:
It should be one place that we don't forget to update. 
Since the described app submission is more developer related than user related, a GH wiki page is fine for me.

PR #992, opened 2025-01-14
feat: display channel note above open channel buttons on increase incoming/outgoing flows

rolznz:
This allows us to alert users of important info about certain channel partners. Shows on both incoming and outgoing flows.

image

The UI/UX could be improved, it's a bit difficult because by default the user is not informed of who they open a channel with unless they open the advanced options.


bumi:
I am wondering if we should consider the on-chain channel open flow always "advanced" and show the "advanced options" by default?

PR #997, opened 2025-01-15
fix: allow unlink account before node start

rolznz:
I ran into the issue that I tried to unlink my account on a hub with expired tokens. I could not call this endpoint because I did not have a JWT yet, and you can only get one by starting the node.


bumi:
What could I do if I know a public URL of a hub?

rolznz:
@bumi the hub would be unlinked. The attacker needs to enter the unlock password to continue, which they do not know.

I do not think it can cause loss of funds, but it is still a vulnerability.

Also, this only happens when a refresh token expires which means the hub was not used for a long time. The alternative is we can handle this as a support case, and tell the user they need to delete certain keys from their sqlite DB. I think this is a better option. I will close this PR

rolznz:
Thanks for the review!

Issue #965, opened 2025-01-09
Add general node command API endpoint that can be executed from settings -> debug tools


rolznz:
for cases where we need to execute LNClient-specific functionality (like resetting the cashu wallet https://github.com/getAlby/hub/pull/962) it would be nice to have a single API call that LNClients can receive a command and arguments, and easily parse the request to execute some custom functionality.

PR #1007, opened 2025-01-17
feat: proposal: custom node commands

rdmitr:
Closes https://github.com/getAlby/hub/issues/965


rolznz:
I've added a simple frontend for it:

image

image

api/api.go
```diff
@@ -1068,6 +1069,91 @@ func (api *api) Health(ctx context.Context) (*HealthResponse, error) {
 	return &HealthResponse{Alarms: alarms}, nil
 }
 
+func (api *api) GetNodeCommands() (*NodeCommandsResponse, error) {
```
rolznz, line 1072: Minor: should we name it `GetCustomNodeCommands` (and related methods/types etc)?

http/http_service.go
```diff
@@ -999,6 +1001,35 @@ func (httpSvc *HttpService) getLogOutputHandler(c echo.Context) error {
...
+	}
+
+	execCommandResponse, err := httpSvc.api.ExecuteNodeCommand(c.Request().Context(), execCommandRequest.Command)
+	if err != nil {
+		return c.JSON(http.StatusInternalServerError, ErrorResponse{
+			Message: fmt.Sprintf("Failed to execute command: %v", err),
+		})
+	}
+
+	return c.JSONBlob(http.StatusOK, execCommandResponse)
```
rolznz, line 1030: Why do we return it this way rather than `c.JSON`?

Also the design of returning `RawJson` - is it possible instead to return an interface that we can then encode consistently like the other methods?

lnclient/cashu/cashu.go
```diff
@@ -375,3 +379,65 @@ func (cs *CashuService) GetSupportedNIP47NotificationTypes() []string {
...
+	switch command.Name {
+	case nodeCommandRestore:
+		return cs.executeCommandRestore(ctx)
+	case exampleCommandWithArg:
+		if len(command.Args) != 1 {
+			return nil, errors.New("please provide an argument")
+		}
+
+		return &lnclient.NodeCommandResponse{
+			RawJson: []byte("{\"" + command.Args[0].Name + "\": \"" + command.Args[0].Value + "\"}"),
```
rolznz, line 413: Here also if we returned an interface instead of RawJson I think it would be simpler?

utils/utils_test.go
```diff
@@ -0,0 +1,89 @@
...
+	testCases := []testCase{
+		{
+			name:            "empty input",
+			input:           "",
+			expectedSuccess: []string{},
+			expectedError:   "",
+		},
+		{
+			name:            "single argument",
+			input:           "arg1",
```
rolznz, line 28: isn't the first word the command? e.g. `command arg1 arg2`?

frontend/src/components/ExecuteCustomNodeCommandDialogContent.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <AlertDialogContent>
+      <form onSubmit={onSubmit}>
+        <AlertDialogHeader>
+          <AlertDialogTitle>Execute Custom Node Command</AlertDialogTitle>
+          <AlertDialogDescription className="text-left">
+            <Textarea
+              className="h-36 font-mono"
+              value={command}
+              onChange={(e) => setCommand(e.target.value)}
+              placeholder="commandname --arg1=value1"
```
rolznz, line 78: @rdmitr could you confirm this is the _ideal_ way for args to be passed?

utils/utils_test.go
rolznz: Also, can the tests be more aligned with reality? do we normally do `commandname --arg1=value1`?

rolznz:
I've tested the updates, still seem to be working as before. I think the tests need to be improved a bit and then from my side it's good to merge :rocket:

rolznz (approved):
tACK

bumi (commented):
utACK

why did you not split the command and the arguments input? the we would not need to do the string parsing, would we?

Issue #328, opened 2024-07-24
Design new 1-click App Connection flow that works on all environments


rolznz:
Try at https://paper-scissors-hodl.fly.dev/

On mobile this leaves the app connection completely broken - and copying the connection secret cannot work either because Alby Hub does not know it

**Update**: this cannot be fixed with nwc.getalby.com. We need to design a new 1-click app connection flow that works in all environments

bumi:
we miss this event then: 

```
// notify the opener if present of the success full connect
  window.addEventListener("load", (event) => {

    // dispatch a success event which can be listened to by the opener or by the app that embeddes the webview
    // this gives those apps the chance to know the user has enabled the connection
    const nwcEvent = new CustomEvent("nwc:success", { detail: {} });
    window.dispatchEvent(nwcEvent);

    // notify the opener of the successful connection
    if (window.opener) {
      window.opener.postMessage(
        {
          type: "nwc:success",
          payload: {success: true},
        },
        "*"
      );
    }
  });
```

bumi:
is this useEffect correct? https://github.com/getAlby/hub/blob/master/frontend/src/screens/apps/AppCreated.tsx#L76-L90

rolznz:
We need to design a new 1-click app connection flow that works in all environments

PR #1009, opened 2025-01-17
fix: new connection deeplink flow

rolznz:
Part of https://github.com/getAlby/hub/issues/328

The deeplink flow does not work well with a hash router, this will likely break for developers who try to implement it and inject query parameters to configure the new app connection.

We might need to update a few guides, etc. For now I added a redirect to ensure old links still work.

### DIFF
--- a/frontend/src/screens/apps/AppCreated.tsx
+++ b/frontend/src/screens/apps/AppCreated.tsx
@@ -61,7 +61,11 @@ function AppCreatedInternal() {
     }
     // dispatch a success event which can be listened to by the opener or by the app that embedded the webview
     // this gives those apps the chance to know the user has enabled the connection
-    const nwcEvent = new CustomEvent("nwc:success", { detail: {} });
+    const nwcEvent = new CustomEvent("nwc:success", {
+      detail: {
+        nostrWalletConnectUrl: pairingUri,
+      },
+    });
     window.dispatchEvent(nwcEvent);
 
     // notify the opener of the successful connection
@@ -69,12 +73,12 @@ function AppCreatedInternal() {
       window.opener.postMessage(
         {
           type: "nwc:success",
-          payload: { success: true },
+          nostrWalletConnectUrl: pairingUri,
         },
         "*"
       );
     }
-  }, [appstoreApp]);
+  }, [appstoreApp, pairingUri]);
 
   if (!createAppResponse) {
     return <Navigate to="/apps/new" />;
--- a/frontend/src/screens/apps/AppCreated.tsx
+++ b/frontend/src/screens/apps/AppCreated.tsx
@@ -56,6 +56,7 @@ function AppCreatedInternal() {
   }, [app?.lastEventAt, navigate, toast]);
 
   useEffect(() => {
+    // TODO: find a better way to only execute for deeplink flow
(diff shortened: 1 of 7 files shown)


frontend/src/screens/apps/AppCreated.tsx
```diff
@@ -61,20 +61,24 @@ function AppCreatedInternal() {
     }
     // dispatch a success event which can be listened to by the opener or by the app that embedded the webview
     // this gives those apps the chance to know the user has enabled the connection
-    const nwcEvent = new CustomEvent("nwc:success", { detail: {} });
+    const nwcEvent = new CustomEvent("nwc:success", {
```
rolznz, line 64: do we actually consume this custom event anywhere? 

I know `window.opener.postMessage` is consumed by Alby JS SDK. Do we need both?

rolznz: Ah, maybe this one is for a native app (I need to test it)

bumi:
why did we use the hashrouter? because of wails?

frontend/src/screens/apps/AppCreated.tsx
```diff
@@ -61,20 +61,24 @@ function AppCreatedInternal() {
...
+    });
     window.dispatchEvent(nwcEvent);
 
     // notify the opener of the successful connection
     if (window.opener) {
       window.opener.postMessage(
         {
           type: "nwc:success",
-          payload: { success: true },
+          nostrWalletConnectUrl: pairingUri,
```
bumi, line 76: can we avoid this. 
doesn't the client create the secret? and thus the pairing Uri does not need to be passed back in full.

places where the paring secret is passed should be avoided as it could be a security issue. (e.g. in what hack somebody could listen to this postMessage)

also best to avoid making changes with sensitive data in a non-senstive pr.

rolznz:
> why did we use the hashrouter? because of wails?

@bumi yes, Wails recommends it for the Wails app, I am not sure why they need it there. But BrowserRouter is the recommended for standard websites.

frontend/src/screens/apps/AppCreated.tsx
rolznz: It does not contain the secret.
But we could return just the `relayUrl` and `walletServicePubkey` instead, to be very clear. Maybe this is better anyway, since the app has to construct its own client.

rolznz: Because this is part of the deeplink flow, we would need to create a spec around this, and by doing it this way we force all other wallet services to use a client-created secret. But that's probably for the best.

im-adithya: I think passing the pairing uri without the secret param is good, would also help if some clients need lud16 for some reason (or if anything else is added in future to the pairing uri, instead of passing it separately)

im-adithya (approved):
tACK

frontend/src/screens/apps/AppCreated.tsx
rolznz: Reusing connections is a flaw and this also opens up security issues.
If the wallet doesn't support client-created secrets, then the user should use the standard copy paste flow.

PR #1016, opened 2025-01-20
Feat: NWA auth for self-hosted hubs

rolznz:
Closes https://github.com/getAlby/hub/issues/328 https://github.com/getAlby/hub/issues/1107

TODOs:
- [x] budget might still make sense if you have superuser access (new apps can be created to bypass the permission) - but users need to understand that other apps can be created with larger budgets, circumventing it. - If someone installs a malicious version of Alby Go it could drain a user's funds.
- [x] Limit superuser scope to Alby Go for now
- [x] TODOs in https://github.com/getAlby/hub/pull/907
- [x] Instead of adding the create connection method to the normal create connection UI, make Alby Go a custom detail page
- [x] Use standard NWC info event instead of new NWA event
- [x] Ensure superuser permission cannot be included as part of standard deeplink flow and NWA flow - an attacker could take advantage of this - _how do we explicitly limit it to Alby Go?_
- [x] User needs to be very clearly warned to be careful with this connection secret and not put them in other apps
- [x] create_connection: no spec for now
- [x] Alby detail page: give a default budget of 100K sats, editable after the connection is created
- [x] Remove "superuser" from standard connection card
- [x] Remove "superuser" text from permissions on view app connection page
- [x] Replace checkbox with "One tap connections" card and copy
- [x] NWA changes in JS SDK for testing - https://github.com/getAlby/js-sdk/pull/298
- [x] confirm way of passing lud16 as tag info event (but lightning addresses in profile metadata are leaked anyway in nostr and it's generally considered public info? the Alby relay does not allow crawling info events but this will not apply to other relays.)

Next steps:
- [x] review UX with Jakub
- [x] NWA in bitcoin connect - https://github.com/getAlby/bitcoin-connect/pull/279
- [x] UI for accepting a connection in Alby Go (MVP: show all the settings, and Alby Go can only choose yes or no. Ensure connection secret with create_connection method cannot be exported OR remove export function completely!) - https://github.com/getAlby/go/pull/272

Notes for spec later:
- `p` tag added to NWC info event
- `lud16` field added to get_info
- NWA:
  - no new NWA event - app subscribes to info event by `p` tag (same style as command responses - + no relay changes necessary)
  - no NWA secret
  - parameter changes: naming kept consistent with http flow


nip47/controllers/create_connection_controller.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/getAlby/hub/constants"
+	"github.com/getAlby/hub/logger"
+	"github.com/getAlby/hub/nip47/models"
+	"github.com/getAlby/hub/nip47/permissions"
+	"github.com/nbd-wtf/go-nostr"
+	"github.com/sirupsen/logrus"
+)
+
+type createConnectionBudgetParams struct {
+	Budget        uint64 `json:"budget"`
```
rolznz, line 17: I think these should be changed to be consistent with the http deeplink flow

nip47/models/models.go
```diff
@@ -10,6 +10,7 @@ const (
 	RESPONSE_KIND            = 23195
 	LEGACY_NOTIFICATION_KIND = 23196
 	NOTIFICATION_KIND        = 23197
+	NWA_EVENT_KIND           = 33195 // NIP-44
```
rolznz, line 13: remove

nip47/publish_nip47_info.go
```diff
@@ -56,6 +56,12 @@ func (svc *nip47Service) PublishNip47Info(ctx context.Context, relay nostrmodels
 		}
 		capabilities = svc.permissionsService.GetPermittedMethods(&app, lnClient)
 		permitsNotifications = svc.permissionsService.PermitsNotifications(&app)
+
+		// NWA: associate the info event with the app so that the app can receive the wallet pubkey
+		tags = append(tags, []string{"p", app.AppPubkey})
+		// also include the LN address for apps that use it (e.g. nostr apps assign lud16 to the profile)
+		lightningAddress, _ := svc.albyOAuthSvc.GetLightningAddress()
+		tags = append(tags, []string{"lud16", lightningAddress})
```
rolznz, line 64: check if `lightningAddress != ""`

nip47/controllers/create_connection_controller.go
```diff
@@ -0,0 +1,120 @@
...
+	RenewalPeriod string `json:"renewal_period"`
+}
+
+type createConnectionParams struct {
+	Pubkey    string                       `json:"pubkey"` // pubkey of the app connection
+	Name      string                       `json:"name"`
+	Methods   []string                     `json:"methods"`
```
rolznz, line 24: request_methods

rolznz: also add notification_types

frontend/src/screens/apps/NewApp.tsx
```diff
@@ -187,10 +185,6 @@ const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
 
     setLoading(true);
     try {
-      if (apps?.some((existingApp) => existingApp.name === appName)) {
```
rolznz, line 190: On the backend maybe we should automatically add a suffix e.g. `Alby Go (1)` if an app already exists with the same name

nip47/publish_nip47_info.go
```diff
@@ -56,6 +56,14 @@ func (svc *nip47Service) PublishNip47Info(ctx context.Context, relay nostrmodels
 		}
 		capabilities = svc.permissionsService.GetPermittedMethods(&app, lnClient)
 		permitsNotifications = svc.permissionsService.PermitsNotifications(&app)
+
+		// NWA: associate the info event with the app so that the app can receive the wallet pubkey
+		tags = append(tags, []string{"p", app.AppPubkey})
+		// also include the LN address for apps that use it (e.g. nostr apps assign lud16 to the profile)
+		lightningAddress, _ := svc.albyOAuthSvc.GetLightningAddress()
```
rolznz, line 63: Remove this, leaks metadata and allows searching by relay for multiple connections by a lightning address

api/api.go
```diff
@@ -67,6 +67,13 @@ func (api *api) CreateApp(createAppRequest *CreateAppRequest) (*CreateAppRespons
 			"sub-wallets are currently not supported on your node backend. Try LDK or LND")
 	}
 
+	if slices.Contains(createAppRequest.Scopes, constants.SUPERUSER_SCOPE) {
+		if !api.cfg.CheckUnlockPassword(createAppRequest.UnlockPassword) {
+			return nil, fmt.Errorf(
+				"cannot create app with superuser permission without specifying unlock password")
```
im-adithya, line 73: (minor) Can we have two checks? Because even if I give the wrong password it says I didn't specify

nip47/controllers/create_connection_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	if len(scopes) == 0 {
+		publishResponse(&models.Response{
+			ResultType: nip47Request.Method,
+			Error: &models.Error{
+				Code:    constants.ERROR_INTERNAL,
+				Message: "No methods provided",
+			},
+		}, nostr.Tags{})
+		return
+	}
```
im-adithya, line 109: Can this be moved to `CreateApp` itself?

nip47/controllers/create_connection_controller_test.go
```diff
@@ -0,0 +1,331 @@
...
+
+	publishResponse := func(response *models.Response, tags nostr.Tags) {
+		publishedResponse = response
+	}
+
+	permissionsSvc := permissions.NewPermissionsService(svc.DB, svc.EventPublisher)
+	transactionsSvc := transactions.NewTransactionsService(svc.DB, svc.EventPublisher)
+	albyOAuthSvc := alby.NewAlbyOAuthService(svc.DB, svc.Cfg, svc.Keys, svc.EventPublisher)
+	NewNip47Controller(svc.LNClient, svc.DB, svc.EventPublisher, permissionsSvc, transactionsSvc, svc.AppsService, albyOAuthSvc).
+		HandleCreateConnectionEvent(ctx, nip47Request, dbRequestEvent.ID, publishResponse)
```
im-adithya, line 65: nit: all this is repeated across tests every time, can we somehow DRY this part?

service/create_app_consumer.go
```diff
@@ -33,6 +34,18 @@ func (s *createAppConsumer) ConsumeEvent(ctx context.Context, event *events.Even
...
+	err := s.svc.db.First(&app, &db.App{
+		ID: id,
+	}).Error
+	if err != nil {
+		logger.Logger.WithFields(logrus.Fields{
+			"id": id,
+		}).WithError(err).Error("Failed to find app for id")
+		return
+	}
+
```
im-adithya, line 48: Whoops 😄 Nice catch!

frontend/src/components/Scopes.tsx
```diff
@@ -87,10 +87,17 @@ const Scopes: React.FC<ScopesProps> = ({
   }, [capabilities.scopes]);
 
   const [scopeGroup, setScopeGroup] = React.useState<ScopeGroup>(() => {
-    if (isolated && scopes.length === capabilities.scopes.length) {
+    if (
+      isolated &&
+      scopes.length === fullAccessScopes.length &&
+      scopes.every((scope) => fullAccessScopes.includes(scope))
```
im-adithya, line 93: unrelated to this PR but shouldn't this be `isolatedScopes`? Since it doesn't have `sign_message` (is that intentional btw)

frontend/src/components/Scopes.tsx
```diff
@@ -52,7 +52,7 @@ const Scopes: React.FC<ScopesProps> = ({
   onScopesChanged,
 }) => {
   const fullAccessScopes: Scope[] = React.useMemo(() => {
-    return [...capabilities.scopes];
+    return capabilities.scopes.filter((scope) => scope !== "superuser");
```
im-adithya, line 55: (backend) The capabilities will never have this, we should return `"superuser"` in `api.GetWalletCapabilities`

(or add `create_connection` in `lnclient.GetSupportedNIP47Methods` however I don't have a good feeling about this as it is rather a _Hub_ supported nip47 method)

frontend/src/screens/internal-apps/AlbyGo.tsx
```diff
@@ -0,0 +1,321 @@
...
+        scopes: [
+          "pay_invoice",
+          "get_balance",
+          "get_info",
+          "make_invoice",
+          "lookup_invoice",
+          "list_transactions",
+          "sign_message",
+          "notifications",
+          "superuser",
```
im-adithya, line 70: Fetch these scopes from `useCapabilities` to avoid errors in other backends

nip47/permissions/permissions.go
```diff
@@ -79,6 +79,11 @@ func (svc *permissionsService) GetPermittedMethods(app *db.App, lnClient lnclien
 	// only return methods supported by the lnClient
 	lnClientSupportedMethods := lnClient.GetSupportedNIP47Methods()
 	requestMethods = utils.Filter(requestMethods, func(requestMethod string) bool {
+		// TODO: better way to exclude methods unrelated to the lnclient
+		if requestMethod == models.CREATE_CONNECTION_METHOD {
+			return true
+		}
```
im-adithya, line 85: We can remove this as currently this is not hit at all since `GetSupportedNIP47Methods` doesn't return `create_connection`

im-adithya: If we can have a function like `GetServiceSupportedNIP47Methods` then it could be used both here and in `api.GetWalletCapabilities` like mentioned above

api/api.go
rolznz: hmm. We already have a `required` on the frontend so the user can't provide an empty password, and that's the only way right now. I'll update the error message to `"incorrect unlock password to create app with superuser permission"`

apps/apps_service.go
```diff
@@ -44,6 +45,14 @@ func (svc *appsService) CreateApp(name string, pubkey string, maxAmountSat uint6
...
 	}
 
+	if budgetRenewal == "" {
+		budgetRenewal = constants.BUDGET_RENEWAL_NEVER
+	}
+
+	if !slices.Contains(constants.GetBudgetRenewals(), budgetRenewal) {
+		return nil, "", fmt.Errorf("invalid budget renewal. Must be one of %s", strings.Join(constants.GetBudgetRenewals(), ","))
+	}
+
```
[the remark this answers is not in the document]
rolznz: I think in the frontend we don't allow an empty string.

Now just above I also added this to ensure we don't use an empty string:

```
if budgetRenewal == "" {
	budgetRenewal = constants.BUDGET_RENEWAL_NEVER
}
```

Before these checks it was possible to execute the API with an incorrect budget renewal and we would just allow it.

nip47/controllers/create_connection_controller.go
rolznz: this is difficult - the apps service deals with scopes, not request_methods or notification_types, and by doing that we will have unclear error messages for requests via NWC. There might be a small amount I can move.

api/api.go
im-adithya: Sorry I forgot to update the comment, previously there wasn't `required` so I ran into this, `incorrect unlock password` should be good 👍

nip47/controllers/create_connection_controller.go
im-adithya: Ah then just moving the `len(scopes) == 0` check should be good ig

frontend/src/screens/apps/NewApp.tsx
rolznz: done

frontend/src/components/Scopes.tsx
rolznz: fixed, thanks!

rolznz: it's intentional to not give a sub wallet sign_message currently (although we do say sub-wallets should still be used by "trusted" users)

frontend/src/components/Scopes.tsx
rolznz: I have undone this change, thanks

frontend/src/screens/internal-apps/AlbyGo.tsx
rolznz: thanks, fixed

nip47/permissions/permissions.go
rolznz: Thanks for checking. Currently we don't return it in `api.GetWalletCapabilities`. For now I would keep it as-is. I'm not sure there is any point returning service capabilities because it's constant, no matter what wallet backend

tests/create_app.go
```diff
@@ -27,7 +29,8 @@ func CreateAppWithPrivateKey(svc *TestService, senderPrivkey, nip47Version strin
 	}
 
 	var expiresAt *time.Time
-	app, pairingSecretKey, err := svc.AppsService.CreateApp("test", senderPubkey, 0, "monthly", expiresAt, nil, false, nil)
+	appsSvc := apps.NewAppsService(svc.DB, svc.EventPublisher, svc.Keys, svc.Cfg)
```
im-adithya, line 32: Same as above

nip47/controllers/create_connection_controller.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// ensure there is at least one request method
+	if len(params.RequestMethods) == 0 {
+		publishResponse(&models.Response{
+			ResultType: nip47Request.Method,
+			Error: &models.Error{
+				Code:    constants.ERROR_INTERNAL,
+				Message: "No methods provided",
```
im-adithya, line 73: I'm assuming this is because otherwise the user will see `scopes` instead of `methods`. So should this be `request methods`

nip47/controllers/get_info_controller_test.go
```diff
@@ -130,7 +128,8 @@ func TestHandleGetInfoEvent_WithMetadata(t *testing.T) {
 		"a": 123,
 	}
 
-	app, _, err := svc.AppsService.CreateApp("test", "", 0, "monthly", nil, nil, false, metadata)
+	appsSvc := apps.NewAppsService(svc.DB, svc.EventPublisher, svc.Keys, svc.Cfg)
+	app, _, err := appsSvc.CreateApp("test", "", 0, "monthly", nil, []string{constants.GET_INFO_SCOPE}, false, metadata)
 	assert.NoError(t, err)
```
im-adithya, line 133: Why are we initializing `AppsService` again? Can't we do `svc.AppsService.CreateApp`?

im-adithya (approved):
tACK 🎉

Issue #1008, opened 2025-01-17
Explain how wallet is used in app store apps


rolznz:
A lot of apps in our Alby Hub app store have instructions, but I think it would be nice to add a short description of how it interacts with your wallet.

E.g. for Amethyst it could be: _By connecting your hub you can zap content from people you follow_

For Zap Store: _By connecting your hub you can support the developers of apps you install from zapstore_

rolznz:
Current UI:

![Image](https://github.com/user-attachments/assets/6b23d473-00f8-4871-8b74-f9115b8e1ce1)

rolznz:
@stackingsaunter any chance you could prioritize a design for this one?

stackingsaunter:
I would use it as a separate card:

![Image](https://github.com/user-attachments/assets/c4b49bb3-8465-46de-b662-406cd2e3ec94)

When it's connected, this card can be in place of "Get This App":

![Image](https://github.com/user-attachments/assets/20648fcc-7337-45b3-b9cb-25eb82de1a97)

PR #1022, opened 2025-01-21
feat: show hub interactions on app store detail pages

rolznz:
Closes https://github.com/getAlby/hub/issues/1008

image

Copy needs review.

Also commented out nostr.kiwi for now, seems broken.

### DIFF
--- a/frontend/src/components/SuggestedAppData.tsx
+++ b/frontend/src/components/SuggestedAppData.tsx
@@ -50,7 +50,7 @@ export type SuggestedApp = {
   chromeLink?: string;
   firefoxLink?: string;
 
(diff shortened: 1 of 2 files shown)


MoritzKa:
Is it about the interactions between Alby Hub and the app or is it about a short description what you can do in this app?

I'd vote for the latter and rename the title from "Alby Hub Interaction" to "About this App"

I added minor changes. Most apps are similar so I would just copy the description over. 
What's easier to handle for you? An own PR or comments as above?

rolznz:
@MoritzKa you could edit this branch directly, or give me comments and I can address them

PR #1024, opened 2025-01-24
fix: change sticky to fixed

im-adithya:
Some adblockers remove sticky elements, and that causes the sidebar to not appear in Safari with ADBlock extension.

This fix replaces it with`fixed` class instead to prevent that from happening. (Also does the same for mobile just-in-case)

image

### DIFF
--- a/frontend/src/components/layouts/AppLayout.tsx
+++ b/frontend/src/components/layouts/AppLayout.tsx
@@ -225,7 +225,7 @@ export default function AppLayout() {
       <div className="font-sans min-h-screen w-full flex flex-col">
         <div className="flex-1 h-full md:grid md:grid-cols-[280px_minmax(0,1fr)]">
           <div className="hidden border-r bg-muted/40 md:block">
-            <div className="flex h-full max-h-screen flex-col gap-2 sticky z-10 top-0 overflow-y-auto">
+            <div className="flex h-full max-h-screen flex-col gap-2 fixed w-[280px] z-10 top-0 overflow-y-auto">
               <div className="flex-1">
                 <nav className="grid items-start px-4 py-2 text-sm font-medium">
                   <div className="p-3 flex justify-between items-center mt-2 mb-6">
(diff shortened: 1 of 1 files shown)


frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -225,7 +225,7 @@ export default function AppLayout() {
       <div className="font-sans min-h-screen w-full flex flex-col">
         <div className="flex-1 h-full md:grid md:grid-cols-[280px_minmax(0,1fr)]">
           <div className="hidden border-r bg-muted/40 md:block">
-            <div className="flex h-full max-h-screen flex-col gap-2 sticky z-10 top-0 overflow-y-auto">
+            <div className="flex h-full max-h-screen flex-col gap-2 fixed w-[280px] z-10 top-0 overflow-y-auto">
```
rolznz, line 228: just wondering, why are also the width changes needed on all of them?

frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -314,7 +314,7 @@ export default function AppLayout() {
                 </DropdownMenu>
               </Sheet>
             </header>
-            <div className="flex flex-1 flex-col gap-4 p-4 lg:gap-6 lg:p-8">
+            <div className="flex flex-1 flex-col gap-4 p-4 pt-[4.5rem] md:pt-4 lg:gap-6 lg:p-8">
```
rolznz, line 317: we try to avoid using custom units, why is pt-[4.5rem] needed?

rolznz:
@im-adithya was there an issue for this to link to? also I think it should be a `fix:` not a `chore:`

frontend/src/components/layouts/AppLayout.tsx
im-adithya: This adds the height of the header (`h-14` i.e `56px`) to `p-4` (`16px`) that already exists making it `72px` (`4.5 rem`)

im-adithya: Added margin instead

im-adithya:
> @im-adithya was there an issue for this to link to?

Nope, no issue, changed title and commit to fix from chore

rolznz (approved):
tACK

Issue #1034, opened 2025-01-28
Re-enable Subwallets for Phoenixd


rolznz:
A few users said this worked in the past and now is explicitly disabled.

In https://github.com/getAlby/hub/pull/786 we moved the check to the backend - which I think before the Friends & Family app frontend did not have this check.

Unfortunately Phoenixd also does not provide a preimage when an invoice is created, but can be fetched by looking up the transaction.

PR #1035, opened 2025-01-28
feat: phoenixd subwallets

rolznz:
Closes https://github.com/getAlby/hub/issues/1034

A few users said this used to work and no longer does.

It didn't properly work because we did not set the preimage on the transaction, but it is available by looking up the invoice.


rolznz:
@bumi I tested the changes in testnet3 sending and receiving from another node. The transaction list looks good.

PR #1038, opened 2025-01-28
fix: do not show 0 in withdraw alert

### LINKED ISSUE #1029: Bug: "0" in withdraw savings balance alert message when user has no channels
Image

im-adithya:
Fixes #1029


rolznz (approved):
Thanks! utACK

PR #1039, opened 2025-01-28
chore: remove network info and debug info in node status

im-adithya:
This causes issues because the payload becomes too large and it is also not related to the node status.

And users run LND elsewhere so it's best to keep it short as they can do debugging in the LND app itself.


rolznz (approved):
utACK

rolznz:
@im-adithya was there an issue for this?

Issue #842, opened 2024-12-02
Add security page as part of onboarding flow


rolznz:
From @bumi :

I was actually thinking of adding some "security" page. Which describes all the details from the importance of the password, the seed, info about backup options, and also to tell the user not to put ALL their funds on there as it's still a new product. So we can link there and people have to acknowledge it similar to a ToS.

stackingsaunter:
Interesting idea. But why you think this cannot be a part of Guides?

I think importance of seed and password should be highlighted enough when they are introduced (+ in settings when you can change them). I am not sure how a seperate page adds anything?

rolznz:
@bumi @stackingsaunter  I thought the user had to explicitly acknowledge it. I doubt many people will click an optional link underneath "Get Started".

stackingsaunter:
I think we need to sync on this

rolznz:
Initial design from @stackingsaunter 
![image](https://github.com/user-attachments/assets/6b8e322c-369c-4a9a-8236-998a43be9a7f)

rolznz:
Do we want some experimental warning: Do not put all your funds on Alby Hub?

stackingsaunter:
I don't know, do we know any case someone did that?

stackingsaunter:
@rolznz I added it, although not sure If we need it


![image](https://github.com/user-attachments/assets/35e4f207-7d91-4f31-a6c1-28e0e04ea214)


https://www.figma.com/design/2IyilDewDrdNP3G1NixPYd/%F0%9F%8C%93-Alby-Hub?node-id=6490-76423&t=ijjCFf5CEh1wCsPb-4

MoritzKa:
How does this page look different for users with connected LND nodes?

rolznz:
For LND I think we should be clear right now we do not handle channel backups, with a link to a guide page to learn more (that this must be done through Umbrel/Start9 backups or manually if the user has a more custom setup)

MoritzKa:
I thought about adding to point 3 from Jakub's design: 
Your wallet can be recovered or migrated only with the 12-word recovery phrase. If you connect an external lightning node, you'll need to recover and migrate that node directly without Alby Hub.

But I would keep it simple without further special cases that might confuse users during the onboarding and just keep as suggested above: Your wallet can be recovered or migrated only with the 12-word recovery phrase.

Issue #823, opened 2024-11-25
albyhub isn't open source - signing up for albyhub requires a EULA


sneak:
> Subject to the following terms, Alby grants to you a perpetual, worldwide license to the Application pursuant to the Apache-2.0 license. (https://github.com/getAlby/hub?tab=Apache-2.0-1-ov-file#readme). This EULA pertains solely to the Application and does not limit your rights under, or grant you rights that supersede, the license terms of any particular component.

This is a contradiction.  Either I have rights to the application under Apache 2.0, or I have to abide by these terms.  The Apache 2.0 license is not "Subject to the following terms" - it is the complete license to the software.

If there are additional terms to which I am subject, then the application is NOT licensed under Apache 2.0, as the Apache 2.0 license does not have those additional terms.

bumi:
Thanks for bringing this up! quite important actually. 

The account (which has some ToS) is optional and not required by the app actually. but I have to review if there that needs to be reflected here.

MoritzKa:
Thanks for highlighting it. Which paragraphs of the EULA terms interfere with the Apache License? 

We could highlight more the original intention that that the EULA refers only to the combination of using Alby Hub together with an Alby account.

sneak:
> *Which paragraphs of the EULA terms interfere with the Apache License?*

All of them.  The Apache license defines a set of licensing terms.  If you add any other terms, such as a EULA, it is no longer the set of terms that is defined by the name “Apache license”.  It is also thus not open source, as the definition of open source includes “no EULA”.

EULAs are incompatible with open source software by their nature.  Open source software, also known as free software, preserves the freedoms of a user of software. EULAs restrict those freedoms.

MoritzKa:
Thanks. Is there a concrete example in the current EULA that restricts the use of the software?
I am learning a lot here. Let's check if Alby's Terms of Service already cover the relation between the user and Alby, the company, sufficiently.

MoritzKa:
I am fine with removing the EULA and privacy policy consent from this screen, because it relates only to the combined usage with an Alby account. However Alby accounts have their own comprehensive terms of service and privacy policy and in the meantime Alby Hub can be used without an Alby account.   
![image](https://github.com/user-attachments/assets/4b552dc0-9774-4299-835b-5b48d0b2555a)

PR #1043, opened 2025-01-29
feat: onboarding security page

rolznz:
Closes https://github.com/getAlby/hub/issues/842
Closes https://github.com/getAlby/hub/issues/823

Removes the TOS on the welcome screen and adds a new security page as the last step of the node creation process.

LDK & Cashu

image

LND & Phoenix

image

LDK (no Alby Account)

image


frontend/src/screens/setup/SetupSecurity.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  {!info?.albyUserIdentifier &&
+                  store.nodeInfo.backendType === "LDK" ? (
+                    <>
+                      Your on-chain balance can be recovered only with your
+                      12-word recovery phrase. You must also take care of your
+                      own channel backups.
+                    </>
+                  ) : (
+                    <>
+                      Your wallet can be recovered only with your 12-word
```
rolznz, line 79: we have not introduced the recovery phrase at this point

MoritzKa:
Thanks for removing the ToS. 
On the LND & Phoenix page, we could remove "LND" from point 3  and just say: "Channel backups...." , because it is for both LND and Phoenix.

rolznz:
> On the LND & Phoenix page, we could remove "LND" from point 3 and just say: "Channel backups...." , because it is for both LND and Phoenix.

@MoritzKa thanks, I will remove it. I indirectly was trying to hint that we only support channel backups on LDK which is probably useless. I changed it to this, what do you think? 

image

frontend/src/screens/setup/SetupSecurity.tsx
rolznz: I updated this

rolznz:
I have made some minor copy improvements. We can iterate on this page over time.

MoritzKa:
> @MoritzKa thanks, I will remove it. I indirectly was trying to hint that we only support channel backups on LDK which is probably useless. I changed it to this, what do you think?

Good point. We could mention both. "LND and Phoenixd channel backups are not handled ..."

But I like the suggestion: "...go back and choose the LDK node type." 👍

PR #1044, opened 2025-01-29
feat: add option to edit fee while withdrawing

### LINKED ISSUE #684: Withdraw Savings Balance: allow user to set fee-rate
Fee estimation isn't great here. Ended up having a tx broadcast at quite a low rate with no way to bump it.

It would be nice to set the fee rate on this view.

### LINKED ISSUE #1288: Withdraw Savings Balance: allow user to set custom fee-rate
In https://github.com/getAlby/hub/issues/684 we didn't actually allow the user to set the fee rate, but updated LND to ensure the fee is set to 1 block.

This is not 100% perfect because it doesn't account for spikes in time critical situations (e.g. swaps/ P2P exchange)

See https://feedback.getalby.com/-alby-hub-request-a-feature/posts/see-and-set-miner-fee-when-withdrawing-savings-balance

Since LDK-node 0.5 we can now set a custom fee rate.

@im-adithya wrote some code originally here: https://github.com/getAlby/hub/pull/1044 - maybe it could be hidden behind our standard "Advanced" toggle?

im-adithya:
# Description

Fixes #1288

This adds the option to edit the fee rate while withdrawing on-chain balance. The **default** is set to **fastest**, but can be edited on clicking "Advanced Options" (min of the number input is `recommendedFees.minimumFee`)

:warning: Adds only in LND and LDK users for now

## Screenshots

<table>
<tr>
<td></td>
<td></td>
<td></td>
</tr>


frontend/src/screens/wallet/WithdrawOnchainFunds.tsx
```diff
@@ -239,6 +262,36 @@ export default function WithdrawOnchainFunds() {
...
+            <>
+              {showAdvanced && (
+                <div className="">
+                  <Label htmlFor="fee-rate">Fee Rate (Sat/Vbyte)</Label>
+                  <Input
+                    id="fee-rate"
+                    type="number"
+                    value={feeRate}
+                    required
+                    min={recommendedFees.minimumFee}
```
im-adithya, line 275: Should we also add `max` with `recommendedFees.fastestFee` to save users?

api/api.go
```diff
@@ -637,11 +637,11 @@ func (api *api) SignMessage(ctx context.Context, message string) (*SignMessageRe
 	}, nil
 }
 
-func (api *api) RedeemOnchainFunds(ctx context.Context, toAddress string, amount uint64, sendAll bool) (*RedeemOnchainFundsResponse, error) {
+func (api *api) RedeemOnchainFunds(ctx context.Context, toAddress string, amount uint64, feeRate float64, sendAll bool) (*RedeemOnchainFundsResponse, error) {
```
rolznz, line 640: I don't think the feeRate needs to be a float, does it? it's in sats/vByte, how many users will choose e.g. `2.5`?

frontend/src/screens/wallet/WithdrawOnchainFunds.tsx
rolznz: I don't think so, because it doesn't account for fee spikes. Would a confirmation would be better if the user chooses a rate larger than the fastest fee?

rolznz:
From @reneaaron : we should just set the highest fee rate (needs re-review)

frontend/src/screens/wallet/WithdrawOnchainFunds.tsx
im-adithya: Added confirmation alert

Screenshot 2025-03-12 at 2 00 35 PM

lnclient/lnd/lnd.go
```diff
@@ -970,11 +970,12 @@ func (svc *LNDService) GetOnchainBalance(ctx context.Context) (*lnclient.Onchain
...
-func (svc *LNDService) RedeemOnchainFunds(ctx context.Context, toAddress string, amount uint64, sendAll bool) (txId string, err error) {
+func (svc *LNDService) RedeemOnchainFunds(ctx context.Context, toAddress string, amount uint64, feeRate float64, sendAll bool) (txId string, err error) {
 	resp, err := svc.client.SendCoins(ctx, &lnrpc.SendCoinsRequest{
-		Addr:    toAddress,
-		SendAll: sendAll,
-		Amount:  int64(amount),
+		Addr:        toAddress,
+		SendAll:     sendAll,
+		Amount:      int64(amount),
+		SatPerVbyte: uint64(feeRate),
```
rolznz, line 978: Can we use `TargetConf: 1` instead? this might use the highest fee rate? 

```
// ... If 
// neither target_conf, or sat_per_vbyte are set, then the internal wallet will
// consult its fee model to determine a fee for the default confirmation
// target.
```

api/api.go
im-adithya: But the highest fee (default setting) which comes from the mempool api can be a float right? Or is it set to be an integer?

im-adithya:
Tested on both LDK (mutinynet) and LND (testnet) ✅

lnclient/ldk/ldk.go
```diff
@@ -1195,19 +1195,22 @@ func (ls *LDKService) GetOnchainBalance(ctx context.Context) (*lnclient.OnchainB
 	}, nil
 }
 
-func (ls *LDKService) RedeemOnchainFunds(ctx context.Context, toAddress string, amount uint64, sendAll bool) (string, error) {
+func (ls *LDKService) RedeemOnchainFunds(ctx context.Context, toAddress string, amount uint64, feeRate uint64, sendAll bool) (string, error) {
+	var fee *ldk_node.FeeRate
+	fee = ldk_node.FeeRateFromSatPerVbUnchecked(feeRate)
```
rolznz, line 1200: @im-adithya  I think we should only do this if a `feeRate` is set. By default we should not set the fee rate and we should let the node decide.

frontend/src/screens/wallet/WithdrawOnchainFunds.tsx
```diff
@@ -31,14 +38,29 @@ import { request } from "src/utils/request";
...
   const { data: channels } = useChannels();
   const [onchainAddress, setOnchainAddress] = React.useState("");
   const [amount, setAmount] = React.useState("");
+  const [feeRate, setFeeRate] = React.useState("");
   const [sendAll, setSendAll] = React.useState(false);
+  const [showAdvanced, setShowAdvanced] = React.useState(false);
   const [transactionId, setTransactionId] = React.useState("");
   const [confirmDialogOpen, setConfirmDialogOpen] = React.useState(false);
 
+  React.useEffect(() => {
```
rolznz, line 58: I don't think we should do it this way. This will break if the fee rates cannot be loaded.

lnclient/lnd/lnd.go
```diff
@@ -1219,12 +1219,12 @@ func (svc *LNDService) GetOnchainBalance(ctx context.Context) (*lnclient.Onchain
...
+func (svc *LNDService) RedeemOnchainFunds(ctx context.Context, toAddress string, amount uint64, feeRate uint64, sendAll bool) (txId string, err error) {
 	resp, err := svc.client.SendCoins(ctx, &lnrpc.SendCoinsRequest{
-		Addr:       toAddress,
-		SendAll:    sendAll,
-		Amount:     int64(amount),
-		TargetConf: 1,
+		Addr:        toAddress,
+		SendAll:     sendAll,
+		Amount:      int64(amount),
+		SatPerVbyte: feeRate,
```
rolznz, line 1227: I think `TargetConf: 1` should be set instead of SatPerVbyte unless the explicitly sets the fee rate.

lnclient/ldk/ldk.go
im-adithya: Done (changed to be a pointer)

rolznz (approved):
two minor comments, otherwise tACK

frontend/src/assets/lotties/loading-dark.json
```diff
@@ -1 +1,333 @@
...": 4, "o": { "a": 0, "k": 100, "ix": 4 }, "w": { "a": 0, "k": 5, "ix": 5 }, "c": { "a": 0, "k": [255, 255, 255], "ix": 3 } }, { "ty": "fl", "bm": 0, "hd": false, "mn": "ADBE Vector Graphic - Fill", "nm": "Fill 1", "c": { "a": 0, "k": [0.5882, 0.5882, 0.5882], "ix": 4 }, "r": 1, "o": { "a": 0, "k": 0, "ix": 5 } }, { "ty": "tr", "a": { "a": 0, "k": [0, 0], "ix": 1 }, "s": { "a": 0, "k": [100, 100], "ix": 3 }, "sk": { "a": 0, "k": 0, "ix": 4 }, "p": { "a": 0, "k": [1.5, 1.5], "ix": 2 }, "r": { "a": 0, "k": 0, "ix": 6 }, "sa": { "a": 0, "k": 0, "ix": 5 }, "o": { "a": 0, "k": 100, "ix": 7 } } ] }, { "ty": "tm", "bm": 0, "hd": false, "mn": "ADBE Vector Filter - Trim", "nm": "Trim Paths 1", "ix": 2, "e": { "a": 1, "k": [ { "o": { "x": 0.927, "y": 0 }, "i": { "x": 0.833, "y": 0.833 }, "s": [0], "t": 0 }, { "s": [100], "t": 34.0000013848484 } ], "ix": 2 }, "o": { "a": 1, "k": [ { "o": { "x": 0.794, "y": 0 }, "i": { "x": 0.3, "y": 1 }, "s": [0], "t": 0 }, { "s": [360], "t": 69.0000028104276 } ], "ix": 3 }, "s": { "a": 1, "k": [ { "o": { "x": 0.167, "y": 0.167 }, "i": { "x": 0.485, "y": 1 }, "s": [0], "t": 22 }, { "s": [100], "t": 69.0000028104276 } ], "ix": 1 }, "m": 1 } ], "ind": 2 } ] } ]}
```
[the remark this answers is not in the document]
im-adithya: I just hit Save and it formatted it, it's just prettier/linting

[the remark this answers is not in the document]
rolznz: I don't think we should lint these files, they're not human generated or intended to be modified by a human. Maybe the folder can be added to the prettier ignore file?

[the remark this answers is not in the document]
rolznz: or at least, undone in this PR?

lnclient/ldk/ldk.go
```diff
@@ -1195,22 +1195,27 @@ func (ls *LDKService) GetOnchainBalance(ctx context.Context) (*lnclient.OnchainB
...
 
-func (ls *LDKService) RedeemOnchainFunds(ctx context.Context, toAddress string, amount uint64, feeRate uint64, sendAll bool) (string, error) {
-	var fee *ldk_node.FeeRate
-	fee = ldk_node.FeeRateFromSatPerVbUnchecked(feeRate)
+func (ls *LDKService) RedeemOnchainFunds(ctx context.Context, toAddress string, amount uint64, feeRate *uint64, sendAll bool) (string, error) {
+	var feePtr **ldk_node.FeeRate
+	if feeRate != nil {
+		fee := ldk_node.FeeRateFromSatPerVbUnchecked(*feeRate)
+		feePtr = &fee
+	} else {
```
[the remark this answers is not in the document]
im-adithya: Done!

PR #1046, opened 2025-01-30
fix: do not allow creating duplicate pending transactions

rolznz:
We originally relied on the LNClient to not allow creating duplicate pending transactions, but there seems to be a race condition that allows the creation of two pending transactions in our transactions database. When the payment is settled, only one of the pending transactions is marked as sent, and the other is left pending, which creates issues with budgets and especially sub-wallets (the money is stuck and forever inaccessible)

### DIFF
--- a/transactions/payments_test.go
+++ b/transactions/payments_test.go
@@ -88,7 +88,7 @@ func TestSendPaymentSync_MetadataTooLarge(t *testing.T) {
 	assert.Nil(t, transaction)
 }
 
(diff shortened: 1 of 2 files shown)


bumi:
is there no constraint on the DB?

A find will also not fully prevent a race condition.

rolznz:
@bumi 

> is there no constraint on the DB?

No. For example if you try to pay once and fail, and then pay and succeed, this will be two different entries

> A find will also not fully prevent a race condition.

Here it's done within a database transaction.

PR #1051, opened 2025-01-31
feat: add same counterparty alerts while opening channels

### LINKED ISSUE #1027: Add alert when users plan to open another channel with the same counterparty
This in probably the majority of cases is not what the user actually wants.

Other options: 
- Choose a different LSP/counterparty
- Use swaps / manage their existing channels better (link to node page -> advanced)

We should add a generic link to a guide so that we can easily update the top 2-3 recommendations.

im-adithya:
Fixes #1027

Adds an alert when users try to open a channel with the same counterparty.

### Using Lightning
Screenshot 2025-01-31 at 2 36 43 PM
Screenshot 2025-01-31 at 2 36 57 PM

### Using Onchain (recommended peers)
Screenshot 2025-01-31 at 2 37 12 PM
Screenshot 2025-01-31 at 2 37 25 PM

### Using Onchain (custom peers)
Screenshot 2025-01-31 at 2 38 31 PM


frontend/src/screens/channels/IncreaseIncomingCapacity.tsx
```diff
@@ -212,6 +214,7 @@ function NewChannelInternal({
           </div>
         }
       />
+      <PeerAlert pubkey={selectedPeer?.pubkey} name={selectedPeer?.name} />
```
rolznz, line 217: Can these new alerts be moved down to just before the confirm/next button in both flows? you will see other alerts there too.

As it is now it looks weird at the top and also is way too wide

rolznz: For <IncreaseIncomingCapacity/> it might only apply if the user picks the advanced option. But I think we need to calculate the channel upfront if advanced is not used, so we can check the automatic flow too.

im-adithya: > For IncreaseIncomingCapacity it might only apply if the user picks the advanced option

Ah true, added now!

> But I think we need to calculate the channel upfront if advanced is not used, so we can check the automatic flow too

But we already check and open only if there isn't no channel with the counterparty here: 
https://github.com/getAlby/hub/blob/85d89bf93af73d0b3024caf53e94d0492ca99ef4/frontend/src/screens/channels/IncreaseIncomingCapacity.tsx#L161-L162

and we then show a toast saying the user should choose one themselves from advanced options (if there isn't any partner without a channel) So the user will then see the alert on clicking advanced options

rolznz: Ah, great!

im-adithya: Also pushed alerts to bottom ✅

rolznz (approved):
tACK

PR #1052, opened 2025-01-31
feat: lightning messageboard widget

rolznz:
image

image


rolznz:
There might be a lot of requests because there will be N per user each time they visit the homepage (N pages of transactions).

rolznz:
@reneaaron thanks for the feedback. The message board no longer loads by default (this is not optimal, but is a prevention mechanism against too many requests).

Also, to post the minimum is now 1000 sats. This ensures that the messages being retrieved are economical and not useless 1-sat messages.

PR #1054, opened 2025-01-31
fix: conflict warning is not removed on switching

### LINKED ISSUE #1053: Public channel checkbox is not unchecked on switching to a partner which only offers private channels
## Description
If you click on public channel checkbox with all private channels you'd see a warning saying there's a conflict. And then when you shift to a channel partner where public channels are not even allowed (Olympus), the warning still stays as the checkbox is still checked (but is disabled).

Screenshot 2025-01-31 at 4 32 18 PM

And on clicking "Next" you see this error:

Image

im-adithya:
Fixes #1053


rolznz (approved):
utACK

im-adithya:
Don't merge this, have to check it once

PR #1055, opened 2025-01-31
feat: use DATABASE_URL if present

frnandu:
If DATABASE_URL env variable is set, it will overwrite DATABASE_URI with that value
This allows to use the secret that fly sets on the fly app when attaching a postgres instance which also creates the user and database.


frnandu:
not needed, since attach method has a `--variable-name` option

PR #1056, opened 2025-01-31
feat: alby lite app

im-adithya:
Screenshot 2025-01-31 at 9 12 33 PM

Screenshot 2025-01-31 at 9 13 04 PM


rolznz:
Some thoughts + ones we talked about yesterday:
- After our conversation about NIP-05 not being a good way to contact the user, I am wondering if this should be optional, as it's only really needed if you specifically want a albylite NIP-05 address.
- This now "competes" with the Friends & Family app (which do we choose?). Would it make sense to move it there as an optional step after the account is created?
- We have the issue that sub-wallets have spending permission. But we have not found a graceful way to handle this. Some of us think Alby Lite should not have spending access to wallets. It should have receive-only access.
- From Rene: "Wasn't the main problem with Alby Lite that we didn't have a good answer for how the customer journey would look like for these people and it would make the existing product landscape even more complicated?" - we plan to have a call about this next week.
- I like the idea about paying up-front especially if we find a way to do receive-only connections for alby lite.

rolznz:
Marking as draft for now, since we plan to not host alby lite

PR #1066, opened 2025-02-05
fix: improve recently used apps card

reneaaron:
- Text on cards was bigger than the heading, which disturbs the hierarchy
 - Filter apps without any events


rolznz (approved):
utACK

Issue #1065, opened 2025-02-05
Set PerChannelReserveSats rather than TrustedPeersNoReserve


rolznz:
After a channel is closed with stuck HTLCs, Alby Hub will not try to claim contestable funds since it trusts the LSP. This means we need to actively contact the LSP for a refund.

We can do a middle ground by instead of keeping a list of TrustedPeersNoReserve we can set the PerChannelReserveSats to 0. This should still allow channels to be made with 0 on-chain balance, but if the user does have some on-chain balance, they can spend it to ensure their funds are returned to them. It's not perfect because the user might not have on-chain funds. But it should be a bit better than now.

PR #1069, opened 2025-02-06
fix: replace LDK trusted 0-anchor reserve peers with 0 per channel reserve

rolznz:
Closes https://github.com/getAlby/hub/issues/1065

This PR stops requiring all users to trust the LSPs to correctly return the correct amount on channel closure (which when channels are closed due to stuck payments they do not), and instead allow the user to choose whether or not they will protect themselves or trust the counterparty. It introduces one extra warning icon to the UI but I think this is important for users to understand.

Tests in mutinynet:
- [x] Existing channels are unaffected
- [x] New channels with no on-chain balance are still opened successfully.

image

image

image

### DIFF
--- a/frontend/src/screens/channels/Channels.tsx
+++ b/frontend/src/screens/channels/Channels.tsx
@@ -594,7 +594,7 @@ export default function Channels() {
                       Your on-chain balance can be used to open new outgoing
                       lightning channels and to ensure channels can be closed
                       when required. When channels are closed, funds on your
-                      side of your channel will be returned to your savings
+                      side of your channel will be returned to your on-chain
                       balance.
                     </TooltipContent>
                   </Tooltip>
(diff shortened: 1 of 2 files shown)


frontend/src/screens/channels/Channels.tsx
```diff
@@ -592,8 +592,9 @@ export default function Channels() {
                     </TooltipTrigger>
                     <TooltipContent className="w-[300px]">
                       Your on-chain balance can be used to open new outgoing
-                      lightning channels. When channels are closed, funds on
-                      your side of your channel will be returned to your savings
+                      lightning channels and to ensure channels can be closed
+                      when required. When channels are closed, funds on your
+                      side of your channel will be returned to your savings
```
MoritzKa, line 597: ... wil be returned to your on-chain balance.

frontend/src/screens/channels/Channels.tsx
```diff
@@ -610,11 +611,31 @@ export default function Channels() {
...
+                          <Tooltip>
+                            <TooltipTrigger>
+                              <AlertTriangle className="w-3 h-3 text-muted-foreground" />
+                            </TooltipTrigger>
+                            <TooltipContent className="w-[300px]">
+                              You have insufficient funds in reserve to close
+                              channels or bump on-chain transactions and
+                              currently rely on the counterparty. It is
+                              recommended to deposit at least 25,000 sats
+                              on-chain.
```
MoritzKa, line 634: .... to deposit at least 25,000 sats to your on-chain balance.

Issue #1071, opened 2025-02-06
(v1.14.0) Opening my.albyhub.com URL directly returns JWT error


rolznz:
e.g. open https://my.albyhub.com/apps/new

`{"message":"missing or malformed jwt"}`

PR #1074, opened 2025-02-06
fix: only use JWT middleware on /api paths

rolznz:
Fixes https://github.com/getAlby/hub/issues/1071 by only applying the JWT config to the `/api` prefix.

I do not understand why this works, as this should mean the non-restricted API endpoints would not work, but they still do. This needs to be understood properly before merging.

Also moves a few endpoints into the restricted group (they were still protected, but it's better for them to be protected by the JWT middleware)

```
restrictedApiGroup := e.Group("/api")
restrictedApiGroup.Use(echojwt.WithConfig(jwtConfig))
```


im-adithya (approved):
LGTM

rolznz:
As far as I can tell this is due to buggy handling of groups when the path is set to "", combined with how the static middleware handles not found routes

PR #1077, opened 2025-02-10
fix: offline channel detection in health endpoint

im-adithya:
When a channel is being opened, health endpoint showsss red saying one or more channels are offline, which is wrong.

Also fixes the wrong calculation of Channel Confs in LND (we don't count the funding tx block currently)

## Screenshots
Screenshot 2025-02-10 at 7 27 39 PM
Screenshot 2025-02-10 at 7 30 35 PM


rolznz (approved):
utACK

PR #1078, opened 2025-02-10
fix: show total channels and peers

### LINKED ISSUE #1070: Bug: channel count & peers in stats for nerds show only online, not total
It should say 6/7 and 11/13

Image

im-adithya:
Fixes #1070

In LND, ListPeers only returns the list of connected peers so this fix will not make a difference there, not sure about LDK.
Channels should be fixed though.

## Screenshots
Screenshot 2025-02-10 at 7 57 52 PM

### On switching off a node in Polar: (Notice Peers)
Screenshot 2025-02-10 at 7 57 16 PM

### DIFF
--- a/frontend/src/components/home/widgets/NodeStatusWidget.tsx
+++ b/frontend/src/components/home/widgets/NodeStatusWidget.tsx
@@ -25,12 +25,12 @@ export function NodeStatusWidget() {
           <p className="text-muted-foreground text-xs">Channels Online</p>
           <p className="text-xl font-semibold">
             {channels.filter((c) => c.active).length || 0} /{" "}
-            {channels.filter((c) => c.active).length || 0}
+            {channels.length || 0}
           </p>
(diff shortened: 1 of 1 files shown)


rolznz (approved):
utACK

im-adithya:
Mentioned in the PR comment, `ListPeers` in LND only returns list of active connected peers

rolznz:
@im-adithya sorry, I read your comment. That's weird in LND. Is there a way to get offline peers? it would be good to show those. LDK shows offline peers.

im-adithya:
I checked the LND codebase as well, sadly we can't fetch the offline peers. But one thing we can do is to calculate those from the offline channels if any and add them up, but I feel it might be unnecessary work

rolznz:
@im-adithya that won't be perfect either since you can have peers with no channels. I guess it's ok for now. Thanks!

Issue #1076, opened 2025-02-07
Bug: Health check always shows cannot connect to relay


rolznz:
![Image](https://github.com/user-attachments/assets/eeb5968a-9820-4805-8be8-9dc6cf15e340)

im-adithya:
Couldn't reproduce this issue, maybe the relay was actually down? Were you using albylabs relay?

rolznz:
@im-adithya I found the issue, it can only be reproduced with legacy apps. I've made a PR

PR #1085, opened 2025-02-12
fix: relay always offline for hubs with legacy subscription

rolznz:
Fixes https://github.com/getAlby/hub/issues/1076

Marking the relay as online was never reached if we started the legacy app wallet subscription since it's a blocking method.

The start app wallet subscription method has been moved to a goroutine to not block the loop.

### DIFF
--- a/service/start.go
+++ b/service/start.go
@@ -115,21 +115,22 @@ func (svc *service) startNostr(ctx context.Context) error {
 				return
 			}
 			if legacyAppCount > 0 {
-				// re-publish single NIP47 event info for legacy apps
-				_, err := svc.GetNip47Service().PublishNip47Info(ctx, relay, svc.keys.GetNostrPublicKey(), svc.keys.GetNostrSecretKey(), svc.lnClient)
-				if err != nil {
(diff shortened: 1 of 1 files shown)


service/start.go
```diff
@@ -115,21 +115,22 @@ func (svc *service) startNostr(ctx context.Context) error {
...
+					}
+					logger.Logger.WithField("legacy_app_count", legacyAppCount).Info("Starting legacy app subscription")
+					// legacy single wallet subscription - only subscribe once for all legacy apps
+					// to ensure we do not get duplicate events
+					err = svc.startAppWalletSubscription(ctx, relay, svc.keys.GetNostrPublicKey())
+					if err != nil && !errors.Is(err, context.Canceled) {
+						// err being non-nil means that we have an error on the websocket error channel. In this case we just try to reconnect.
+						logger.Logger.WithError(err).Error("Got an error from the relay while listening to legacy subscription.")
+					}
+				}()
```
im-adithya, line 133: Ah, nice catch 🙌

Issue #1026, opened 2025-01-27
Bug: Missing node name/ID in pending channel closure alert


rolznz:
![Image](https://github.com/user-attachments/assets/15e12a46-1cf8-4075-9924-679b797a606a)

im-adithya:
I think this happens when there is a `PendingSweepBalance` because then we don't add the `pendingBalancesDetails`

https://github.com/getAlby/hub/blob/73540ef38b0c16d717e789482a405221004d3f07/lnclient/ldk/ldk.go#L1066-L1075

I think it's best to skip when `pendingBalancesDetails` is empty since by then the channel is properly closed and is not listed in ListChannels response so we don't even fetch the remote node's info to display the alias (and we only have the channel id (see [PendingSweepBalance docs](https://docs.rs/ldk-node/latest/ldk_node/enum.PendingSweepBalance.html)) so I'm not sure if we can even fetch the node pubkey using it)

rolznz:
@im-adithya thanks for looking into it! I don't agree though - if the funds are not swept yet then there are no incoming funds that we can show next to the onchain balance. `ListChannels` _only_ returns channels which have no closure initiated yet.

@rdmitr are you able to see if it's possible to include some additional fields in `balances.PendingBalancesFromChannelClosures` such as the `CounterpartyNodeId`, `FundingTxId` and `FundingTxIndex` similar to what you did in https://github.com/getAlby/ldk-node/pull/60? in Alby Hub as Adithya linked above, we handle those ones differently.

PR #1087, opened 2025-02-12
feat: return pending sweep balance details in LDK

rdmitr:
Fixes https://github.com/getAlby/hub/issues/1026

### DIFF
--- a/lnclient/ldk/ldk.go
+++ b/lnclient/ldk/ldk.go
@@ -1071,15 +1071,27 @@ func (ls *LDKService) GetOnchainBalance(ctx context.Context) (*lnclient.OnchainB
 		}
 	}
 
+	pendingSweepBalanceDetails := make([]lnclient.PendingBalanceDetails, 0)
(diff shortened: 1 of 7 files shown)


lnclient/ldk/ldk.go
```diff
@@ -1071,15 +1071,27 @@ func (ls *LDKService) GetOnchainBalance(ctx context.Context) (*lnclient.OnchainB
 		}
 	}
 
+	pendingSweepBalanceDetails := make([]lnclient.PendingBalanceDetails, 0)
+	increasePendingBalanceFromClosure := func(nodeId, channelId *string, amount uint64, fundingTxId *ldk_node.Txid, fundingTxIndex *uint16) {
+		pendingBalancesFromChannelClosures += amount
+		pendingSweepBalanceDetails = append(pendingSweepBalanceDetails, lnclient.PendingBalanceDetails{
+			NodeId:        utils.UnwrapOrZero(nodeId),
```
rolznz, line 1078: If we do it this way we can add an invalid `lnclient.PendingBalanceDetails`. Can we instead check all the properties are not nil, and only if so add the element to the array?

rolznz (approved):
tACK

PR #1088, opened 2025-02-12
fix: avoid empty currency selector initially

pavanjoshi914:
before: 
image

after: 
image

### DIFF
--- a/frontend/src/screens/settings/Settings.tsx
+++ b/frontend/src/screens/settings/Settings.tsx
@@ -38,11 +38,9 @@ function Settings() {
           ([code, details]) => [code.toUpperCase(), details.name]
         );
 
-        const filteredCurrencies = mappedCurrencies
-          .sort((a, b) => a[1].localeCompare(b[1]))
-          .filter((a) => a[0] !== "SATS");
+        mappedCurrencies.sort((a, b) => a[1].localeCompare(b[1]));
 
-        setFiatCurrencies(filteredCurrencies);
+        setFiatCurrencies(mappedCurrencies);
       } catch (error) {
         console.error(error);
         handleRequestError(toast, "Failed to fetch currencies", error);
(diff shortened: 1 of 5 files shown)


frontend/src/screens/settings/Settings.tsx
```diff
@@ -28,7 +28,7 @@ function Settings() {
   const { data: info } = useInfo();
 
   const [selectedCurrency, setSelectedCurrency] = useState<string | undefined>(
-    info?.currency
+    info?.currency.toUpperCase()
```
rolznz, line 31: can you fix the root cause? isn't it because the default currency set in the backend is `usd` instead of `USD`?

pavanjoshi914: resolved

rolznz: Is this change still needed then?

wails/wails_handlers.go
```diff
@@ -374,7 +374,7 @@ func (app *WailsApp) WailsRequestRouter(route string, method string, body string
 				"route":    route,
 				"method":   method,
 				"body":     body,
-				"currency": "usd",
+				"currency": app.svc.GetConfig().GetCurrency(),
```
rolznz, line 377: is this necessary?

pavanjoshi914: if we fail to get bitcoin rate logs shall show for which currency it failed. no?

frontend/src/screens/settings/Settings.tsx
pavanjoshi914: removed

wails/wails_handlers.go
rolznz: Why is it only logged in the wails handler? I don't see this in the http version despite it being much more used.

Also, I think it's wrong to log the currency at this level. I think it's better to include that in the errors logged at `GetBitcoinRate` in the OAuth service

pavanjoshi914: resolved

rolznz (approved):
tACK

PR #1089, opened 2025-02-12
feat: add peer and return_to query parameters to peer connection page

danielbroll:
## Why

Instructing a user to copy and paste connection strings to establish a connection (or channel) is tedious and error-prone. Allowing a service to pass the connection string and a return URL as query parameters simplifies this experience.

## How

Similar to the new app page, a redirect URL (`return_to`) and the connection string (`peer`) can now be passed as query parameters to the new peer page.

Example: https://my.albyhub.com/#/peers/new?peer=03b65c19de2da9d35895b37e6fa263cefbf8d184e6b61c0713747ebf12409d219f%40164.90.241.111%3A9735&return_to=https%3A%2F%2Fwebhook.site%2Fa2a6fd93-0f48-4ef9-8b74-1baf201d4cae

grafik

### DIFF
--- a/frontend/src/screens/peers/ConnectPeer.tsx
+++ b/frontend/src/screens/peers/ConnectPeer.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { useNavigate } from "react-router-dom";
+import { useLocation, useNavigate } from "react-router-dom";
 import AppHeader from "src/components/AppHeader";
 import { Input } from "src/components/ui/input";
 import { Label } from "src/components/ui/label";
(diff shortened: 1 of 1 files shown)


rolznz:
@danielbroll thanks for the PR.

But what service would use this? how many users will this apply to?

Normally managing peers is for advanced users already. Standard users would just use an automatically suggested channel partner and we auto-peer. The user does not need to know about peers by default.

frontend/src/screens/peers/ConnectPeer.tsx
```diff
@@ -44,15 +49,19 @@ export default function ConnectPeer() {
...
+        setLoading(false);
+      }
     } catch (e) {
       toast({
         variant: "destructive",
         title: "Failed to connect peer: " + e,
       });
       console.error(e);
-    } finally {
       setLoading(false);
```
im-adithya, line 65: nit: I think the `finally` can stay instead of repeating twice

rolznz (approved):
tACK, addressed @im-adithya 's feedback

Issue #1073, opened 2025-02-06
Hubs with VSS sometimes take a long time to start


rolznz:
This causes the timeout error to show, which is not correct.

The simplest way would be to increase the timeout, I am not sure if there is something better we can do...?

PR #1090, opened 2025-02-13
feat: replace start timeout with hub startup state

rolznz:
Fixes https://github.com/getAlby/hub/issues/1073

Originally we added an arbitrary timeout and delayed messages to give the user the feeling that something is happening.

The problem with the timeout is even after the timeout we cannot do anything because we lock the startup and it's impossible to try again until the startup definitely succeeds or fails.

I have replaced the timeout with showing the actual current state of the node startup. (Only on the start screen - the finish screen for first node startup is fast enough)


frontend/src/screens/Start.tsx
```diff
@@ -129,6 +94,11 @@ export default function Start() {
               <LoadingButton type="submit" loading={loading}>
```
im-adithya, line 94: Should we then also change the title of the `TwoColumnLayoutHeader` to Start as well?

im-adithya (approved):
Just a small comment, otherwise tACK

Issue #1101, opened 2025-02-18
Wails hub fails to pay invoice


rolznz:
In two places in the code (LightningMessageBoardWidget, CurrentChannelOrder) we do not include a body for the POST request to `/api/payments/` which the wails handler expects a body. Either we should always pass a body or update the wails handler to check if there is a body.

```
{"body":"","error":"unexpected end of JSON input","level":"error","method":"POST","msg":"Failed to decode request to wails router","route":"/api/payments/lnbc10u1pnmgs0rdqgfpjhj7tenp4qdkt5m0tdusqvzv7ufx9nypz0aq3nt5ktpv3645cq9y6azkv3hw86pp56yx98a4cgqzzl58h8nvzsem3hqu32uf28jxgw3ut2969y3r2gq3ssp5rkexw27jql82jg9ahghnw2636gmezju0d788q797yx5xwj77aqps9qyysgqcqpcxqyz5vqrzjqw9fu4j39mycmg440ztkraa03u5qhtuc5zfgydsv6ml38qd4azymlapyqqqqqqq905qqqqlgqqqq86qqjqrzjqwryaup9lh50kkranzgcdnn2fgvx390wgj5jd07rwr3vxeje0glc7qnukmlyq2ylk5qqqqlgqqqqqeqqjqrzjq26922n6s5n5undqrf78rjjhgpcczafws45tx8237y7pzx3fg8wwxrthryqqvfcqqyqqqqqqqqqqqqqq2q34v62asrpvfzgnr42sgmv2puxzp9rumwn3k8hp5trg038am5sarxuzuh40gqh80q2x2tcv8w5820vx0eeu9ftr5t7jf877a250vw42qp7ry24q","time":"2025-02-18T13:46:36+05:30"}
```

PR #1102, opened 2025-02-18
fix: make body optional for payments endpoint in wails handler

rolznz:
Fixes https://github.com/getAlby/hub/issues/1101

I accidentally broke this when adding support to pay 0-amount invoices


im-adithya (approved):
tACK

PR #1105, opened 2025-02-18
chore: show total and reserved lightning balances in LND

im-adithya:
This helps reduce confusion in LND users as we can now point people to check this if people raise support requests titled "Sats are missing/not showing in Alby Hub"

## Screenshot
Screenshot 2025-02-18 at 10 07 26 PM

### DIFF
--- a/frontend/src/screens/channels/Channels.tsx
+++ b/frontend/src/screens/channels/Channels.tsx
@@ -70,6 +70,7 @@ import {
 import { useAlbyBalance } from "src/hooks/useAlbyBalance.ts";
 import { useBalances } from "src/hooks/useBalances.ts";
 import { useChannels } from "src/hooks/useChannels";
+import { useInfo } from "src/hooks/useInfo";
 import { useIsDesktop } from "src/hooks/useMediaQuery.ts";
 import { useNodeConnectionInfo } from "src/hooks/useNodeConnectionInfo.ts";
 import { useOnchainAddress } from "src/hooks/useOnchainAddress";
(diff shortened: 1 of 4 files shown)


frontend/src/screens/channels/Channels.tsx
```diff
@@ -506,8 +508,18 @@ export default function Channels() {
                         </div>
                       </TooltipTrigger>
                       <TooltipContent className="w-[300px]">
-                        Your spending balance is the funds on your side of your
-                        channels, which you can use to make lightning payments.
+                        Your spending balance is the spendable funds on your
+                        side of your channels, which you can use to make
+                        lightning payments.{" "}
+                        {info?.backendType === "LND" &&
```
rolznz, line 514: I am wondering if we should always show this info, then users running LDK also can see the reserved amount and total on this tooltip?

im-adithya: I just discovered that we don't remove the reserves in LDK.

See `LocalBalance` calculation:
https://github.com/getAlby/hub/blob/eaf72825c0001dfacab97d7bcb22dffa122dfd82/lnclient/ldk/ldk.go#L858

And see `totalSpendable` calculation:
https://github.com/getAlby/hub/blob/eaf72825c0001dfacab97d7bcb22dffa122dfd82/lnclient/ldk/ldk.go#L1638

(Similarly we don't exclude the reserves in `totalReceivable` as well)

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/1161

frontend/src/screens/channels/Channels.tsx
rolznz: `LocalBalance` should include the reserve based on the current logic. `LocalSpendableBalance` does not have the reserves.

PR #1106, opened 2025-02-18
fix: remove lnd certificate from user config if not provided

### LINKED ISSUE #405: Bug: Self-signed LND certificates error in Windows Desktop

im-adithya:
Fixes #405

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -72,7 +72,8 @@ func (cfg *config) init(env *AppConfig) error {
 			return err
 		}
 	} else {
-		// If no LNDCertFile is provided, we create a new one during startup
+		// If no LNDCertFile is provided, clear any stored certificate
+		// hex value so that no certificate is used for TLS verification.
 		err := cfg.SetUpdate("LNDCertHex", "", "")
 		if err != nil {
 			return err
--- a/config/config.go
+++ b/config/config.go
@@ -71,6 +71,12 @@ func (cfg *config) init(env *AppConfig) error {
 		if err != nil {
 			return err
 		}
+	} else {
+		// If no LNDCertFile is provided, we create a new one during startup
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -71,6 +71,12 @@ func (cfg *config) init(env *AppConfig) error {
 		if err != nil {
 			return err
 		}
+	} else {
+		// If no LNDCertFile is provided, we create a new one during startup
```
rolznz, line 75: what do you mean by "create a new one"? this comment seems incorrect to me

rolznz (approved):
utACK (once comment is fixed)

config/config.go
rolznz: This is a bit ungraceful - now all users even ones who do not run LND will have this entry in their user config

rolznz: should we only run that code if `cfg.Env.LNBackendType` is LND?

im-adithya: I think the comment is misleading, it just skips the tls verification if no certificate is provided. And for that to happen, we need to leave this empty. So other backends should be fine.

rolznz (approved):
tACK

Issue #1108, opened 2025-02-19
Inconsistency: Deeplink max_amount field is in sats, not millisats


rolznz:
The unit for NWC is millisats, but for some reason the budget is specified in sats. As we hope to have a spec around this it should be fixed before there is usage.

PR #1110, opened 2025-02-19
fix: use msats for deeplink max_amount param to be consistent with NIP-47

rolznz:
Fixes https://github.com/getAlby/hub/issues/1108

Minimal change to consistently use msat values. The deep link flow changes were a breaking change and hopefully a spec can be created around this for http-accessible wallets, so it's good to fix this inconsistency now.

image


im-adithya (approved):
tACK

PR #1111, opened 2025-02-19
fix: convert budget from msats to sats

im-adithya:
Also converts budget renewal to lowercase before checking

### DIFF
--- a/nip47/controllers/create_connection_controller.go
+++ b/nip47/controllers/create_connection_controller.go
@@ -53,6 +53,8 @@ func (controller *nip47Controller) HandleCreateConnectionEvent(ctx context.Conte
 		expiresAt = &expiresAtValue
 	}
 
+	maxAmountSat := params.Budget.Budget / 1000
+
 	// explicitly do not allow creating an app with create_connection permission
 	if slices.Contains(params.Methods, models.CREATE_CONNECTION_METHOD) {
 		publishResponse(&models.Response{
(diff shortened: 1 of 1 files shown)


apps/apps_service.go
```diff
@@ -47,6 +47,8 @@ func (svc *appsService) CreateApp(name string, pubkey string, maxAmountSat uint6
 
 	if budgetRenewal == "" {
 		budgetRenewal = constants.BUDGET_RENEWAL_NEVER
+	} else {
+		budgetRenewal = strings.ToLower(budgetRenewal)
```
rolznz, line 51: This doesn't follow the spec, I would rather throw an error which we do below

nip47/controllers/create_connection_controller.go
```diff
@@ -53,6 +53,8 @@ func (controller *nip47Controller) HandleCreateConnectionEvent(ctx context.Conte
 		expiresAt = &expiresAtValue
 	}
 
+	maxAmountSat := uint64(params.Budget.Budget / 1000)
```
rolznz, line 56: why is casting needed here?

im-adithya: Ah I thought you have to do that while dividing, will remove :+1:

apps/apps_service.go
im-adithya: Hmmm yeah I think it should be debuggable from the error, removing this :+1:

nip47/controllers/create_connection_controller.go
rolznz: I think go is different and will preserve the current type. The type is already a uint64 right? so it will do truncation instead

im-adithya: Yeah that's good

rolznz (approved):
utACK

PR #1116, opened 2025-02-21
Feat: update to gonuts 0.4.0

rolznz:
@elnosh any chance you could review this? I found a couple issues with the fees.


elnosh:
yeah fee from melt will be in sats. One note, there may be a discrepancy in the fee calculated in `SendPaymentSync` vs `cashuMeltQuoteToTransaction` in the case where there is change returned. Fee calculated in `SendPaymentSync` takes the change into account (if there was any) for the fee. It's not stored in the melt quote tho so it is not taken into account in `cashuMeltQuoteToTransaction`. So if there was change, those values will differ.

rolznz:
@elnosh I think that's ok, Normally the fee should be saved from `SendPaymentSync`. Thanks for checking!

PR #1119, opened 2025-02-23
refactor: use a more straightforward return value

beforetech:


beforetech:
Thanks for your review.

Issue #1123, opened 2025-02-24
Bug: chance of crash in shutdown if node is syncing


rolznz:
When the LDK node `Shutdown` method is called we set `ls.node` to nil to avoid any other access to it. https://github.com/getAlby/hub/blob/master/lnclient/ldk/ldk.go#L408-L418

The problem is we still access the node in some places by the LDK Node LNClient itself https://github.com/getAlby/hub/blob/master/lnclient/ldk/ldk.go#L325-L394

I don't think this should actually be set to nil. We already set the lnClient to nil higher above, which should be sufficient: https://github.com/getAlby/hub/blob/master/service/stop.go#L24 (from the history this was added 3 months later)

PR #1124, opened 2025-02-24
fix: LDK panic on shutdown due to nil pointer error when deleting old payments

rolznz:
Fixes https://github.com/getAlby/hub/issues/1123

rather than setting the node to nil on the lnclient which caused a nil pointer exception if it is accessed while shutting down (which happened when deleting old payments after the node sync finished), use a mutex and shuttingDown boolean to ensure shutdown can only be called once.

### DIFF
--- a/lnclient/ldk/ldk.go
+++ b/lnclient/ldk/ldk.go
@@ -14,6 +14,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/getAlby/ldk-node-go/ldk_node"
@@ -47,6 +48,7 @@ type LDKService struct {
 	cfg                   config.Config
 	lastWalletSyncRequest time.Time
 	pubkey                string
+	shuttingDown          bool
 }
 
 const resetRouterKey = "ResetRouter"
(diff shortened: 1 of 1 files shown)


lnclient/ldk/ldk.go
```diff
@@ -399,14 +401,16 @@ func NewLDKService(ctx context.Context, cfg config.Config, eventPublisher events
 	return &ls, nil
 }
 
+var shutdownMutex sync.Mutex
+
 func (ls *LDKService) Shutdown() error {
-	if ls.node == nil {
-		logger.Logger.Debug("LDK client already shut down")
+	shutdownMutex.Lock()
+	defer shutdownMutex.Unlock()
```
im-adithya, line 408: Why both `mutex` and `bool`?

lnclient/ldk/ldk.go
```diff
@@ -436,7 +440,7 @@ func (ls *LDKService) Shutdown() error {
 	}
 
 	logger.Logger.Debug("Destroying LDK node object")
-	node.Destroy()
+	ls.node.Destroy()
```
im-adithya, line 444: Should we here set `shuttingDown` to false again? So we prevent shutdown being blocked on restart/2nd attempt?

rolznz: It's an instance variable, so a new LDKNode will have `shuttingDown` as false

im-adithya: Ah true :+1:

PR #1127, opened 2025-02-25
fix: skeleton component

reneaaron:
- Fix layout shifts due to different heights
 - Use shadcn Skeleton component
 - Make loading animation work for different text sizes
 - Some more responsive improvements


im-adithya (approved):
tACK, nice 🙌

PR #1129, opened 2025-02-25
fix: transaction list scrolling on visiting wallet page

im-adithya:
This fixes wallet page (and app connection page) scrolling to the transaction list on every visit.


reneaaron (approved):
utACK

PR #1138, opened 2025-03-01
feat: add app cleanup page

rolznz:
A satisfying way to delete apps with one click. Button only shows when the user has more than 5 connections, apps are ordered by unused and then oldest first. Only apps used more than a month ago are considered.

I have many apps I didn't clean up, leading to poor performance especially with firing unnecessary notifications.

image

image

image

### DIFF
--- a/frontend/src/components/connections/AppCard.tsx
+++ b/frontend/src/components/connections/AppCard.tsx
@@ -17,9 +17,10 @@ dayjs.extend(relativeTime);
 type Props = {
   app: App;
   actions?: React.ReactNode;
+  readonly?: boolean;
 };
 
-export default function AppCard({ app, actions }: Props) {
+export default function AppCard({ app, actions, readonly = false }: Props) {
   const navigate = useNavigate();
 
   return (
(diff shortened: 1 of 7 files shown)


im-adithya (approved):
tACK, this is cool 👌 

I'm just worried about the distance between Keep and Delete (accidental clicks) but these are unused apps and connection secret might've been lost in most cases so it's fine

reneaaron:
> Right now though F&F connections are not filtered out - so they probably would need to be in that case

:+1:, that definitely makes sense.

> @reneaaron Yeah, I could extract the code that filters the apps, so that we can check that rather than showing the button if there are more than 5 apps

So you'd only display the button if there are more than 5 "unused" apps then?

rolznz:
> So you'd only display the button if there are more than 5 "unused" apps then?

@reneaaron no, I mean if there is at least 1 "unused" app, then show this button on the connections page. But then it becomes more complicated when we consider what is an "unused" app - but for this reason I would only consider the last used date, not the type of app (whether it is a F&F app or some other app).

rolznz:
@reneaaron I added a new hook to return unused apps and use that instead to display the button on the apps list page. I also changed the "unused" date to 2 months

reneaaron (approved):
tACK

PR #1145, opened 2025-03-03
fix: close swap dialogs after opening new window

reneaaron:
Currently you have to first manually close the dialog before you can e.g. pay the invoice from this flow.


rolznz (approved):
utACK

PR #1146, opened 2025-03-04
chore: show wallet pubkey on connection page

rolznz:
We were not showing the wallet service pubkey anywhere which is useful info. I also added an alert to show if the app is a legacy one or not. It's not a critical issue, but I think it's nice to know and be able to see.

image

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -284,20 +284,29 @@ func (api *api) GetApp(dbApp *db.App) *App {
 		}
 	}
 
+	walletPubkey := api.keys.GetNostrPublicKey()
+	uniqueWalletPubkey := false
+	if dbApp.WalletPubkey != nil {
+		walletPubkey = *dbApp.WalletPubkey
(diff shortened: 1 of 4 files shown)


reneaaron (approved):
I think for the average user most of the "connection info" is not important at all and confuses more than it helps, but we'll tackle that in a different issue. 

What are the privacy implications of this shared pubkey exactly?

rolznz:
> What are the privacy implications of this shared pubkey exactly?

One client/app which knows the wallet pubkey can see activity across all clients that use the same pubkey. Although it's E2E encrypted, it still leaks metadata about what a single user has been doing. @frnandu was there another reason?

reneaaron:
Okay, but for me that's not enough to justify urging users to re-create the connection, I think it's fine the way you implemented it :+1:

rolznz:
@matjaz unfortunately that would be a breaking change (the clientside will not know that the pubkey has changed)

PR #1152, opened 2025-03-05
fix: channels sorting on node page

rolznz:
Fixes https://github.com/getAlby/hub/issues/1063

consistently sorts channels - channel capacity first, then spending balance second

image


reneaaron (approved):
tACK

PR #1153, opened 2025-03-06
feat: alby pro subscriptions

reneaaron:


frontend/src/screens/Home.tsx
```diff
@@ -59,7 +60,15 @@ function Home() {
 
   return (
     <>
-      <AppHeader title={getGreeting(albyMe?.name)} description="" />
+      <AppHeader
+        title={getGreeting(albyMe?.name)}
+        description=""
+        contentRight={
+          <UpgradeDialog>
+            <Button variant="premium">Upgrade</Button>
```
rolznz, line 68: image

Does the upgrade button feel strange here? I wonder if later we could move it to a card on the home page to explain the benefits that users can see before they click?

frontend/src/routes.tsx
```diff
@@ -19,6 +19,7 @@ import Start from "src/screens/Start";
 import Unlock from "src/screens/Unlock";
 import { Welcome } from "src/screens/Welcome";
 import AlbyAuthRedirect from "src/screens/alby/AlbyAuthRedirect";
+import Support from "src/screens/alby/Support";
```
rolznz, line 22: It looks like this is not linked from anywhere. Do we still plan to use it?

frontend/src/components/UpgradeDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+                    Keep your node running smoothly
+                  </p>
+                </div>
+              </li>
+              */}
+            </ul>
+          </div>
+        </div>
+        <div className="flex-col gap-1">
+          <Button
```
rolznz, line 107: Can we use the `ExternalLinkButton` component instead? buttons that are not proper links are bad UX

frontend/src/components/UpgradeDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+} from "src/components/ui/dialog";
+import { ALBY_PRO_PLAN } from "src/constants";
+import { useAlbyMe } from "src/hooks/useAlbyMe";
+import { openLink } from "src/utils/openLink";
+
+interface UpgradeDialogProps {
+  children: ReactNode;
+}
+
+export const UpgradeDialog = ({ children }: UpgradeDialogProps) => {
```
rolznz, line 20: As a follow up, I think we could consider Alby Cloud too. Right now it's hardcoded to the pro plan.

We have a menu item called "Alby Cloud" on the sidebar when the user is logged in but does not have a plan - should we change anything here? why does it only show when logged in for example? (I think this no longer applies?)

frontend/src/constants.ts
```diff
@@ -13,3 +13,5 @@ export const LIST_TRANSACTIONS_LIMIT = 20;
 
 export const SUPPORT_ALBY_CONNECTION_NAME = `ZapPlanner - Alby Hub`;
 export const SUPPORT_ALBY_LIGHTNING_ADDRESS = "hub@getalby.com";
+
+export const ALBY_PRO_PLAN = `pro_202411_usd_year`;
```
rolznz, line 17: Is it too late to change the plan name? e.g. include `diy_pro_202411_usd_year`.

frontend/src/components/UpgradeDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+                    Share with friends, family, coworkers
+                  </p>
+                </div>
+              </li>
+              <li className="flex items-center">
+                <RefreshCw className="w-5 h-5 mr-3 text-primary" />
+                <div>
+                  <span className="font-medium">Encrypted Remote Backups</span>
+                  <p className="text-sm text-muted-foreground">
+                    Secure LDK wallet backups in the cloud
```
rolznz, line 61: I think we should only show this for LDK

frontend/src/components/UpgradeDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+export const UpgradeDialog = ({ children }: UpgradeDialogProps) => {
+  const { data: albyMe } = useAlbyMe();
+
+  const handleUpgrade = () => {
+    openLink(
+      `https://www.getalby.com/subscription/confirm?plan=${ALBY_PRO_PLAN}`
+    );
+  };
+
+  if (albyMe?.subscription.buzz) {
```
rolznz, line 29: to check if we should use `buzz` or some other property instead like if the plan code exists

frontend/src/screens/Home.tsx
rolznz: also that removes duplication on other pages (e.g. on the alby jim page the button is shown twice)

frontend/src/components/UpgradeDialog.tsx
reneaaron: Good point, we decided to have a general "upgrade" button in there for now and allow users to decide between diy + cloud themselves for now. (since for desktop users a hosted option might also be relevant)

frontend/src/components/UpgradeDialog.tsx
```diff
@@ -0,0 +1,112 @@
+import { GemIcon, LifeBuoy, Mail, RefreshCw, Users, Zap } from "lucide-react";
```
rolznz, line 1: We will need to remove the old SupportAlby flow right?

frontend/src/components/UpgradeDialog.tsx
```diff
@@ -0,0 +1,112 @@
...
+              <li className="flex items-center">
+                <Users className="w-5 h-5 mr-3 text-primary" />
+                <div>
+                  <span className="font-medium">Unlimited Sub-wallets</span>
+                  <p className="text-sm text-muted-foreground">
+                    Share with friends, family, coworkers
+                  </p>
+                </div>
+              </li>
+              {info?.backendType === "LDK" && (
```
rolznz, line 53: To also review in the onboarding / ConnectAlby page

frontend/src/components/UpgradeDialog.tsx
rolznz: @reneaaron I can have a look at this tomorrow

frontend/src/components/UpgradeDialog.tsx
reneaaron: Not sure I get your point? :thinking:

rolznz: Old comment - you now decided to keep it and allow user to either get pro or support via v4v :raised_hands:

frontend/src/components/UpgradeDialog.tsx
reneaaron: Since it was in there before, I'd postpone this one. I don't now how to handle an odd number of features in this list :sweat_smile:

frontend/src/components/UpgradeDialog.tsx
reneaaron: I think it's good enough for now.

frontend/src/screens/ConnectAlbyAccount.tsx
```diff
@@ -26,40 +28,76 @@ export function ConnectAlbyAccount({ connectUrl }: ConnectAlbyAccountProps) {
...
         />
         <div className="grid grid-cols-1 md:grid-cols-2 w-full gap-2 mt-5">
-          <Card className="w-full">
+          <Card className="w-full relative">
             <CardHeader className="flex flex-col justify-center items-center text-center p-4">
               <Zap className="w-6 h-6" />
-              <CardTitle className="text-sm">Lightning Address</CardTitle>
+              <CardTitle className="text-sm">
+                Lightning Address
+                <Badge
```
rolznz, line 39: outlined badge inside card looks weird here. What about just a plain div? 

image

frontend/src/screens/Home.tsx
```diff
@@ -59,7 +60,15 @@ function Home() {
 
   return (
     <>
-      <AppHeader title={getGreeting(albyMe?.name)} description="" />
+      <AppHeader
```
rolznz, line 63: When the user has Alby Pro would it be nice to show somewhere on the app that they have Alby Pro and "thank you for supporting Alby"?

rolznz (approved):
tACK

frontend/src/screens/Home.tsx
reneaaron: Yeah, I like that idea. I'd add it with the new sidebar though as I think a main menu item is a bit too prominent and I'd rather move that to some dropdown: (maybe instead of "Upgrade"?

image

PR #1155, opened 2025-03-06
fix: adjust animation for darkmode

### LINKED ISSUE #1096: Node starting animation broken in dark mode on the finish setup page
Image

Do we have a dark mode version for the lottie animation?

reneaaron:
Fixes #1096

It's not the most elegant solution (I learned lotties are themable now :exploding_head:), but that would require some changes in the animation file itself and to move to dotLottie file format. 

Let's consider moving if we ever have other animations that need theming.


rolznz (approved):
utACK

Issue #1134, opened 2025-02-27
Remind users to swap when channels are unbalanced


rolznz:
- When user plans to open new channel and they already have channels with 2+ counterparties
- When user plans to close a channel
- Show on low receiving capacity / unbalanced channels warnings

PR #1158, opened 2025-03-07
feat: add swap alert

rolznz:
Closes https://github.com/getAlby/hub/issues/1134

Adds an alert to swap rather than opening/closing channels in certain relevant places:
- when opening a channel
- when closing a channel
- low receiving capacity on channels page

The alert on other pages will open a link to `/channels?swap=true` which will auto-detect which swap should be used.

Channels page - low receiving capacity

image

Open new channel pages

image

Close channel dialog

image

I'm wondering if we should have a similar alert somewhere to swap in to "top up" channels. Maybe on the receive page?

Copy needs review.

Swap dialogs were moved into a separate file to keep the channels component as small as possible.

### DIFF
--- a/frontend/src/components/CloseChannelDialogContent.tsx
+++ b/frontend/src/components/CloseChannelDialogContent.tsx
@@ -5,6 +5,7 @@ import {
   ExternalLinkIcon,
 } from "lucide-react";
 import React from "react";
+import { SwapAlert } from "src/components/channels/SwapAlert";
 import ExternalLink from "src/components/ExternalLink";
 import { MempoolAlert } from "src/components/MempoolAlert";
 import { Alert, AlertDescription, AlertTitle } from "src/components/ui/alert";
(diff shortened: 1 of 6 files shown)


reneaaron (approved):
Improved the overall layout & sizing a bit. Not 100% happy with the info boxes stacked on top of each other, but also have no better idea right now. 

tACK

> I'm wondering if we should have a similar alert somewhere to swap in to "top up" channels. Maybe on the receive page?

Not sure if I understand that: Can you give an example of where you'd add this hint?

rolznz:
> Not sure if I understand that: Can you give an example of where you'd add this hint?

@reneaaron we have this warning when you have low receiving capacity:

image

but no warning for low balance (and we decided to remove that). There are a few places we could suggest swapping in (on this page, on the wallet -> receive screen, on the send page where it has a Top Up button - which would show this alert once they click it). But none of these places really feel right to me, so I left it out for now.

MoritzKa:
Copy looks good. 

I was wondering if we could simplify copy to just "swap in" or "swap out" according to the context (e.g. on the closing channel modal). E.g. instead of closing a channel I can "swap out". But different users might see it differently.

rolznz:
@MoritzKa good idea, thanks. We already do the calculation on the node page to know which swap dialog to open. The calculation is simple so I added it to the alert too.

image

image

Issue #1159, opened 2025-03-10
"Channel reserves unmet" - change the popup's information


itstomekk:
User comes with a valid comment, that this is a bit confusing

![Image](https://github.com/user-attachments/assets/1dcb2f0f-cc4f-47bf-a798-0c392d2ece95)

rolznz:
@itstomekk I updated the copy here https://github.com/getAlby/hub/pull/1161

![Image](https://github.com/user-attachments/assets/c7a5e72d-e51f-4067-b7f7-e9e2bdf6873d)

PR #1161, opened 2025-03-10
feat: show total and reserved balance in spending balance tooltip

rolznz:
We can point people to check this if they think sats are missing from the spending balance (actually are part of channel reserves)

Fixes https://github.com/getAlby/hub/issues/1159

Alternative to https://github.com/getAlby/hub/pull/1105


rolznz:
@im-adithya I tested on both LDK and LND, it looks good to me

PR #1163, opened 2025-03-10
fix: do not publish info events on relay reconnect

rolznz:
This is redundant and causes a lot of events to be published on the relay at the same time


im-adithya:
Nice 👌

PR #1175, opened 2025-03-12
fix: set target conf to 1 while withdrawing

im-adithya:
Fixes https://github.com/getAlby/hub/issues/684

Sets the target conf to 1 so LND can adjust the fee accordingly to process the tx ASAP ⚡


rolznz (approved):
utACK

PR #1176, opened 2025-03-12
fix: cleanup sidebar

reneaaron:
Guides, Feedback board and support options will be linked from support.getalby.com

image


rolznz (approved):
utACK

PR #1185, opened 2025-03-13
fix: initialize pending sweep balance to empty slice to avoid null value in frontend

im-adithya:
If you close a channel in LND and stay there, the frontend crashes because `PendingSweepBalancesDetails` is `null` instead of empty array `[]`


rolznz (approved):
utACK

Sorry about introducing this bug :see_no_evil:

PR #1186, opened 2025-03-13
feat: add routing hints for private nodes in LND

### LINKED ISSUE #1171: Bug: LND only recommends one routing hint to receive payments (+ has issues receiving on ACINQ channel)
We could recommend up to 3 unique counterparties (this is what LDK does). LND team says we can specify them manually, since the default seems to be a single route hint.

I am not sure why users with an ACINQ channel with plenty of incoming liquidity are having trouble receiving payments. I contacted the LND team but so far there is no mention of a bug. But adding other route hints is at least a partial workaround.

im-adithya:
Fixes #1171

Goes through all the private channels, if there is a routing policy, it adds that channel as a hint, skips if it doesn't OR if the peer is already processed i.e. added as a hint.

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -468,12 +468,23 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
 		}
 	}
 
+	isPrivate := !hasPublicChannels
+
+	var hints []*lnrpc.RouteHint
+	if isPrivate {
+		hints, err = svc.getRoutingHints(ctx)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	addInvoiceRequest := &lnrpc.Invoice{
 		ValueMsat:       amount,
 		Memo:            description,
 		DescriptionHash: descriptionHashBytes,
 		Expiry:          expiry,
-		Private:         !hasPublicChannels, // use private channel hints in the invoice
+		RouteHints:      hints,
+		Private:         isPrivate, // use private channel hints in the invoice
 	}
 
 	resp, err := svc.client.AddInvoice(ctx, addInvoiceRequest)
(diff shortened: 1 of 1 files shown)


lnclient/lnd/lnd.go
```diff
@@ -492,6 +503,60 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
 	return transaction, nil
 }
 
+func (svc *LNDService) getRoutingHints(ctx context.Context) (hints []*lnrpc.RouteHint, err error) {
```
rolznz, line 506: Did you take inspiration from anywhere for this code? I think there are quite a few issues here.

Can you please check the LND codebase and some other lightning wallets (e.g. https://github1s.com/ZeusLN/zeus/blob/HEAD/lndmobile/index.ts#L632-L633)

I think there's no limit right now, right? we need to limit it to 3 (the "best" options based on incoming liquidity and fees)

Can you also check if we have 1 peer the route hint is _exactly_ the same as is when not using this code. We need to make very sure we don't break this code otherwise no-one with private channels will be able to receive.

lnclient/lnd/lnd.go
```diff
@@ -492,6 +503,60 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
...
+			ChanId: channel.ChanId,
+		})
+		if err != nil {
+			logger.Logger.WithFields(logrus.Fields{
+				"channel_id": channel.ChanId,
+			}).WithError(err).Error("Unable to add routing hint")
+			continue
+		}
+
+		remotePolicy := chanInfo.Node2Policy
```
rolznz, line 530: Node2 is not always the counterparty is it?

im-adithya: Ah true, will add a check using node pubkey

lnclient/lnd/lnd.go
im-adithya: I took inspiration from Breez. I will also check the Zeus code and see if we can do any improvements.

> I think there's no limit right now, right? we need to limit it to 3 (the "best" options based on incoming liquidity and fees)

LND has a max limit of 20 hints, but yeah will limit this to use best 3 hints. I think the only parameter to decide what's "better" would be incoming liquidity right?

> Can you also check if we have 1 peer the route hint is exactly the same as is when not using this code.

Yup, I checked that and it matches in my local Polar setup, will make the above changes and check again to be sure :+1:

rolznz:
@im-adithya looked into the code and seems an issue on the LND side, right? I think we should assume LND normally gives good default route hints and not make any code changes here.

We still need to confirm if it's different when used from the hub vs lncli directly and have a recommendation for affected users.

PR #1187, opened 2025-03-13
feat: add info event publish queue

### LINKED ISSUE #1170: Do not publish all info events in parallel on startup
We currently do this to ensure any app info events that might have been dropped are re-published.

Depending on the relay and how many apps the hub has, the hub could be rate limited.

im-adithya:
Fixes #1170

### DIFF
--- a/nip47/publish_nip47_info.go
+++ b/nip47/publish_nip47_info.go
@@ -16,6 +16,35 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+type Nip47InfoPublishRequest struct {
+	AppWalletPubKey  string
+	AppWalletPrivKey string
+}
+
+type nip47InfoPublishQueue struct {
+	channel chan *Nip47InfoPublishRequest
+}
+
+func NewNip47InfoPublishQueue() *nip47InfoPublishQueue {
+	return &nip47InfoPublishQueue{
+		channel: make(chan *Nip47InfoPublishRequest, 1000),
+	}
+}
+
+func (q *nip47InfoPublishQueue) AddToQueue(req *Nip47InfoPublishRequest) {
+	select {
+	case q.channel <- req: // Put in the channel unless it is full
+		// successfully sent to channel
+	default:
+		// channel full
+		logger.Logger.WithField("info_request", req).Error("Nip47InfoPublishQueue channel full. Discarding value")
+	}
+}
+
+func (q *nip47InfoPublishQueue) Channel() <-chan *Nip47InfoPublishRequest {
+	return q.channel
+}
+
 func (svc *nip47Service) GetNip47Info(ctx context.Context, relay *nostr.Relay, appWalletPubKey string) (*nostr.Event, error) {
 	filter := nostr.Filter{
 		Kinds:   []int{models.INFO_EVENT_KIND},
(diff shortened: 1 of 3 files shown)


service/start.go
```diff
@@ -91,6 +91,7 @@ func (svc *service) startNostr(ctx context.Context) error {
 			waitToReconnectSeconds = 0
 
 			svc.nip47Service.StartNotifier(relay.Context(), relay)
+			svc.nip47Service.StartNip47InfoPublisher(relay.Context(), relay, svc.lnClient)
```
im-adithya, line 94: We can add a context.WithCancel and call it where we set `publishInfoEvents = false` but I think it is fine

rolznz: I think we can rewrite this code now to remove the `publishInfoEvents` boolean and only once on startup add the entries to the queue. Does that make sense?

rolznz: > We can add a context.WithCancel and call it where we set publishInfoEvents = false but I think it is fine

Sorry, I don't understand what you mean here. Why do we need it? the publisher already listens to `ctx.Done` which will be consumed when the relay context is cancelled

im-adithya: Yes, but once the publish is done, we can cancel the context to stop the publisher right? Since we pass the relay context to the publisher and it stays waiting for a event to publish without cancelling. (Unlike Notifier which needs to keep listening)

rolznz: Ah ok, thanks

service/start.go
```diff
@@ -158,31 +149,47 @@ func (svc *service) startNostr(ctx context.Context) error {
 	return nil
 }
 
-func (svc *service) startAllExistingAppsWalletSubscriptions(ctx context.Context, relay *nostr.Relay, publishInfoEvent bool) {
+func (svc *service) queueAppInfoEvents() {
```
rolznz, line 152: Can we add a comment to the top of the method:
```
In case the relay somehow loses events or the hub updates with new capabilities, we re-publish info events for all apps on startup to ensure that they are retrievable for all connections.
```

nip47/publish_nip47_info.go
```diff
@@ -16,6 +16,35 @@ import (
...
+	AppWalletPrivKey string
+}
+
+type nip47InfoPublishQueue struct {
+	channel chan *Nip47InfoPublishRequest
+}
+
+func NewNip47InfoPublishQueue() *nip47InfoPublishQueue {
+	return &nip47InfoPublishQueue{
+		channel: make(chan *Nip47InfoPublishRequest, 1000),
```
rolznz, line 30: This might become an issue if a hub has more than 1000 apps. I guess we log an error but I wonder if we need to be explicit here with the number? (I am just thinking about e.g. Yakihonne or https://lnwallet.app/ which allow creating new wallets with one click)

service/start.go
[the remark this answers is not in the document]
im-adithya: Done!

im-adithya: ✅

nip47/publish_nip47_info.go
im-adithya: Removed size restriction :+1:

rolznz: @im-adithya this was completely wrong, please test your changes. I've fixed it in the latest commit

rolznz (approved):
tACK

nip47/publish_nip47_info.go
im-adithya: Sorry for that, should've tested after changing!

PR #1191, opened 2025-03-16
feat: add Ark LNClient

rolznz:
Testing out the Ark SDK. Currently you can send and receive vtxos via custom node commands and see your off-chain balance. You can pay and receive lightning payments (although receive will throw an error because we don't set the preimage yet)

TODOs:
- [ ] Onboarding experimental warning
- [ ] Onboarding actually works (setting backend type to Ark via onboarding)
- [ ] Do not allow in cloud setup (needs local file storage) - or test switching to in-memory? is there a trust tradeoff there?
  - [ ] test same seed recovering with a brand new hub
  - [ ] test same seed in fulmine
- [ ] failed payment tests
  - [x] invoice already paid by other wallet - Ark successfully recovers the funds
    - [ ] re-test that Alby hub correctly marks the payment as failed
    - [ ] invoice already paid by other wallet - exit the goroutine before the recovery is started
  - [x] invoice expired - alby hub prevents this
  - [ ] swap expired (recipient not online at all)
- [ ] pending payment tests
  - [ ] recipient is offline then later comes online after hub shutdown while waiting for swap
- [ ] Save pending swaps
  - [ ] Test restarting app while swap is in process (manual command to recover)
  - [ ] Test recovering funds when boltz failed to pay (e.g. pay a previously valid invoice that then expires)
- [x] Provide preimage when making invoice
- [ ] Refreshing VTXOs background task
- [ ] List VTXOs (via debug tools only)
- [ ] List off-chain transactions?
- [ ] Understand better how the swaps work: how long is the swap valid (2 weeks? or what is the RefundLocktime for? why doesn't the Typescript SDK do the same? do they just base it from the boltz state?)
  - [x] if boltz definitely fails to pay it allows immediate refund
- [ ] is it ok for now to save pending swaps in the config object (none of our other LNClients need this or have their own data stores)


frontend/src/screens/setup/node/ArkForm.tsx
```diff
@@ -0,0 +1,28 @@
+import { wordlist } from "@scure/bip39/wordlists/english";
```
rolznz, line 1: This is not necessary right now because we configure the LNClient via ENV

frontend/src/routes.tsx
```diff
@@ -59,6 +59,7 @@ import { SetupFinish } from "src/screens/setup/SetupFinish";
 import { SetupNode } from "src/screens/setup/SetupNode";
 import { SetupPassword } from "src/screens/setup/SetupPassword";
 import { SetupSecurity } from "src/screens/setup/SetupSecurity";
+import { ArkForm } from "src/screens/setup/node/ArkForm";
```
rolznz, line 62: This is not necessary right now because we configure the LNClient via ENV

PR #1194, opened 2025-03-17
fix: add lock to prevent races when checking balances and creating transactions

rdmitr:


transactions/transactions_service.go
```diff
@@ -226,6 +231,9 @@ func (svc *transactionsService) SendPaymentSync(ctx context.Context, payReq stri
 	}
 
 	err = svc.db.Transaction(func(tx *gorm.DB) error {
+		balanceValidationLock.Lock()
```
rolznz, line 234: Is it correct to put this inside the DB transaction rather than outside? e.g. wrap this db transaction in a function with the locking at the beginning of the function?

It would be really great to create a test that definitely fails without the change in this PR (for example do 100 concurrent payments) and confirm this fixes it

rolznz:
@rdmitr there's a lot of refactoring in this transactions service compared to the initial commit, could you please explain your reasoning for the changes?

rolznz:
@rdmitr I would rather not make large changes here unless necessary and there is some added duplication now. I thought we could just slightly alter your initial commit to instead use a function and put the lock outside of the DB transaction. Would you be OK to revert back to something like this?

```
func() {
    balanceValidationLock.Lock()
    defer balanceValidationLock.Unlock()
    // .. do DB transaction as before
}()
```

rolznz:
@rdmitr thanks, I understand. In this case it's highly critical logic and I would prefer to keep refactoring separate from the bugfix. Even though we have tests I cannot guarantee we have 100% coverage of this part of the code.

rolznz:
@rdmitr it seems even without the fix, the test you created already passes https://github.com/getAlby/hub/pull/1214

first we must be able to have a test that reproduces the issue. Any idea why it passed? can we change the test instead to add delays in e.g. the payment part? then we might not even need the dummy invoice creation, we could probably just do it with 2 invoices?

rolznz:
@rdmitr could you also update the README regarding testing e.g. by default is uses SQLite, but what should be done to run postgres tests locally (and what to be aware of)

rolznz:
@rdmitr thanks for making the changes. I'll try and come up with an elegant way to force this issue to occur in the tests. Please let me know if you have any ideas too

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/1222

Thanks for the fix and documentation @rdmitr !

PR #1195, opened 2025-03-17
feat: bitrefill

reneaaron:


frontend/src/screens/internal-apps/Bitrefill.tsx
```diff
@@ -0,0 +1,139 @@
...
+  AlertDialogHeader,
+  AlertDialogTitle,
+} from "src/components/ui/alert-dialog";
+import { useToast } from "src/components/ui/use-toast";
+import { PayInvoiceResponse } from "src/types";
+import { request } from "src/utils/request";
+
+export function Bitrefill() {
+  const { toast } = useToast();
+  const [open, setOpen] = React.useState(false);
```
rolznz, line 21: can you rename to `paymentDialogOpen` or something?

frontend/src/screens/internal-apps/Bitrefill.tsx
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    try {
+      setLoading(true);
+      const payInvoiceResponse = await request<PayInvoiceResponse>(
+        `/api/payments/${invoice.paymentRequest}`,
+        {
+          method: "POST",
+          body: JSON.stringify({
+            amount: invoice.satoshi * 1000,
```
rolznz, line 62: I don't think the amount should be set, it's for 0-amount invoices. Bitrefill will always have one set

frontend/src/screens/internal-apps/Bitrefill.tsx
```diff
@@ -0,0 +1,139 @@
...
+  useEffect(() => {
+    const handleIframeMessage = (event: MessageEvent) => {
+      if (event.origin === "https://embed.bitrefill.com" && event.data) {
+        // Some of the events have objects in event.data 🤷‍♂️
+        try {
+          const parsedData = JSON.parse(event.data);
+
+          if (parsedData.event === "payment_intent") {
+            setOpen(true);
+            const invoice = new Invoice({ pr: parsedData.paymentAddress });
```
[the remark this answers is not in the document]
reneaaron: i want to display the amount from the parsed invoice, that's why I parsed it here (instead of relying what bitrefill passes on to us)

frontend/src/screens/internal-apps/Bitrefill.tsx
reneaaron: Thanks for the hint, didn't pay enough attention.

rolznz (approved):
tACK (except mainnet payment using the hub)

frontend/src/screens/internal-apps/Bitrefill.tsx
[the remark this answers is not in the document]
rolznz: Ah right, thanks!

PR #1203, opened 2025-03-18
fix: text overflowing in used apps widget

im-adithya:
## Description

Fixes text overflowing in apps widget.

### DIFF
--- a/frontend/src/components/home/widgets/LatestUsedAppsWidget.tsx
+++ b/frontend/src/components/home/widgets/LatestUsedAppsWidget.tsx
@@ -33,21 +33,15 @@ export function LatestUsedAppsWidget() {
           .slice(0, 3)
           .map((app) => (
             <Link key={app.id} to={`/apps/${app.appPubkey}`}>
-              <div className="flex items-center justify-between w-full">
-                <div className="flex items-center justify-center gap-4">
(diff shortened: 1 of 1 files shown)


frontend/src/components/home/widgets/LatestUsedAppsWidget.tsx
```diff
@@ -33,21 +33,15 @@ export function LatestUsedAppsWidget() {
...
-                </div>
-                <div className="flex items-center justify-center gap-4">
-                  <p className="text-xs text-muted-foreground">
-                    {app.lastEventAt
-                      ? dayjs(app.lastEventAt).fromNow()
-                      : "never"}
-                  </p>
-                  <ChevronRight className="text-muted-foreground w-8 h-8" />
-                </div>
+              <div className="flex items-center justify-between w-full gap-4">
```
rolznz, line 36: I don't think the justify-between applies properly now since there are 4 children inside the div. How does it look if you only have shorter app names?

im-adithya: It works fine but yeah this `justify-between` is unnecessary, since the app name `p` tag has `flex-1` which will grow and shrink depending on space available

> How does it look if you only have shorter app names?

You can see the "spammer" app in the above after screenshot

rolznz (approved):
utACK

PR #1205, opened 2025-03-18
fix: success screen crashes instead of redirecting

im-adithya:
## Description

Visiting `/wallet/send/success` page crashes instead of taking to `/wallet/send`. I ran into this on mobile, as often in mobile browsers, the tab refreshes on opening (and it is possible that the tab has last been in success screen)


rolznz (approved):
tACK

Issue #1206, opened 2025-03-18
Bug: Funding source offline, no login possible


AxelHamburch:
I am using AlbHub 1.15.0 with Phoenixd in the backend.

I have noticed AlbyHub login screen just stays black when Phoenixd is stopped. As soon as the funding source is online again, the login screen is displayed correctly again.

If I am logged in, it doesn't matter. I can jump back and forth. I don't even see that my funding source is offline. As soon as I log out, my screen goes black.

rolznz:
I think this is working as designed, same with how LND works right @im-adithya ?

We could have a banner that shows if the node cannot be connected to. Currently we did not add it as we prioritize the LDK backend, which is embedded and always online when Alby Hub is unlocked

AxelHamburch:
The welcome screen is not shown @im-adithya. It's just a black screen. Would be very helpful if at least a banner showed up and informed you that the funding source is not connected.

I moved my system to a different server, and Phoenixd assigned a different API key despite the same seed and channel. This is how I discovered that I get a black screen when I shut down Phoenixd. I can't change it later, right?

However, I've already found a workaround for my move. I simply moved the entire phoenixd and .phoenix folders and assigned the permissions to the new user. It worked perfectly, and everything is running like it did on the source code.

rolznz:
@AxelHamburch thanks for reporting.

I can reproduce this, the GetInfo call in the Phoenix implementation is returning an error, which ends up causing our API GetInfo call to fail, so we just get a blank screen.

We did a similar fix in LND to cache the node info. We don't need a banner actually since we already have a healthcheck indicator.

rolznz:
Fixed by https://github.com/getAlby/hub/pull/1210 and will go out in 1.16

PR #1202, opened 2025-03-18
feat: link to node page from healthcheck indicator and show healthcheck alert

rolznz:
Closes https://github.com/getAlby/hub/issues/1178

Rather than showing a toast when hovering over the healthcheck, it links to the node page. By default the node page will only show an alert if the healthcheck is failing, however if they click on the green healthcheck indicator, it'll show a success alert.

Now the alby services with incidents will show along with their status, and we link directly to the alby status page and a guide where the user can learn more.

image

image

PR #1210, opened 2025-03-19
fix: alby hub blank screen when phoenixd is offline

rolznz:
Fixes https://github.com/getAlby/hub/issues/1206

I did the same fix as we have for LND, that instead we only fetch the node info once on startup.

In https://github.com/getAlby/hub/pull/1202 we link to the node page, which needs changes for Phoenixd and Cashu so they cannot access channel-related pages/features.

image


rolznz:
@im-adithya I think you are right, but also it works ok for now but could be re-visited when it becomes a problem. (possibly with Ark if this turns out to be worthwhile to build more than a minimal implementation)

im-adithya:
Yup, just wanted to mention so we can revisit at some point, all good for now 🙌

PR #1211, opened 2025-03-21
feat: add update button to what's new widget

rolznz:
image


im-adithya (commented):
Since all the buttons are aligned to the right on the HomeScreen widgets, it would be good to follow the same here too

Screenshot 2025-03-21 at 12 50 04 PM

rolznz:
@im-adithya I simplified the card and made it like the others, but kept the primary variant since we want people to update
image

PR #1213, opened 2025-03-21
fix: open links from async functions on ios

reneaaron:
https://stackoverflow.com/questions/20696041/window-openurl-blank-not-working-on-imac-safari


im-adithya (approved):
tACK

PR #1215, opened 2025-03-24
fix: error log when fetching currency

rolznz:
removes excess debug logs and adds error log if the currency actually failed to be fetched


im-adithya (approved):
utACK

PR #1217, opened 2025-03-24
fix: return offline channel balance in get_balance nwc call

### LINKED ISSUE #1183: Include offline channel balance in nwc get_balance
Reasoning:
- It's just the worst feeling you can have if you open your wallet and have "0" balance while you had 100k yesterday (even if you didn't even want to spend)
- It's better to run into problems at the time of sending as it doesn't confront users before it has to

im-adithya:
Fixes #1183

We introduce a new internal variable called `includeInactiveChannels` in the `LNClient` `GetBalances` call. But we only set it to true in the `GetBalancesController` so that we only return offline channel balance externally (so as to not trouble the user).

Internally (in the Hub) they'll see the reduced balance but the user will anyways be informed about their offline channel from the indicator so it shouldn't be an issue.

### Testing

Tested in both LND and LDK by disconnecting peers. NWC `get_balance` includes offline channel balances as well as intended :+1:


rolznz (approved):
utACK (Update: tested LDK)

PR #1218, opened 2025-03-24
chore: include funding tx id and balance in channel close event

### LINKED ISSUE #1197: Include incoming balance in channel closure email
This is hopefully so user can immediately see how much money will be returned (especially if they don't open the hub immediately - they might not see this on the UI)

(Also the open channel TXID if possible).

Both can be added to `nwc_channel_closed` event

im-adithya:
Partly fixes #1197

Have to add the same in LDK

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -250,12 +250,20 @@ func (svc *LNDService) subscribeChannelEvents(ctx context.Context) {
 						"reason":               closureReason,
 					}).Info("Channel closed")
 
+					var fundingTxId string
+					channelPoint, _ := svc.parseChannelPoint(update.ClosedChannel.ChannelPoint)
+					if channelPoint != nil {
+						fundingTxId = channelPoint.GetFundingTxidStr()
+					}
+
 					svc.eventPublisher.Publish(&events.Event{
 						Event: "nwc_channel_closed",
 						Properties: map[string]interface{}{
 							"counterparty_node_id": counterpartyNodeId,
 							"reason":               closureReason,
 							"node_type":            config.LNDBackendType,
+							"funding_tx_id":        fundingTxId,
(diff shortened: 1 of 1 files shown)


lnclient/lnd/lnd.go
```diff
@@ -250,12 +250,20 @@ func (svc *LNDService) subscribeChannelEvents(ctx context.Context) {
...
+					}
+
 					svc.eventPublisher.Publish(&events.Event{
 						Event: "nwc_channel_closed",
 						Properties: map[string]interface{}{
 							"counterparty_node_id": counterpartyNodeId,
 							"reason":               closureReason,
 							"node_type":            config.LNDBackendType,
+							"funding_tx_id":        fundingTxId,
+							"incoming_balance":     update.ClosedChannel.SettledBalance,
```
im-adithya, line 266: This is not _incoming_ since in LND this triggers after close has finished, so the balance is already deposited in the users wallet by this point.

rolznz: Hmm, the intention of this event is to notify the user immediately when their channel begins closing so they know and can instantly do something (e.g. open a new channel if they want). Is there any event for this?

e.g. if this is a force close then this event wouldn't happen for days?

im-adithya: There's `ChannelEventUpdate_InactiveChannel` which triggers when the channel is closed, will try to see if that can work here.

im-adithya: It only gives the chan point, so we can get funding tx id but not the balance. And I tried to see if we can get the closed channel info using `GetChanInfo` but it says `edge not found` because it has been closed...

im-adithya: Also observed that this `ChannelEventUpdate_InactiveChannel` doesn't fire every time (on channel close)

rolznz:
Closing for now, seems like this is not easily possible with LND

PR #1219, opened 2025-03-24
fix: retry start app after failure to avoid downtime in self-hosted LND environments

im-adithya:
Fixes the issue which causes auto unlock to fail in Umbrel/Start 9 on reboot since the `StartApp` call is done while LND is initializing

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -53,9 +53,27 @@ func NewLNDService(ctx context.Context, eventPublisher events.EventPublisher, ln
 		logger.Logger.WithError(err).Error("Failed to create new LND client")
 		return nil, err
 	}
-	nodeInfo, err := fetchNodeInfo(ctx, lndClient)
+
+	var nodeInfo *lnclient.NodeInfo
+	maxRetries := 5
+	for i := range maxRetries {
+		nodeInfo, err = fetchNodeInfo(ctx, lndClient)
+		if err == nil {
+			break
+		}
+		if i < maxRetries-1 {
+			logger.Logger.WithFields(logrus.Fields{
+				"iteration": i,
+			}).WithError(err).Error("Failed to connect to LND, retrying in 10s")
+			time.Sleep(10 * time.Second)
+		} else {
+			logger.Logger.WithFields(logrus.Fields{
+				"iteration": i,
+			}).WithError(err).Error("Failed to connect to LND")
+		}
+	}
 	if err != nil {
-		logger.Logger.WithError(err).Error("Failed to fetch node info")
+		logger.Logger.WithError(err).Error("Failed to fetch node info after 5 attempts")
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


service/service.go
```diff
@@ -150,7 +152,21 @@ func NewService(ctx context.Context) (*service, error) {
 	if autoUnlockPassword != "" {
 		nodeLastStartTime, _ := cfg.Get("NodeLastStartTime", "")
 		if nodeLastStartTime != "" {
-			svc.StartApp(autoUnlockPassword)
+			var err error
```
rolznz, line 155: since this only applies to LND can we move it to the LND LNClient?

lnclient/lnd/lnd.go
```diff
@@ -53,9 +53,27 @@ func NewLNDService(ctx context.Context, eventPublisher events.EventPublisher, ln
...
+		if i < maxRetries-1 {
+			logger.Logger.WithFields(logrus.Fields{
+				"iteration": i,
+			}).WithError(err).Error("Failed to connect to LND, retrying in 10s")
+			time.Sleep(10 * time.Second)
+		} else {
+			logger.Logger.WithFields(logrus.Fields{
+				"iteration": i,
+			}).WithError(err).Error("Failed to connect to LND")
+		}
```
im-adithya, line 73: Wondering if this conditional is worth it

service/service.go
im-adithya: Done ✅

lnclient/lnd/lnd.go
```diff
@@ -1197,7 +1211,6 @@ func (svc *LNDService) GetPubkey() string {
 func fetchNodeInfo(ctx context.Context, client *wrapper.LNDWrapper) (*lnclient.NodeInfo, error) {
 	resp, err := client.GetInfo(ctx, &lnrpc.GetInfoRequest{})
 	if err != nil {
-		logger.Logger.WithError(err).Error("Failed to fetch node info")
```
rolznz, line 1200: why remove the error?

lnclient/lnd/lnd.go
```diff
@@ -53,9 +53,23 @@ func NewLNDService(ctx context.Context, eventPublisher events.EventPublisher, ln
...
+		if err == nil {
+			break
+		}
+		logger.Logger.WithFields(logrus.Fields{
+			"iteration": i,
+		}).WithError(err).Error("Failed to connect to LND, retrying in 10s")
+		time.Sleep(10 * time.Second)
+	}
+
+	nodeInfo, err = fetchNodeInfo(ctx, lndClient)
```
rolznz, line 70: why fetch again?

im-adithya: Ah, that was to not have the conditional (see [this](https://github.com/getAlby/hub/pull/1219#discussion_r2010140451) and first commit)

lnclient/lnd/lnd.go
rolznz: I think it's ok to add it (we are logging at a different function within the LNClient and it's easier to debug)

im-adithya: Done

lnclient/lnd/lnd.go
rolznz: I think you can just remove line 70, you already store the error and nodeInfo because the variables are declared outside the loop.

The only issue then is doing an unnecessary 10 second sleep on the last iteration but this is an edge case and means LND couldn't be connected to anyway. So I don't think it matters.

What do you think?

im-adithya: This is how the logs look like: (they make sense, but we are not actually retrying on the 6th time)

Screenshot 2025-03-25 at 12 49 43 PM

rolznz: I think it's ok, not worth the code duplication or conditional

PR #1220, opened 2025-03-25
feat: verify package signatures

rdmitr:
Fixes https://github.com/getAlby/hub/issues/1133

**Before merging ensure verify.sh can be fetched from getalby.com** :heavy_check_mark:  https://getalby.com/install/hub/verify.sh

### DIFF
--- a/scripts/linux-aarch64/install.sh
+++ b/scripts/linux-aarch64/install.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 ALBYHUB_URL="https://getalby.com/install/hub/server-linux-aarch64.tar.bz2"
+MANIFEST_URL="https://getalby.com/install/hub/manifest.txt"
+SIGNATURE_URL="https://getalby.com/install/hub/manifest.txt.asc"
 echo ""
 echo ""
 echo "⚡️ Welcome to Alby Hub"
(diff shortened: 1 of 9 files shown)


scripts/linux-aarch64/install.sh
```diff
@@ -11,12 +13,75 @@ read -p "Absolute install directory path (default: $HOME/albyhub): " USER_INSTAL
 
 INSTALL_DIR="${USER_INSTALL_DIR:-$HOME/albyhub}"
 
+verify_package() {
```
rolznz, line 16: can this function be moved to a separate script file?

scripts/linux-aarch64/install.sh
```diff
@@ -11,12 +13,75 @@ read -p "Absolute install directory path (default: $HOME/albyhub): " USER_INSTAL
...
+  local archive_file="${1}"
+  local overridden_manifest_name="${2}"
```
rolznz, line 18: the variable name was a bit confusing to be. Could it be `filename_in_manifest`?

bumi:
btw. I don't think we should over engineer those bash scripts. 
They are basic, working, an inspiration but will never be perfect for all the environments. And I trying this will become an endless story. 

we maintain basic packages (docker, umbrel, start9, hosted version, etc.) but for custom setups people have to know a bit computer.

scripts/linux-aarch64/install.sh
bumi: then another file needs to be downloaded (and actually verified 🤔 ) 

I'd keep those simple

scripts/linux-aarch64/update.sh
```diff
@@ -57,6 +58,22 @@ cp -r data albyhub-backup
 
 echo "Downloading latest version"
 wget $ALBYHUB_URL
+
+if [[ ! -f "verify.sh" ]]; then
```
rolznz, line 62: I don't think we need to download it again here too, just in the installer

scripts/linux-aarch64/update.sh
```diff
@@ -57,6 +58,22 @@ cp -r data albyhub-backup
...
+  if ! wget -q "$VERIFIER_URL"; then
+    echo "❌ Failed to download the verification script." >&2
+    exit 1
+  fi
+  chmod +x verify.sh
+fi
+
+./verify.sh server-linux-aarch64.tar.bz2 albyhub-Server-Linux-aarch64.tar.bz2
+if [[ $? -ne 0 ]]; then
+  echo "❌ Verification failed, aborting installation"
```
rolznz, line 73: What happens if the user hasn't imported the gpg key yet?

Should we update this message to also include: `. Visit https://github.com/getAlby/hub/releases for more information on how to verify the release`

scripts/verify.sh
```diff
@@ -0,0 +1,71 @@
...
+  fi
+
+  echo "Downloading manifest signature file..."
+  if ! wget -q "$SIGNATURE_URL"; then
+    echo "❌ Failed to download manifest signature file." >&2
+    return 1
+  fi
+
+  if ! gpg --batch --verify "manifest.txt.asc" "manifest.txt"; then
+    echo "❌ GPG signature verification failed!" >&2
```
rolznz, line 40: What happens if the user hasn't imported the gpg key yet?

Should we update this message to also include: `. Visit https://github.com/getAlby/hub/releases for more information on how to verify the release`

scripts/linux-aarch64/update.sh
rolznz: Edit: moved this question.

scripts/linux-aarch64/install.sh
```diff
@@ -17,6 +18,22 @@ cd $INSTALL_DIR
 
 # download and extract the Alby Hub executable
 wget $ALBYHUB_URL
+
+if [[ ! -f "verify.sh" ]]; then
```
rolznz, line 22: can you update the other install/update files?

rolznz:
@rdmitr thanks! I tested it on x86_64 and it worked great. Everything looks good to me.

However, before merging we need to make sure getalby.com is updated and the verifier URL is working.

rolznz (approved):
tACK (x86_64)

Sakrecoer:
It looks like you forgot to make sure verify.sh is fetched in the `update.sh` script:
https://github.com/getAlby/hub/blob/master/scripts/pi-aarch64/update.sh

Also, the verify.sh does not fetch the public GPG key. Took me a minute to figure out it is available here: https://github.com/getAlby/hub/releases

rolznz:
@Sakrecoer thanks for your feedback.

Normally when you install Alby Hub it will save both the install and update script. If you always fetch the update.sh script from a URL there is the risk that the update script could have been modified. So my idea was that you would download the install/update scripts only once, and continue to use the local scripts to update. What do you think?

Regarding verify.sh not including the public GPG key, it should have a link in the error log, right?

rolznz:
@Sakrecoer thanks for pointing that out - we should update the instructions so people use their local scripts. I will review and update them.

Any chance you could share the verification error you get?

PR #1222, opened 2025-03-25
Fix: postgres pay race condition

rolznz:
Alternative to https://github.com/getAlby/hub/pull/1194


rolznz:
@rdmitr yeah, the actual fix in the transactions service is the same, just the tests are different in this PR.

I've reduced the delay to 1s and confirmed it still fails without the locks.

PR #1223, opened 2025-03-25
fix: connections page layout breaking in mobile

### LINKED ISSUE #1221: Connections page layout breaks in mobile
It happens because we now have two buttons in the header there

Image

im-adithya:
Fixes #1221

## Screenshots (Before vs After)

<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>


frontend/src/screens/apps/AppList.tsx
```diff
@@ -39,16 +41,23 @@ function AppList() {
           <>
             {!!unusedApps.length && (
               <Link to="/apps/cleanup">
-                <Button variant="outline">
-                  <Trash className="h-4 w-4 mr-2" />
-                  Cleanup Unused
-                </Button>
+                {isDesktop ? (
```
rolznz, line 44: can just the text be conditional? like the add connection change?

reneaaron:
Thanks for fixing it. With the new layout suggestions from Jakub I think we should remove the text completely on mobile and just show the icons: 

image
https://www.figma.com/design/2IyilDewDrdNP3G1NixPYd/%F0%9F%8C%93-Alby-Hub?node-id=6580-89958&p=f&m=dev

im-adithya:
Oh didn't notice, thanks for that. I changed other instances too (except Upgrade in Home screen ig it's fine there to highlight it), not sure what to do with this one since Open Channels doesn't have an icon + spacing looks good already:

Screenshot 2025-03-25 at 6 38 08 PM

frontend/src/screens/apps/AppList.tsx
```diff
@@ -8,13 +8,15 @@ import AppCard from "src/components/connections/AppCard";
...
 import { useInfo } from "src/hooks/useInfo";
+import { useIsDesktop } from "src/hooks/useMediaQuery";
 import { useUnusedApps } from "src/hooks/useUnusedApps";
 
 const albyConnectionName = "getalby.com";
 
 function AppList() {
   const { data: apps } = useApps();
   const { data: info } = useInfo();
+  const isDesktop = useIsDesktop();
```
rolznz, line 19: Minor: we could use tailwind breakpoints rather than this hook. Not sure which is better or if it really matters CC @reneaaron

rolznz (approved):
utACK

frontend/src/screens/apps/AppList.tsx
reneaaron: I'd vote for the tailwind classes as all other responsive stuff (columns, grids, etc) is also implemented that way.

im-adithya: It was used first in Credenza but then we removed that (and we are fine without it now ig?), so we can replace that with tailwind classes directly, will do that in this PR itself

im-adithya: Done!

frontend/src/screens/appstore/AppStore.tsx
```diff
@@ -16,10 +16,13 @@ function AppStore() {
               target="_blank"
               rel="noreferrer noopener"
             >
-              <Button variant="outline">
+              <Button className="hidden lg:inline-flex" variant="outline">
                 <CirclePlus className="h-4 w-4 mr-2" />
                 Submit your app
               </Button>
+              <Button className="lg:hidden" variant="outline" size="icon">
```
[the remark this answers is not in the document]
rolznz: Nice idea!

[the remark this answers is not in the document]
im-adithya: Done, added `ResponsiveButton` component for this

rolznz (approved):
tACK

PR #1224, opened 2025-03-25
fix: classname does not exist error in shadcn components

im-adithya:
Related https://github.com/shadcn-ui/ui/issues/1644

Open `frontend/src/components/ui/dialog.tsx` or `frontend/src/components/ui/sheet.tsx`

You'll see redlines with `classname is not assignable to type` errors

PR #1225, opened 2025-03-25
chore: replace all radix icons with lucide

im-adithya:
Title basically, while commiting I was getting the error described in https://github.com/getAlby/hub/pull/1224 which is weird since nothing changed but good to fix it.


im-adithya:
Also changed all icon component names to suffix `Icon`, please use that from now on so we can identify Icon components from others. It gets confusing sometimes Ex: icons like `Loader`

rolznz:
@im-adithya this PR does not seem to do what it says. There is no removal of radix icons is there?

im-adithya:
Can you check the [first commit](https://github.com/getAlby/hub/pull/1225/commits/52dd2cc05ed9ba09a233bbb146ad51b71306f956), it removes the `@radix-ui/react-icons` package. Rest two are for naming changes.

rolznz:
@im-adithya oops, sorry. I think I opened a specific commit instead of the whole PR. Thanks

pavanjoshi914 (approved):
along with replacing radix-ui icons this pr mostly addresses use of lucideIcon with a suffix "Icon". this can happen in future too that icons with and without suffix are used. 

can we also do something like this: https://github.com/lucide-icons/lucide/issues/1830#issuecomment-1913340043

so that library only exports icons with suffix. (library will show "HomeIcon" as suggestion and not "Home")

utAck

rolznz (approved):
tACK

PR #1226, opened 2025-03-26
feat: add funding txid and amount to LDK channel closed event

rolznz:
Fixes https://github.com/getAlby/hub/issues/1197 (LDK only)

Adds new properties to the `nwc_channel_closed` event for both LDK and LND:
```
counterparty_node_url
```

Adds new properties to the `nwc_channel_closed` event to LDK only:
```
pending_balance
funding_tx_id
funding_tx_vout
funding_tx_url
```

CC @MoritzKa (to review the new event properties which we could show in the email)


im-adithya (approved):
tACK

rolznz:
@MoritzKa I think the opening tx is still useful for reference. Could we include it underneath the pending balance? or do you see some downside? (I know it's not optimal, but it's what we have)

PR #1227, opened 2025-03-26
fix: always show swap alert on close channel dialog

rolznz:
Even if the user has 1 channel there's no need to close it when they could swap instead


im-adithya (approved):
tACK, but just faced another issue (unrelated to the changes in this PR). If you open the Swap In modal and then close it and try to open it again, it doesn't work (since we are already on `/channels?swap=true`)

rolznz:
Ah, this also does not work well. Thanks for catching this @im-adithya 

I pushed a fix

PR #1228, opened 2025-03-27
fix: show health alert if vss is enabled but subscription was cancelled

rolznz:
VSS is a paid feature but Alby Hub does not handle if this endpoint is only accessible by subscribers.

image

### DIFF
--- a/service/start.go
+++ b/service/start.go
@@ -449,7 +449,18 @@ func (svc *service) requestVssToken(ctx context.Context) (string, error) {
 		vssToken, err = svc.albyOAuthSvc.GetVssAuthToken(ctx, vssNodeIdentifier)
 		if err != nil {
 			logger.Logger.WithError(err).Error("Failed to fetch VSS JWT token")
+
+			existingVssToken, _ := svc.cfg.Get("VssToken", "")
+			if existingVssToken != "" {
+				logger.Logger.Warn("Using stored VSS JWT token")
+				return existingVssToken, nil
+			}
+
 			return "", err
+		}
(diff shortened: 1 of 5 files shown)


service/start.go
```diff
@@ -449,8 +449,19 @@ func (svc *service) requestVssToken(ctx context.Context) (string, error) {
 		vssToken, err = svc.albyOAuthSvc.GetVssAuthToken(ctx, vssNodeIdentifier)
 		if err != nil {
 			logger.Logger.WithError(err).Error("Failed to fetch VSS JWT token")
+
+			existingVssToken, _ := svc.cfg.Get("VssToken", "")
```
bumi, line 453: so this means if the Alby API does not return the VSS token then it will use the existing one. 

I don't fully get the reasoning here. The existing would also expire and auth in the VSS server would fail, wouldn't it? 

wo we need this? or should the API handle it?

rolznz: It will eventually expire, but the old token can be used since it has a 1 year expiry (or whatever we change it to), giving the user time to do something.

Right now we do not lock down the API at all and if we do we will instantly break any hubs of users who cancelled their subscription.

bumi: imo 1 year is too long. it should be more like weeks only if they don't pay. 

don't you think we can decide this in the token API and give users there a grace period? 

but either way this is actually good because if the API is down the wallet should still work. 🤔

rolznz: @bumi we handle most things OK and most of the wallet is usable even if the alby API is down, I just added a few extra improvements here: https://github.com/getAlby/hub/pull/1229

PR #1243, opened 2025-04-05
chore: use debian:12-slim base image for smaller final container

k0gen:
Using `debian:12-slim` instead of `debian` reduces the final container image size while maintaining all required functionality. This helps optimize resource usage and improves pull/push times.


rolznz (approved):
tACK

PR #1254, opened 2025-04-09
fix: sync passwordinput with external value change

im-adithya:
## Description

When a Start attempt fails (for some node reason let's say) and you try again, you get this weird "invalid password" error even though you haven't changed anything.

It was happening because on receiving an error, the unlockPassword is set to `""` but it is not reflected in the internal state of `PasswordInput` component which makes it render the password without clearing it. This PR syncs it so the component state is updated when value changes externally.

### DIFF
--- a/frontend/src/components/password/PasswordInput.tsx
+++ b/frontend/src/components/password/PasswordInput.tsx
@@ -16,21 +16,14 @@ export default function PasswordInput({
   value,
   ...restProps
 }: PasswordInputProps) {
-  const [password, setPassword] = React.useState(value);
   const [passwordVisible, setPasswordVisible] = React.useState(false);
-
-  React.useEffect(() => {
-    if (onChange) {
-      onChange(password);
-    }
-  }, [password, onChange]);
 
   return (
     <Input
       type={passwordVisible ? "text" : "password"}
-      value={password}
+      value={value}
       required
-      onChange={(e) => setPassword(e.target.value)}
+      onChange={(e) => onChange && onChange(e.target.value)}
       placeholder={placeholder}
       {...restProps}
       endAdornment={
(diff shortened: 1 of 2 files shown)


frontend/src/components/password/PasswordInput.tsx
```diff
@@ -19,6 +19,10 @@ export default function PasswordInput({
   const [password, setPassword] = React.useState(value);
   const [passwordVisible, setPasswordVisible] = React.useState(false);
 
+  React.useEffect(() => {
```
rolznz, line 22: This seems like a hack to me, it shouldn't be needed if indeed `unlockPassword` is changed because it's passed in which will cause a re-render

rolznz: Can you double check if your assumptions are right about the root cause?

im-adithya: Yup, this is the root cause. Prop change doesn't necessarily mean a re-render right? Since the state has been initialized here and changing the `value` prop doesn't change the `password` state variable

im-adithya: But I agree this looks like a hack, I'll try to remove the internal state because even the other useEffect looks bad in this component (can't we just call onChange directly?)

im-adithya: Done

frontend/src/components/password/PasswordInput.tsx
```diff
@@ -16,21 +16,14 @@ export default function PasswordInput({
...
-  }, [password, onChange]);
-
   return (
     <Input
       type={passwordVisible ? "text" : "password"}
-      value={password}
+      value={value}
       required
-      onChange={(e) => setPassword(e.target.value)}
+      onChange={(e) => onChange && onChange(e.target.value)}
```
rolznz, line 26: can we make `onChange` required too?

rolznz: Ah, true. Thanks for checking

rolznz (approved):
utACK

PR #1265, opened 2025-04-16
fix: update workflows to use ubuntu 24.04

im-adithya:
See https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/.github/workflows/http.yml
+++ b/.github/workflows/http.yml
@@ -31,17 +31,17 @@ jobs:
             {
               arch: x86_64,
               alby_arch: x86_64-unknown-linux-gnu,
-              os: ubuntu-20.04,
+              os: ubuntu-24.04,
             },
             {
               arch: armv6,
               alby_arch: arm-unknown-linux-gnueabihf,
-              os: ubuntu-20.04,
+              os: ubuntu-24.04,
             },
             {
               arch: aarch64,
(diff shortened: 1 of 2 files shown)


.github/workflows/wails.yml
```diff
@@ -31,7 +31,6 @@ jobs:
       matrix:
         build:
           [
-            { platform: linux/amd64, os: ubuntu-20.04 },
```
rolznz, line 34: this removes the Ubuntu 22.04 Wails build, can you make sure this line is left in but change it to 22.04?

.github/workflows/http.yml
```diff
@@ -31,17 +31,17 @@ jobs:
             {
               arch: x86_64,
               alby_arch: x86_64-unknown-linux-gnu,
-              os: ubuntu-20.04,
+              os: ubuntu-24.04,
```
rolznz, line 34: if we do this, will the build still work on 22.04?

.github/workflows/wails.yml
im-adithya: Done!

.github/workflows/http.yml
im-adithya: Whoops, missed that 22.04 is still supported

rolznz (approved):
utACK

PR #1266, opened 2025-04-16
feat: auto-swaps

### LINKED ISSUE #1303: LDK SendPaymentSync will not time out after 50 seconds if there is no other LDK event in the meantime
See https://github.com/getAlby/hub/blob/master/lnclient/ldk/ldk.go#L488-L489

I think we should do a select to either take the next event or a timeout occurs, like we do here: https://github.com/getAlby/hub/blob/feat/hold-invoices-ldk/lnclient/ldk/ldk_event_broadcaster.go#L120-L129

reneaaron:
Also fixes #1303

## TODOs

- [x] Do final testing with LDK/LND on mainnet
- [x] Improve fee display on the frontend, currently looks scary
- [x] ~Shift Swaps from settings~ (Agreeed to do in follow-up)
- [x] ~Add option to do one-time swaps~ (Agreeed to do in follow-up)

### Minor
- [x] `LDK_NETWORK` config var should now be `NETWORK`
- [x] ~Use LNClient method to publish transactions~ (Agreeed to do in follow-up)

### Testing
- [ ] Phoenix
- [ ] Cashu
- [x] Wails


swaps/swaps_service.go
```diff
@@ -0,0 +1,250 @@
...
+				lightningBalance := uint64(balance.Lightning.TotalSpendable)
+				balanceThresholdMilliSats := parsedBalanceThreshold * 1000
+				if lightningBalance >= balanceThresholdMilliSats {
+					// TODO: Change this calcuation
+					amount := lightningBalance - balanceThresholdMilliSats
+					logger.Logger.WithFields(logrus.Fields{
+						"amount":      amount,
+						"destination": swapDestination,
+					}).Info("Initiating swap")
+					// TODO: Should we ourselves add a check that the amount is < 50000
```
im-adithya, line 82: Mention this in frontend

swaps/swaps_service.go
```diff
@@ -0,0 +1,250 @@
...
+					err := svc.ReverseSwap(ctx, amount/1000, swapDestination, lnClient)
```
im-adithya, line 83: Should this be in a goroutine since it blocks calls till this fails/succeeds (I think it's good to block)

swaps/swaps_service.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	preimage := make([]byte, 32)
+	_, err = rand.Read(preimage)
+	if err != nil {
+		return err
+	}
+	preimageHash := sha256.Sum256(preimage)
+
+	boltzApi := &boltz.Api{URL: svc.cfg.GetEnv().BoltzApi}
```
jackstar12, line 134: You might want to reuse the api and websocket structs across swaps, although its not that big of a deal since you are not swapping very frequently

swaps/swaps_service.go
```diff
@@ -0,0 +1,270 @@
...
+						"swap":   swap,
+						"update": update,
+					}).Error("Error paying the swap invoice")
+					return
+				}
+				logger.Logger.WithField("transaction", transaction).Info("Swap payment succeeded")
+			}()
+			break
+
+		case boltz.TransactionMempool:
```
jackstar12, line 204: You are currently always accepting 0-conf from boltz, which leaves you to the potential risk of boltz replacing its lockup transaction after you revealed the preimage by broadcasting the claim transaction. Since the swaps arent exposed in any user interface you can always wait for `boltz.TransactionConfirmed` to make sure boltz could not replace the transaction. You can still leave a log statement for `boltz.TransactionMempool` but only claim once its confirmed.
```suggestion
		case boltz.TransactionMempool:
```
```suggestion
		case boltz.TransactionConfirmed:
```

swaps/swaps_service.go
```diff
@@ -0,0 +1,270 @@
...
+			if err != nil {
+				return fmt.Errorf("could not create claim transaction: %w", err)
+			}
+
+			txHex, err := claimTransaction.Serialize()
+			if err != nil {
+				return fmt.Errorf("could not serialize claim transaction: %w", err)
+			}
+
+			txId, err := boltzApi.BroadcastTransaction(boltz.CurrencyBtc, txHex)
```
jackstar12, line 244: While using the Boltz API here is convenient, it's recommended to use your own BTC / LN node or third-party API for broadcasting the claim transaction to minimize trust in boltz.

api/api.go
```diff
@@ -543,6 +544,72 @@ func (api *api) GetNodeConnectionInfo(ctx context.Context) (*lnclient.NodeConnec
 	return api.svc.GetLNClient().GetNodeConnectionInfo(ctx)
 }
 
+func (api *api) GetAutoSwaps(ctx context.Context) (*GetAutoSwapsResponse, error) {
```
rolznz, line 547: This is more about the auto swaps configuration? would it make sense to rename? (GetAutoSwaps felt like it was fetching swaps that had been done)

api/api.go
```diff
@@ -543,6 +544,72 @@ func (api *api) GetNodeConnectionInfo(ctx context.Context) (*lnclient.NodeConnec
...
+	err = api.cfg.SetUpdate(config.AutoSwapDestinationKey, enableAutoSwapsRequest.Destination, "")
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to save autoswap destination to config")
+		return err
+	}
+
+	return api.svc.StartAutoSwaps()
+}
+
+func (api *api) DeleteAutoSwaps() error {
```
rolznz, line 594: Why delete and not disable?

service/models.go
```diff
@@ -8,19 +8,23 @@ import (
...
 	"github.com/getAlby/hub/lnclient"
 	"github.com/getAlby/hub/service/keys"
+	"github.com/getAlby/hub/swaps"
 	"github.com/getAlby/hub/transactions"
 )
 
 type Service interface {
 	StartApp(encryptionKey string) error
 	StopApp()
+	StartAutoSwaps() error
```
rolznz, line 18: should these two methods be inside the SwapsService?

service/stop.go
```diff
@@ -11,11 +11,16 @@ func (svc *service) StopApp() {
 	if svc.appCancelFn != nil {
 		logger.Logger.Info("Stopping app...")
 		svc.appCancelFn()
+		svc.swapsService.StopAutoSwaps()
```
rolznz, line 14: What happens if a swap is in progress when we shut down? there's no issue, right?

service/models.go
rolznz: Edit: maybe they can just be removed here

http/http_service.go
```diff
@@ -1140,3 +1143,44 @@ func (httpSvc *HttpService) healthHandler(c echo.Context) error {
...
+
+func (httpSvc *HttpService) enableAutoSwapsHandler(c echo.Context) error {
+	var enableAutoSwapsRequest api.EnableAutoSwapsRequest
+	if err := c.Bind(&enableAutoSwapsRequest); err != nil {
+		return c.JSON(http.StatusBadRequest, ErrorResponse{
+			Message: fmt.Sprintf("Bad request: %s", err.Error()),
+		})
+	}
+
+	err := httpSvc.api.EnableAutoSwaps(c.Request().Context(), &enableAutoSwapsRequest)
```
rolznz, line 1166: @im-adithya I saw you added a fix https://github.com/getAlby/hub/pull/1270/commits/86e5f47732bf3761165d27867c41784f6b51ff7b but isn't it that we use the HTTP request context rather than the service context if we enable it here?

im-adithya: I didn't get you, we can't use the HTTP request context to enable auto-swaps right? Because it has to keep running?

The fix makes ure we pass in a short lived context to make it work, since during the swap we don't have the http context (it is killed as soon as the swaps are enabled) and have to use the context of the swap service.

(I think you got confused because of the same naming, the api.EnableAutoSwaps is different from swap service EnableAutoSwaps method, can you suggest good naming 😬)

rolznz: > The fix makes sure we pass in a short lived context to make it work, since during the swap we don't have the http context (it is killed as soon as the swaps are enabled) and have to use the context of the swap service.

This is the thing - that fix should not be needed if you are using the service context, because it won't be cancelled until the hub is shut down. Can you check you're using the correct context?

service/stop.go
im-adithya: Added a commit to make sure we also stop the boltz websocket, should be good now

swaps/swaps_service.go
im-adithya: Nice, thanks!

lnclient/ldk/ldk.go
```diff
@@ -435,7 +435,7 @@ func getMaxTotalRoutingFeeLimit(amountMsat uint64) ldk_node.MaxTotalRoutingFeeLi
 	}
 }
 
-func (ls *LDKService) SendPaymentSync(ctx context.Context, invoice string, amount *uint64) (*lnclient.PayInvoiceResponse, error) {
+func (ls *LDKService) SendPaymentSync(ctx context.Context, invoice string, amount *uint64, isHoldInvoice bool) (*lnclient.PayInvoiceResponse, error) {
```
rolznz, line 438: I'm not sure this is the best way to do it. This method is `SendPaymentSync` but now we are basically doing an async payment, right?

im-adithya:
Should be ready now 🎉

.env.example
```diff
@@ -35,4 +35,7 @@ FRONTEND_URL=http://localhost:5173
 #LND_ADDRESS=127.0.0.1:10001
 #LND_MACAROON_FILE=/home/YOUR_USERNAME/.polar/networks/1/volumes/lnd/alice/data/chain/bitcoin/regtest/admin.macaroon
 
-#LDK_VSS_URL="http://localhost:8090/vss"
\ No newline at end of file
+#LDK_VSS_URL="http://localhost:8090/vss"
+
+# Boltz API
+#BOLTZ_API=https://api.testnet.boltz.exchange
```
rolznz, line 41: can you add this to the README?

config/models.go
```diff
@@ -44,6 +47,7 @@ type AppConfig struct {
 	EnableAdvancedSetup      bool   `envconfig:"ENABLE_ADVANCED_SETUP" default:"true"`
 	AutoUnlockPassword       string `envconfig:"AUTO_UNLOCK_PASSWORD"`
 	LogDBQueries             bool   `envconfig:"LOG_DB_QUERIES" default:"false"`
+	BoltzApi                 string `envconfig:"BOLTZ_API" default:"https://api.boltz.exchange"`
```
rolznz, line 50: Should we address the NETWORK change mentioned in the PR?

We could remove the `LDK_NETWORK` default but still keep that config, and also add `NETWORK` (also without a default).

Then when initializing the config, check if either are set and use that, otherwise set the value to "bitcoin".

lnclient/ldk/ldk.go
```diff
@@ -485,59 +485,68 @@ func (ls *LDKService) SendPaymentSync(ctx context.Context, invoice string, amoun
 	fee := uint64(0)
 	preimage := ""
 
-	for start := time.Now(); time.Since(start) < time.Second*50; {
-		event := <-ldkEventSubscription
-
-		eventPaymentSuccessful, isEventPaymentSuccessfulEvent := (*event).(ldk_node.EventPaymentSuccessful)
-		eventPaymentFailed, isEventPaymentFailedEvent := (*event).(ldk_node.EventPaymentFailed)
+	timeout := time.Second * time.Duration(timeoutSeconds)
```
rolznz, line 488: why implement it this way compared to just adding a simple case:
`case <-time.After(time * time.Second):`

Then we don't need a new timeout or the `if` do we inside `case <-ctx.Done():`?

It would be good to have 3 different cases:
- `ctx.Done` (the original context)
- `case <-time.After(time * time.Second):`
- `case ev := <-ldkEventSubscription:`

What do you think?

api/transactions.go
```diff
@@ -65,7 +65,7 @@ func (api *api) SendPayment(ctx context.Context, invoice string, amountMsat *uin
 	if api.svc.GetLNClient() == nil {
 		return nil, errors.New("LNClient not started")
 	}
-	transaction, err := api.svc.GetTransactionsService().SendPaymentSync(ctx, invoice, amountMsat, nil, api.svc.GetLNClient(), nil, nil)
+	transaction, err := api.svc.GetTransactionsService().SendPaymentSync(ctx, invoice, amountMsat, nil, api.svc.GetLNClient(), nil, nil, 50)
```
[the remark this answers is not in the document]
im-adithya: Yup that sounds much better, changed!

.env.example
im-adithya: Done

lnclient/ldk/ldk.go
im-adithya: Yeah but that would reset the timeout everytime you get an event. I changed it to use a timer instead, should be good now :+1:

config/models.go
im-adithya: Done!

lnclient/ldk/ldk.go
```diff
@@ -435,7 +435,12 @@ func getMaxTotalRoutingFeeLimit(amountMsat uint64) ldk_node.MaxTotalRoutingFeeLi
 	}
 }
 
-func (ls *LDKService) SendPaymentSync(ctx context.Context, invoice string, amount *uint64, timeoutSeconds int64) (*lnclient.PayInvoiceResponse, error) {
+func (ls *LDKService) SendPaymentSync(ctx context.Context, invoice string, amount *uint64, timeoutSeconds *int64) (*lnclient.PayInvoiceResponse, error) {
+	sendPaymentTimeout := int64(50)
```
rolznz, line 439: Can we move 50 to a constant inside constants.go?

lnclient/ldk/ldk.go
```diff
@@ -491,20 +491,19 @@ func (ls *LDKService) SendPaymentSync(ctx context.Context, invoice string, amoun
 	preimage := ""
 
 	timeout := time.Second * time.Duration(*timeoutSeconds)
-	ctx, cancel := context.WithTimeout(ctx, timeout)
-	defer cancel()
+	timer := time.NewTimer(timeout)
```
rolznz, line 494: why do it this way compared to putting the timer directly in the `case`?

lnclient/ldk/ldk.go
rolznz, line 439: can we move 50 to a constant in `constants.go`?

lnclient/ldk/ldk.go
rolznz, line 494: why add a timer here compared to using it directly in the case like it's done in other places?

lnclient/ldk/ldk.go
```diff
@@ -485,7 +490,7 @@ func (ls *LDKService) SendPaymentSync(ctx context.Context, invoice string, amoun
 	fee := uint64(0)
 	preimage := ""
 
-	timeout := time.Second * time.Duration(timeoutSeconds)
+	timeout := time.Second * time.Duration(*timeoutSeconds)
```
rolznz, line 493: this will crash

im-adithya: Why? I tested by replacing 50 above with 1 and it worked for me on LDK (paid refund@lnurl.mutinynet.com)

lnclient/ldk/ldk.go
im-adithya: Because if we get any event on LDK which doesn't cause `SendPaymentSync` to return, it would go via the for loop again, checking 50s from the time it looped again, extending the actual timeout...

lnclient/ldk/ldk.go
im-adithya: Done

rolznz (approved):
utACK

lnclient/ldk/ldk.go
rolznz: Ah right, thanks!

PR #1274, opened 2025-04-23
feat: improve querying performance

### LINKED ISSUE #1263: Review DB indexes and speed up slow queries
For our cloud hubs we have noticed some warnings of slow (> 200ms) queries from gorm. 

Do we miss some indexes?

- Counting the number of transactions (`transactions/transactions_service.go:591`), and fetching transactions (`transactions/transactions_service.go:605`), fetching app (`transactions/transactions_service.go:578`) (25% of slow queries)
- Persisting a request event (`nip47/event_handler.go:52`) and updating it (`event_handler.go:165`) (15% of slow queries)
- Fetching user config by a particular key (`config.go:155`) (5%)
- Listing apps from the http API (`api.go:335`) (5%)
- Persisting response event (`event_handler.go:464-496`) (5%)
- Looking up transaction for app id (`transactions_service.go:493`) (5%)

im-adithya:
Fixes #1263

Leaves out a few things mentioned in the issue:
- Persisting a request event and response event (updates are covered though)
- Fetching user config by a particular key (key is already indexed and are very less, not sure why this is happening)
- Listing apps from the http API, the ideal solution is to have the permissions within the apps table, we should do a migration for that)

## Indexes removed

- `idx_transactions_app_id` (we never filter only on app_id, added composite indexes to cover all cases)
- `idx_transactions_request_event_id` (same as above)
- `idx_transactions_created_at` (same as above)
- `idx_transactions_settled_at` (same as above)
- `idx_transactions_app_id_type_state_created_at_settled_at_payment_hash` (we should always use most differing column first, so it can use the index when required, and payment_hash is at the end, also we never filter based on created and settled at dates, so we can remove it)

## Indexes added

- `idx_transactions_state_type` (used while listing txs)
- `idx_transactions_state_type_updated_at` (used while listing txs)
- `idx_transactions_app_id_state_type_updated_at` (used while listing txs)
- `idx_transactions_payment_hash_settled_at_created_at` (for lookup transaction)

### List Transactions - Possible queries

- We always chain on `state` and `type`
- `updated_at` is not used much, unless it's via NWC controller
- Also pushed `updated_at desc` ordering to do it after counting to save some time
- Conditionally `app_id` (I took out code to find `isIsolated` outside to reduce multiple requests to db at once)

So for all these possibilities, I added `idx_transactions_state_type`, `idx_transactions_state_type_updated_at` and `idx_transactions_app_id_state_type_updated_at` (`idx_transactions_app_id_type_state` already exists)

### Lookup Transactions - Possible queries

- We don't use `type` at all - both internally and via NWC, we should remove the param itself in a separate issue
- We do a find query on `payment_hash` by ordering by `settled_at` and `created_at`
- Conditionally `app_id` (I took out code to find `isIsolated` outside to reduce multiple requests to db at once)

And this time app_id doesn't make much difference so I only added `idx_transactions_payment_hash_settled_at_created_at`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -152,7 +152,7 @@ func (cfg *config) Get(key string, encryptionKey string) (string, error) {
 
 func (cfg *config) get(key string, encryptionKey string, gormDB *gorm.DB) (string, error) {
 	var userConfig db.UserConfig
-	err := gormDB.Where(&db.UserConfig{Key: key}).Limit(1).Find(&userConfig).Error
+	err := gormDB.Where("key = ?", key).First(&userConfig).Error
 	if err != nil {
 		return "", fmt.Errorf("failed to get configuration value: %w", gormDB.Error)
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -152,7 +152,7 @@ func (cfg *config) Get(key string, encryptionKey string) (string, error) {
 
 func (cfg *config) get(key string, encryptionKey string, gormDB *gorm.DB) (string, error) {
 	var userConfig db.UserConfig
-	err := gormDB.Where("key = ?", key).First(&userConfig).Error
+	err := gormDB.Where(&db.UserConfig{Key: key}).Limit(1).Find(&userConfig).Error
 	if err != nil {
 		return "", fmt.Errorf("failed to get configuration value: %w", gormDB.Error)
 	}
--- a/db/migrations/202504231037_add_indexes.go
+++ b/db/migrations/202504231037_add_indexes.go
@@ -14,7 +14,6 @@ var _202504231037_add_indexes = &gormigrate.Migration{
 		if err := db.Transaction(func(tx *gorm.DB) error {
 
 			if err := tx.Exec(`
-DROP INDEX IF EXISTS idx_transactions_app_id;
 DROP INDEX IF EXISTS idx_transactions_request_event_id;
 DROP INDEX IF EXISTS idx_transactions_created_at;
 DROP INDEX IF EXISTS idx_transactions_settled_at;
--- a/db/migrations/202504231037_add_indexes.go
+++ b/db/migrations/202504231037_add_indexes.go
@@ -1,0 +1,45 @@
+package migrations
+
+import (
+	_ "embed"
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+var _202504231037_add_indexes = &gormigrate.Migration{
+	ID: "202504231037_add_indexes",
+	Migrate: func(db *gorm.DB) error {
+
+		if err := db.Transaction(func(tx *gorm.DB) error {
+
+			if err := tx.Exec(`
+DROP INDEX IF EXISTS idx_transactions_app_id;
+DROP INDEX IF EXISTS idx_transactions_request_event_id;
+DROP INDEX IF EXISTS idx_transactions_created_at;
+DROP INDEX IF EXISTS idx_transactions_settled_at;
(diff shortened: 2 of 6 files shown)


db/migrations/202504231037_add_indexes.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+var _202504231037_add_indexes = &gormigrate.Migration{
+	ID: "202504231037_add_indexes",
+	Migrate: func(db *gorm.DB) error {
+
+		if err := db.Transaction(func(tx *gorm.DB) error {
+
+			if err := tx.Exec(`
+DROP INDEX IF EXISTS idx_transactions_app_id;
```
reneaaron, line 17: These are not being used currently?

config/config.go
```diff
@@ -152,7 +152,7 @@ func (cfg *config) Get(key string, encryptionKey string) (string, error) {
 
 func (cfg *config) get(key string, encryptionKey string, gormDB *gorm.DB) (string, error) {
 	var userConfig db.UserConfig
-	err := gormDB.Where(&db.UserConfig{Key: key}).Limit(1).Find(&userConfig).Error
+	err := gormDB.Where("key = ?", key).First(&userConfig).Error
```
rolznz, line 155: `First` returns an error if the key is not found, which will lead us to return errors in a lot of places the code currently does not expect.

db/migrations/202504231037_add_indexes.go
rolznz: @im-adithya how did you confirm the indexes are not used?

db/migrations/202504231037_add_indexes.go
```diff
@@ -0,0 +1,45 @@
...
+DROP INDEX IF EXISTS idx_transactions_request_event_id;
+DROP INDEX IF EXISTS idx_transactions_created_at;
+DROP INDEX IF EXISTS idx_transactions_settled_at;
+DROP INDEX IF EXISTS idx_transactions_app_id_type_state_created_at_settled_at_payment_hash;
```
rolznz, line 21: for example this one is removed, but what is it replaced with for list transactions?

rolznz: should this just be replaced with updated_at instead of `created_at_settled_at`?

im-adithya: `idx_transactions_app_id_type_state_created_at_settled_at_payment_hash` is not used for `ListTransactions`, since there we never use payment hash, I think this was for `LookupInvoice` only

I added `idx_transactions_payment_hash_settled_at_created_at ` instead of this, since app_id doesn't differ much in the table and an index on these three should work better.

db/migrations/202504231037_add_indexes.go
im-adithya: I confirmed by seeing the possible queries from `transaction_service.go` etc.

For example, we never filter based on `idx_transactions_created_at`, `idx_transactions_settled_at` or `idx_transactions_request_event_id` so I removed those (and also as mentioned above, we don't need single indexes as most of our queries are chained involving more than one column so single col indexes won't help anyways)

db/migrations/202504231037_add_indexes.go
rolznz: > since app_id doesn't differ much in the table
@im-adithya what do you mean by this?

rolznz: But I saw you added the app id back in, thanks

im-adithya:
I tried populating my db with some 2000 txs but it still gets all transactions in 0.01/0.02s so not able to test much. But it must be the same in real world situations, I think the main reason must be because it has too many things happening at once which causes gorm to slow down. Good to monitor :+1:

PR #1276, opened 2025-04-25
fix: use md breakpoint for sidebar trigger

im-adithya:
Because mobile breakpoint is `768px`

https://github.com/getAlby/hub/blob/3ccdb1e16196bd62ba8463ba400fc0338fce3b6a/frontend/src/hooks/use-mobile.tsx#L3


reneaaron (approved):
Nice catch! :100: 

utACK

Issue #1272, opened 2025-04-23
Bug: possible to mark a failed payment as settled when a successful re-try works


rolznz:
The Alby Hub transaction table allows for multiple payment attempts for the same invoice and will show up as different entries. e.g. You failed to pay an invoice, and 10 minutes later you try again and it succeeds, you should have one failed payment and one successful payment in your transaction list.

I noticed recently I had this issue, and re-trying failed payment and successfully paying marks both outgoing transaction entries as settled rather than only the second attempt. This seems again to be a postgres issue that the transactions are not executed sequentially.

This will be a problem for isolated sub-wallets because their balance will decrease twice for the same invoice.

I think we need a better solution to ensure our transactions work the same in both Sqlite and postgres.

rolznz:
@rdmitr nice catch, this sounds like it could be the issue.

rolznz:
@rdmitr any chance you could create a PR with a corresponding test for this case just to confirm it and ensure the fix works? or possibly @im-adithya ?

PR #1278, opened 2025-04-28
fix: filter transactions by state when processing the payment_sent event in the transactions service

rdmitr:
Fixes https://github.com/getAlby/hub/issues/1272


rolznz (approved):
tACK

Nice fix, thanks @rdmitr !

PR #1279, opened 2025-04-28
fix: overflow due to long descriptions

pavanjoshi914:
before: 
image
image

after: 

image
image

### DIFF
--- a/frontend/src/components/TransactionItem.tsx
+++ b/frontend/src/components/TransactionItem.tsx
@@ -144,7 +144,7 @@ function TransactionItem({ tx }: Props) {
                 </span>
               </p>
             </div>
-            <p className="text-sm md:text-base text-muted-foreground break-all w-full truncate">
+            <p className="text-sm md:text-base text-muted-foreground break-all line-clamp-1">
               {tx.description}
             </p>
           </div>
--- a/frontend/src/screens/wallet/send/ConfirmPayment.tsx
+++ b/frontend/src/screens/wallet/send/ConfirmPayment.tsx
@@ -83,7 +83,7 @@ export default function ConfirmPayment() {
           <FormattedFiatAmount amount={amount || invoice.satoshi} />
         </div>
         {invoice.description && (
-          <div className="mt-2 max-w-lg break-words">
+          <div className="mt-2 break-all">
             <Label>Description</Label>
             <p className="text-muted-foreground">{invoice.description}</p>
           </div>
--- a/frontend/src/screens/wallet/send/ConfirmPayment.tsx
+++ b/frontend/src/screens/wallet/send/ConfirmPayment.tsx
@@ -83,7 +83,7 @@ export default function ConfirmPayment() {
           <FormattedFiatAmount amount={amount || invoice.satoshi} />
         </div>
         {invoice.description && (
-          <div className="mt-2">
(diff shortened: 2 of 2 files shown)


frontend/src/screens/wallet/send/ConfirmPayment.tsx
```diff
@@ -83,7 +83,7 @@ export default function ConfirmPayment() {
           <FormattedFiatAmount amount={amount || invoice.satoshi} />
         </div>
         {invoice.description && (
-          <div className="mt-2">
+          <div className="mt-2 max-w-lg break-words">
```
reneaaron, line 86: Why `max-w-lg`? https://tailwindcss.com/docs/max-width

reneaaron: Shouldn't that be limited by the container (instead of setting it explicitly)

pavanjoshi914: true. sendLayout already uses max-w-lg. depending on the text we have to use break-all instead of break-words to make it work without max-w-lg setting again at the <p> level. fixed

frontend/src/components/TransactionItem.tsx
```diff
@@ -144,7 +144,7 @@ function TransactionItem({ tx }: Props) {
                 </span>
               </p>
             </div>
-            <p className="text-sm md:text-base text-muted-foreground break-all w-full truncate">
+            <p className="text-sm md:text-base text-muted-foreground break-all line-clamp-1">
```
reneaaron, line 147: I don't think break-all is the behavior we want, since it breaks words inline, regardless if they could be moved to the next line: 

https://tailwindcss.com/docs/word-break

Could we give this a try instead? 

https://tailwindcss.com/docs/overflow-wrap#wrapping-mid-word

pavanjoshi914: image

sounds good. looks a bit inconsistent where it breaks though. depends solely on the text passed. some text introduced ellipsis earlier and some at the last

pavanjoshi914: i used break all so that if continuous texts are passed it breaks at the end. for non continous texts break-all and break-words doesn't make much difference

im-adithya: I think the problem is because of the sidebar. If we use `break-word`, it assumes the entire viewport width and cuts it off once that width is exceeded.

But `break-all` cuts it off according to available width and is better in this scenario

<table>

<tr>
<td>
break-word
</td>
<td>
break-all
</td>
</tr>

<tr>
<td>
<video src="https://github.com/user-attachments/assets/d17b5830-1ec5-4902-969b-4a595f5bcaa3" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/5cfb9408-d016-4934-b20c-8f6de4d0c588" />
</td>
</tr>

</table>

im-adithya (approved):
tACK

frontend/src/components/TransactionItem.tsx
rolznz: @reneaaron applied a similar fix to the messageboard widget on the home page using `break-anywhere`

PR #1280, opened 2025-04-28
fix: use full version of go library

reneaaron:
Fixes https://stackoverflow.com/questions/78519711/toolchain-not-available-error-prevents-me-from-using-any-go-commands


im-adithya:
Closing because we are anyways upgrading to `1.24.1` in https://github.com/getAlby/hub/pull/1259/files

PR #1287, opened 2025-04-30
chore: increase node shutdown timeout

### LINKED ISSUE #1253: Increase LDK-node shutdown timeout
Currently we force the LDK node to shutdown after 2 minutes which if it takes longer than that can cause the channel manager to become out of sync with the channel monitor, and result in force closures on startup.

This is possible to happen if the user has a pending payment when they shut down the node.

im-adithya:
Fixes #1253

### DIFF
--- a/fly.toml
+++ b/fly.toml
@@ -6,7 +6,7 @@
 app = 'nwc'
 primary_region = 'lax'
 swap_size_mb = 2048
-kill_timeout = 120
+kill_timeout = 300
 
 [build]
   image = 'ghcr.io/getalby/hub:latest'
--- a/fly.toml
+++ b/fly.toml
@@ -6,6 +6,7 @@
 app = 'nwc'
 primary_region = 'lax'
 swap_size_mb = 2048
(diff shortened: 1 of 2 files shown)


fly.toml
```diff
@@ -6,7 +6,7 @@
 app = 'nwc'
 primary_region = 'lax'
 swap_size_mb = 2048
-kill_timeout = 120
+kill_timeout = 300
```
rolznz, line 9: I noticed, this should not be exactly the same as the LDK timeout otherwise the hub could be killed while it's shutting down. Could we change it to 330?

Also can you do a PR inside getalby.com?

fly.toml
```diff
@@ -6,7 +6,7 @@
 app = 'nwc'
 primary_region = 'lax'
 swap_size_mb = 2048
-kill_timeout = 120
+kill_timeout = 330
```
rolznz, line 9:
```suggestion
# Add a kill timeout 30s longer than LDK node shutdown timeout
kill_timeout = 330
```

Issue #1291, opened 2025-05-01
List On-chain transactions


rolznz:
With LDK-node 0.5 we can now easily list on-chain transactions. But there are not much context around them yet (in the list below are a mix of open outbound channel TXs, close channel TXs, standard on-chain send and receive transactions)

TODOs:
- [x] Ask LDK team for guidance on how to add this context (Update: needs changes inside LDK-node and rust lightning, see https://github.com/lightningdevkit/rust-lightning/issues/3566)

```json
[
{"level":"info","msg":"Onchain payment!","onchain":{"Txid":"1ac3470dee4c4ed2c8f793c37d2ff68dbc465159f77ec05860e68bfa75e61b82","Status":{"BlockHash":"000001d90d009eac2974fbdfef950beff97204ab800cf25919189ef65c01c1ac","Height":2067495,"Timestamp":1746079231}},"payment":{"Id":"821be675fa8be66058c07ef7595146bc8df62f7dc393f7c8d24e4cee0d47c31a","Kind":{"Txid":"1ac3470dee4c4ed2c8f793c37d2ff68dbc465159f77ec05860e68bfa75e61b82","Status":{"BlockHash":"000001d90d009eac2974fbdfef950beff97204ab800cf25919189ef65c01c1ac","Height":2067495,"Timestamp":1746079231}},"AmountMsat":500000000,"FeePaidMsat":294000,"Direction":2,"Status":2,"CreatedAt":0,"LatestUpdateTimestamp":1746079410},"time":"2025-05-01T14:47:11+07:00"},
[...]

PR #1292, opened 2025-05-02
Feat: Add Onchain transactions to node page

rolznz:
Closes https://github.com/getAlby/hub/issues/1291
Note: createdAt is not set right now, only updatedAt is used in the UI. (See https://github.com/getAlby/ldk-node/issues/69). It's a custom field implemented by Alby inside ldk-node (maybe we don't need it really)

TODOs:
- [x] ~Review UI~ (we agreed to address this in a follow-up)
- [x] Remove UpdatedAt field
- [x] Update LDK-node version (fixes CreatedAt field, in case we want it in the future)

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1035,6 +1035,12 @@ func (api *api) SyncWallet() error {
 	api.svc.GetLNClient().UpdateLastWalletSyncRequest()
 	return nil
 }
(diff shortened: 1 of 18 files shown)


im-adithya (approved):
tACK

Issue #1212, opened 2025-03-21
Info event publish can skip if relay disconnects after consuming event from channel


im-adithya:
Follow-up issue to https://github.com/getAlby/hub/pull/1187

>There is a small risk still that it takes an item from the channel but then the relay disconnects while it's being processed.

We can add it back to the queue in those cases, but there is [this](https://github.com/getAlby/hub/blob/ee5ba98037cbd9e28ce15207ef1b327bc82af541/nip47/publish_nip47_info.go#L59) and other errors because of which it can fail leading in an infinite loop so we have to be a bit more careful there.

Not high priority since
1) the chance of that happening is rare
2) only effects the publish of one event
3) it is just a re-publish and the app most probably already has an updated event in the relay

im-adithya:
Would it be good to accumulate failed events in memory and have a go routine in StartNotifier which runs in parallel to keep publishing those?

(we can actually just store the *nostr.Event once it fails [here](https://github.com/getAlby/hub/blob/3a5c989cf2964019be815bb087f7a697b723726a/nip47/notifications/nip47_notifier.go#L186) and try publishing it again)

rolznz:
I'm not sure it's good to create something completely separate. Could the item be re-added to the channel to be re-tried? (as long as we can be sure we do not continue to try republishing the event forever if for some reason the relay doesn't accept it)

PR #1298, opened 2025-05-06
feat: add hold invoices support for LND & LDK

### LINKED ISSUE #1295: Hold invoices
Add support for generating hold invoices using LND and LDK backends.
Would require new NWC methods `make_hold_invoice`, `settle_hold_invoice` & `cancel_hold_invoice` and new notification `hold_invoice_accepted` to inform the client when the invoice has been accepted and is waiting to be settled or cancelled.
Would need also to add these changes to NIP-47 spec.

frnandu:
fixes #1295 
- [x] NWC new methods 
- [x] new `hold_invoice_accepted` notification
- [x] LND implementation
- [x] LDK implementation
- [x] handle self-payments -> https://github.com/getAlby/hub/pull/1304
- [x] Add hold invoice support to JS SDK (@rolznz to do) -> https://github.com/getAlby/js-sdk/pull/382
- [X] Draft NIP-47 spec PR -> https://github.com/nostr-protocol/nips/pull/1913
- [x] expiry does not need to be mandatory any more in the spec (add more documentation regarding the expiry field, so that it's more obvious to the developer that it will not guarantee the receiver to be able to settle the invoice and that it should check the deadline timestamp in the notification.)
- [x] return claim deadline for LDK and LND + documentation about why it's in blocks
- [x] review if we need to lower the inactive background sync time for LDK (currently 1h) if it could cause issues with hold invoices having a shorter than expected claim deadline - update: I think it's ok for now

### DIFF
--- a/transactions/self_hold_payments_test.go
+++ b/transactions/self_hold_payments_test.go
@@ -33,15 +33,17 @@ func TestSelfHoldPaymentSettled(t *testing.T) {
 	go func() {
 		result, err := transactionsService.SendPaymentSync(ctx, transaction.PaymentRequest, nil, nil, svc.LNClient, nil, nil, nil)
 		assert.NoError(t, err)
+		require.NotNil(t, result)
 		assert.Equal(t, constants.TRANSACTION_STATE_SETTLED, result.State)
 		assert.Equal(t, true, result.SelfPayment)
 		assert.Equal(t, false, result.Hold)
 	}()
 
 	// wait for payment to be accepted
-	time.Sleep(10 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
 	settledTransaction, err := transactionsService.SettleHoldInvoice(ctx, preimage, svc.LNClient)
 	assert.NoError(t, err)
+	require.NotNil(t, settledTransaction)
 	assert.Equal(t, constants.TRANSACTION_STATE_SETTLED, settledTransaction.State)
 	assert.Equal(t, true, settledTransaction.SelfPayment)
 	assert.Equal(t, true, settledTransaction.Hold)
@@ -70,13 +72,14 @@ func TestSelfHoldPaymentCanceled(t *testing.T) {
 		outgoingTransactionType := constants.TRANSACTION_TYPE_OUTGOING
 		failedOutgoingTransaction, err := transactionsService.LookupTransaction(ctx, paymentHash, &outgoingTransactionType, svc.LNClient, nil)
 		assert.Nil(t, err)
+		require.NotNil(t, failedOutgoingTransaction)
 		assert.Equal(t, constants.TRANSACTION_STATE_FAILED, failedOutgoingTransaction.State)
 		assert.Equal(t, true, failedOutgoingTransaction.SelfPayment)
 		assert.Equal(t, false, failedOutgoingTransaction.Hold)
 	}()
 
 	// wait for payment to be accepted
-	time.Sleep(10 * time.Millisecond)
+	time.Sleep(100 * time.Millisecond)
 	err = transactionsService.CancelHoldInvoice(ctx, paymentHash, svc.LNClient)
 	assert.NoError(t, err)
 


constants/constants.go
```diff
@@ -37,6 +38,7 @@ const (
 	LOOKUP_INVOICE_SCOPE    = "lookup_invoice"
 	LIST_TRANSACTIONS_SCOPE = "list_transactions"
 	SIGN_MESSAGE_SCOPE      = "sign_message"
+	HOLD_INVOICES_SCOPE     = "hold_invoices" // covers all hold invoice operations (make/settle/cancel)
```
rolznz, line 41: I think we could use the existing `MAKE_INVOICE_SCOPE` for this. What do you think?

lnclient/lnd/lnd.go
```diff
@@ -278,6 +282,79 @@ func (svc *LNDService) subscribeChannelEvents(ctx context.Context) {
 	}
 }
 
+func (svc *LNDService) subscribeSingleInvoice(paymentHashBytes []byte) {
```
rolznz, line 285: There's already a goroutine to listen to all invoices, is this one necessary? CC @im-adithya

rolznz: This makes it more complicated right, what happens if the hub restarts? do we need to setup subscriptions for unsettled HODL invoices?

frnandu: You're right. Will try to fix that. Thanks

nip47/controllers/cancel_hold_invoice_controller.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type cancelHoldInvoiceParams struct {
+	PaymentHash string `json:"payment_hash"`
+}
+type cancelHoldInvoiceResponse struct{}
+
+func (controller *nip47Controller) HandleCancelHoldInvoiceEvent(ctx context.Context, nip47Request *models.Request, requestEventId uint, appId uint, publishResponse func(*models.Response, nostr.Tags)) {
+	if nip47Request == nil {
```
rolznz, line 19: this can't be null, can you remove it from the new controller methods and make sure they follow the same conventions as the existing methods?

constants/constants.go
frnandu: Not sure. Hard to say if there will be a need to have this separated or not.
We could always merge it later without big pain/consequence to clients right?

rolznz: Our current goal is to not introduce any scopes unless absolutely necessary. For example the PAY_INVOICE covers all payment methods.

I don't think from a permissions standpoint there is any need for another scope. If you let an app generate normal invoices it should be able to generate hold ones too I think

nip47/controllers/make_hold_invoice_controller.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		logger.Logger.WithFields(logrus.Fields{
+			"request_event_id": requestEventId,
+			"appId":            appId,
+			"amount":           makeHoldInvoiceParams.Amount,
+			"description":      makeHoldInvoiceParams.Description,
+			"descriptionHash":  makeHoldInvoiceParams.DescriptionHash,
+			"expiry":           makeHoldInvoiceParams.Expiry,
+			"paymentHash":      makeHoldInvoiceParams.PaymentHash,
+		}).Infof("Failed to make invoice: %v", err)
```
rolznz, line 75: Make sure to use JSON logging (check all places) + correct log level

nip47/controllers/make_hold_invoice_controller_test.go
```diff
@@ -0,0 +1,84 @@
...
+	assert.NoError(t, err)
+
+	var publishedResponse *models.Response
+
+	publishResponse := func(response *models.Response, tags nostr.Tags) {
+		publishedResponse = response
+	}
+
+	NewTestNip47Controller(svc).
+		HandleMakeInvoiceEvent(ctx, nip47Request, dbRequestEvent.ID, *dbRequestEvent.AppId, publishResponse)
```
rolznz, line 67: is this the right method?

nip47/controllers/settle_hold_invoice_controller_test.go
```diff
@@ -0,0 +1,128 @@
...
+	// svc.LNClient is already a *tests.MockLn, so no specific mock setup is needed here for SettleHoldInvoice
+	// as the default mock implementation returns nil, which is the success case.
+
+	var publishedResponse *models.Response
+	publishResponse := func(response *models.Response, tags nostr.Tags) {
+		publishedResponse = response
+	}
+
+	controller := NewTestNip47Controller(svc)
+	// Ensure the transactionsService within the controller uses the EventPublisher from the TestService
```
rolznz, line 107: this looks wrong :see_no_evil:

transactions/transactions_service.go
```diff
@@ -188,6 +191,59 @@ func (svc *transactionsService) MakeInvoice(ctx context.Context, amount uint64,
...
+		preimage = &lnClientTransaction.Preimage
+	}
+
+	var expiresAt *time.Time
+	if lnClientTransaction.ExpiresAt != nil {
+		expiresAtValue := time.Unix(*lnClientTransaction.ExpiresAt, 0)
+		expiresAt = &expiresAtValue
+	}
+
+	dbTransaction := db.Transaction{
```
rolznz, line 225: I wonder if it's worth identifying the invoice as a HOLD invoice. We could add a boolean to the invoice table?

constants/constants.go
```diff
@@ -9,6 +9,7 @@ const (
 	TRANSACTION_STATE_PENDING = "PENDING"
 	TRANSACTION_STATE_SETTLED = "SETTLED"
 	TRANSACTION_STATE_FAILED  = "FAILED"
+	TRANSACTION_STATE_ACCEPTED = "ACCEPTED"
```
rolznz, line 12: we need to make sure the frontend displays this new state correctly

transactions/transactions_service.go
```diff
@@ -730,6 +786,68 @@ func (svc *transactionsService) ConsumeEvent(ctx context.Context, event *events.
...
+			}
+
+			logger.Logger.WithFields(logrus.Fields{
+				"paymentHash": lnClientTransaction.PaymentHash,
+				"dbTxID":      dbTransaction.ID,
+			}).Info("Updated hold invoice state to accepted in DB")
+
+			var updatedDbTransaction db.Transaction
+			tx.First(&updatedDbTransaction, dbTransaction.ID)
+			svc.eventPublisher.Publish(&events.Event{
```
rolznz, line 839: can publishing be moved out of the DB transaction, otherwise this will lock the entire DB

nip47/controllers/settle_hold_invoice_controller.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/getAlby/hub/constants"
+	"github.com/getAlby/hub/logger"
+	"github.com/getAlby/hub/nip47/models"
+	"github.com/nbd-wtf/go-nostr"
+	"github.com/sirupsen/logrus"
+)
+
+type settleHoldInvoiceParams struct {
+	Preimage string `json:"preimage"`
```
rolznz, line 14: How does this work from the NWC end?

if you call make_hold_invoice you get a transaction back without a preimage (we do not reveal the preimage if the invoice is not settled)

Should this be called with payment_hash instead (and internally we lookup the invoice to get the preimage?)

transactions/transactions_service.go
```diff
@@ -975,6 +1093,140 @@ func (svc *transactionsService) getAppIdFromCustomRecords(customRecords []lnclie
...
+func (svc *transactionsService) SettleHoldInvoice(ctx context.Context, preimage string, lnClient lnclient.LNClient) (*Transaction, error) {
+	if len(preimage) != 64 {
+		return nil, errors.New("invalid preimage format")
+	}
+	preimageBytes, err := hex.DecodeString(preimage)
+	if err != nil {
+		return nil, fmt.Errorf("invalid preimage hex: %w", err)
+	}
+
+	err = lnClient.SettleHoldInvoice(ctx, preimage)
```
rolznz, line 1105: should we check the invoice state before calling the lnClient?

transactions/transactions_service.go
```diff
@@ -975,6 +1093,140 @@ func (svc *transactionsService) getAppIdFromCustomRecords(customRecords []lnclie
...
+			}
+			logger.Logger.WithFields(logrus.Fields{
+				"paymentHash": paymentHash,
+				"preimage":    preimage,
+			}).Warn("No pending hold invoice found in DB to mark as settled")
+			return NewNotFoundError()
+		}
+
+		var err error
+		settledTransaction, err = svc.markTransactionSettled(tx, &dbTransaction, preimage, 0, false) // Assuming not self-payment for now
```
rolznz, line 1152: hmm, self-payments are a problem. I'll add this to the PR TODO list

transactions/transactions_service.go
```diff
@@ -975,6 +1093,140 @@ func (svc *transactionsService) getAppIdFromCustomRecords(customRecords []lnclie
...
+		}).WithError(err).Error("Failed to cancel hold invoice via LN client")
+		// Don't mark DB as failed here, cancellation might have already happened or might succeed later.
+		return err
+	}
+
+	err = svc.db.Transaction(func(tx *gorm.DB) error {
+		var dbTransaction db.Transaction
+		result := tx.Limit(1).Find(&dbTransaction, &db.Transaction{
+			Type:        constants.TRANSACTION_TYPE_INCOMING,
+			State:       constants.TRANSACTION_STATE_PENDING,
```
rolznz, line 1185: is this right? should it be accepted?

transactions/transactions_service.go
```diff
@@ -975,6 +1093,140 @@ func (svc *transactionsService) getAppIdFromCustomRecords(customRecords []lnclie
...
+			"paymentHash": paymentHash,
+		}).WithError(err).Error("Failed DB transaction while canceling hold invoice")
+		return nil
+	}
+
+	logger.Logger.WithFields(logrus.Fields{
+		"paymentHash": paymentHash,
+	}).Info("Marked hold invoice as failed in DB due to cancellation")
+
+	// TODO: Should we publish a specific nwc_hold_invoice_canceled event?
```
rolznz, line 1225: I'm not sure if it's needed at least not for an MVP

nip47/controllers/settle_hold_invoice_controller.go
frnandu: The way it works in LND is the client generates the preimage (random 32 bytes) and keeps it.
Then creates the hold invoice with the payment hash generated from that preimage.
And when it wants to settle the invoice it must provide the preimage.
This causes the Hub to not have the preimage until the client decides to settle the invoice.

frnandu: It does have the preimage because generates it itself, and from that it generates the paymentHash to pass into `AddHoldInvoice` (LND)/`make_hold_invoice` (NWC).
The preimage is kept at the client so that only the client can decide to settle it.

frnandu: https://api.lightning.community/api/lnd/invoices/add-hold-invoice

constants/constants.go
rolznz: There's a usecase where you can create a receive-only NWC connection secret and share it publicly (for example the lightning messageboard on the hub homepage). With the addition of HOLD invoices a user could generate an invoice with a preimage and then claim that they paid the invoice since they already have the preimage.

I'm not sure if this is an issue or not. Or if there are other ways that receiving hodl invoices can be abused?

nip47/controllers/settle_hold_invoice_controller.go
rolznz: Ah, I see. Thanks!

constants/constants.go
rolznz: Actually, it makes no sense to give users the ability to request HOLD invoices apart from the service itself. So I think it's good you made it under a different scope :+1:

transactions/transactions_service.go
frnandu: Why? which cases do you see this being needed?

nip47/controllers/make_hold_invoice_controller_test.go
frnandu: fixed, thanks

nip47/controllers/settle_hold_invoice_controller_test.go
frnandu: fixed, thanks

constants/constants.go
frnandu: What should be the UI text state presented to the user for such hold invoice state?
`Holding` or same as `Pending`
@stackingsaunter 
`Accepted` feels like it will create confusion and misunderstanding.

frnandu: Actually the wallets both in the hub and in Alby Go currently seem to be asking with `unpaidOutgoing=true` which doesn't return any unsettled incoming transactions, so the state accepted should never be shown in either UI's. Couldn't test the extension but I suspect it's the same.
And since `state` is not included in the NIP-47 spec, other clients even if they use the state value returned by the hub, they can do whatever they choose to show.

transactions/transactions_service.go
frnandu: For calling `subscribeSingleInvoice` after hub restarts for invoices that where created but still didn't get accepted by the payer, we need to find them in DB in order to resubscribe to them.
And since they all have `PENDING` state we can't actually differentiate which ones are hold. 
So adding this column could help with that.

Another option is to add a new state `OPEN` which matches the state in LND -> https://api.lightning.community/api/lnd/invoices/subscribe-single-invoice#lnrpcinvoiceinvoicestate and find hold invoices by such a state.
Wdyt @rolznz @im-adithya ?

frnandu: Hmm, a quicker way is to filter by state `PENDING` and empty preimage.

rolznz: I don't think it's good to add another state - it's another the frontend + all NWC clients have to account for and display correctly in the transaction list (Both Alby Hub + Alby Go, extension etc.), and as soon as it leaves that state we don't know if it was a HOLD invoice any more. Same with inferring it from  state `PENDING` AND `preimage` NULL which seems a bit hacky to me.

I think it makes sense for the `ACCEPTED` state as that's definitely something new, but not for an `OPEN` state which while it's unpaid cannot be differentiated from `PENDING`.

I think it's good to know if an invoice was a hold invoice or not, even just for accounting purposes. What do you think?

constants/constants.go
rolznz: NOTE: @im-adithya plans to do a PR to add state to the spec. But alby Hub could return "pending" for both states until we see a clear need to add "accepted" as the state in the NWC response?

rolznz: @frnandu alternately instead of adding a new transaction state we could add a different column `hold_invoice_state` to the DB. This way it won't have any impacts on the UI by default?

transactions/transactions_service.go
frnandu: AFAIK all Alby wallet UIs don't display unpaid incoming invoices, so I think we don't need to handle such state in the wallet's UIs.
I don't see a reason accounting would need to know if an already settled invoice was hold or not?
Do you?

constants/constants.go
frnandu: I think when we'll add `state` to NWC spec, we can better define what state hold invoices will have before/after accepted, either use `pending` with type `incoming` and let wallets display `Receiving` or whatever if they choose to even display such transactions.
I don't see the need to add new column to DB since wallets don't display such pending incoming invoices, and it would just add more code and complicate things.

rolznz:
Linking https://github.com/getAlby/hub/issues/1212 as I think this is an example where we need to make sure the notification is sent to the relay.

transactions/transactions_service.go
```diff
@@ -985,6 +1095,120 @@ func (svc *transactionsService) getAppIdFromCustomRecords(customRecords []lnclie
...
+	if err != nil {
+		logger.Logger.WithFields(logrus.Fields{
+			"paymentHash": paymentHash,
+			"preimage":    preimage,
+		}).WithError(err).Error("Failed DB transaction while settling hold invoice")
+		return nil, err
+	}
+
+	if settledTransaction == nil {
+		return nil, fmt.Errorf("settled transaction is nil despite no error")
```
rolznz, line 1147: we should avoid code like this. If there is no error the method must return what we expect.

transactions/transactions_service.go
```diff
@@ -740,6 +796,60 @@ func (svc *transactionsService) ConsumeEvent(ctx context.Context, event *events.
...
+
+			err := tx.Model(&dbTransaction).Update("state", constants.TRANSACTION_STATE_ACCEPTED).Error
+			if err != nil {
+				logger.Logger.WithFields(logrus.Fields{
+					"paymentHash": lnClientTransaction.PaymentHash,
+					"dbTxID":      dbTransaction.ID,
+				}).WithError(err).Error("Failed to update hold invoice state to accepted in DB")
+				return err
+			}
+			dbTransaction.State = constants.TRANSACTION_STATE_ACCEPTED
```
rolznz, line 833: We must not set the state manually. It's done by the update above.

transactions/transactions_service.go
```diff
@@ -740,6 +796,60 @@ func (svc *transactionsService) ConsumeEvent(ctx context.Context, event *events.
...
+
+			logger.Logger.WithFields(logrus.Fields{
```
rolznz, line 835: for json logging we currently use `lower_case_separated_by_underscores` for the field names

constants/constants.go
rolznz: Edit: we decided it cannot be easily abused. So we can remove the new scope

lnclient/lnd/lnd.go
```diff
@@ -73,18 +77,20 @@ func NewLNDService(ctx context.Context, eventPublisher events.EventPublisher, ln
...
 
-	lndCtx, cancel := context.WithCancel(ctx)
+	lndCtx, svcCancel := context.WithCancel(ctx)
 
 	lndService := &LNDService{
 		client:         lndClient,
 		nodeInfo:       nodeInfo,
-		cancel:         cancel,
+		cancel:         svcCancel,
+		globalCtx:      ctx,
```
rolznz, line 86: why is a `globalCtx` needed? can't we use the `lndCtx`?

lnclient/lnd/lnd.go
```diff
@@ -279,6 +285,121 @@ func (svc *LNDService) subscribeChannelEvents(ctx context.Context) {
 	}
 }
 
+func (svc *LNDService) subscribeOpenHoldInvoices(ctx context.Context) {
+	// wait a bit for the node to be fully ready before querying for existing invoices
```
rolznz, line 289: can this commented out code be removed?

lnclient/lnd/lnd.go
```diff
@@ -279,6 +285,121 @@ func (svc *LNDService) subscribeChannelEvents(ctx context.Context) {
...
+				if htlc.ExpiryHeight < int32(minExpiry) || minExpiry == 0 {
+					minExpiry = uint32(htlc.ExpiryHeight)
+				}
+			}
+			transaction.SettleDeadline = &minExpiry
+			svc.eventPublisher.Publish(&events.Event{
+				Event:      "nwc_lnclient_hold_invoice_accepted",
+				Properties: transaction,
+			})
+		case lnrpc.Invoice_CANCELED:
```
rolznz, line 388: I don't think this case is needed and that event is not consumed. We already publish an event from the transactions service after successfully canceling the hold invoice

lnclient/lnd/lnd.go
```diff
@@ -512,6 +633,134 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
...
+	}
+
+	// Start subscribing to updates for this specific hold invoice in a separate goroutine
+	go svc.subscribeSingleInvoice(paymentHashBytes)
+	logger.Logger.WithField("paymentHash", paymentHash).Info("Launched single invoice subscription goroutine")
+
+	inv, err := svc.client.LookupInvoice(ctx, &lnrpc.PaymentHash{RHash: paymentHashBytes})
+	if err != nil {
+		logger.Logger.WithField("paymentHash", paymentHash).WithError(err).Error("Failed to lookup hold invoice after creation")
+		return &lnclient.Transaction{
```
rolznz, line 702: Can we remove this and just return an error

nip47/controllers/cancel_hold_invoice_controller.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		logger.Logger.WithFields(logrus.Fields{
+			"request_event_id": requestEventId,
+			"appId":            appId,
+			"paymentHash":      cancelHoldInvoiceParams.PaymentHash,
+		}).WithError(err).Error("Failed to cancel hold invoice")
+
+		publishResponse(&models.Response{
+			ResultType: nip47Request.Method,
+			Error: &models.Error{
```
rolznz, line 45: We can use `mapNip47Error` to use the correct error code (see other usages)

nip47/controllers/cancel_hold_invoice_controller.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type cancelHoldInvoiceParams struct {
+	PaymentHash string `json:"payment_hash"`
+}
+type cancelHoldInvoiceResponse struct{}
+
+func (controller *nip47Controller) HandleCancelHoldInvoiceEvent(ctx context.Context, nip47Request *models.Request, requestEventId uint, appId uint, publishResponse func(*models.Response, nostr.Tags)) {
+	resp := &models.Response{}
+	resp.ResultType = models.CANCEL_HOLD_INVOICE_METHOD
```
rolznz, line 20: This is not consistent with our other methods

nip47/controllers/settle_hold_invoice_controller_test.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	dbRequestEvent := &db.RequestEvent{
+		AppId: &app.ID,
+	}
+	err = svc.DB.Create(&dbRequestEvent).Error
+	require.NoError(t, err)
+
+	controller := NewTestNip47Controller(svc)
+	if controller.transactionsService == nil {
```
rolznz, line 96: this should not be needed (remove all occurrences)

nip47/controllers/cancel_hold_invoice_controller_test.go
```diff
@@ -0,0 +1,149 @@
...
+	nonExistentPaymentHash := "nonexistentpaymenthashnonexistentpaymenthashnonexistentpaymenthash"
+	setup := setupCancelHoldInvoiceTest(t, nonExistentPaymentHash, nil) // nil for initialTransactionState means no transaction created
+	defer setup.TearDown()
+
+	NewTestNip47Controller(setup.svc).
+		HandleCancelHoldInvoiceEvent(setup.ctx, setup.nip47Request, setup.dbRequestEvent.ID, *setup.dbRequestEvent.AppId, setup.PublishResponse)
+
+	assert.True(t, setup.publishCalled)
+	require.NotNil(t, setup.response.Error)
+	assert.Equal(t, constants.ERROR_INTERNAL, setup.response.Error.Code)
```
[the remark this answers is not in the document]
rolznz: resolved

rolznz (approved):
tACK

Issue #1290, opened 2025-05-01
If initial publish of info event for new app fails, the app is completely unusable until Alby Hub is restarted


rolznz:
I tried to create a new app but the publish of the info event to the relay failed. The app is then completely unusable and there is nothing on the frontend to indicate that something went wrong. You just have a connection string that doesn't work (e.g. if you put it in the extension it will just fail at the test connection step).

frnandu:
Because of the 1-click flow recent new method `create_connection` , the nip47Service now needs the `appsService` (so it can call `CreateApp` in the `create_connection_controller` ) and so adding a dependency for `nip47Service` in `appsService` would cause a circular dependency.

rolznz:
Closed by https://github.com/getAlby/hub/pull/1300

If we realize it's insufficient we can create a new issue that does it in sync (as Fernando mentioned it is a more complicated task)

PR #1300, opened 2025-05-07
fix: re-enqueue failed info publish request

rolznz:
partial workaround for https://github.com/getAlby/hub/issues/1290
part of https://github.com/getAlby/hub/issues/1212

### DIFF
--- a/nip47/nip47_service.go
+++ b/nip47/nip47_service.go
@@ -2,6 +2,7 @@ package nip47
 
 import (
 	"context"
+	"time"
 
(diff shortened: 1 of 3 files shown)


frnandu (approved):
The only way this might fail is if between consuming the element from `nip47InfoPublishQueue` and calling `svc.PublishNip47Info` the relay disconnects for some reason and it doesn't manage to re-enqueue.
But the chances of that happening are pretty low, so utACK.

PR #1311, opened 2025-05-13
chore: add defaults for avatar and name in sidebar

### LINKED ISSUE #1238: Sidebar doesn't handle alby account not connected state
There is just an icon on the right with no name, ~we can atleast say Satoshi Nakamoto with SN icon like we used to before~

We removed that because people were confused as to who is Satoshi Nakamoto (everyone is ;))

Image

im-adithya:
Fixes #1238 

## Screenshot

sats


frontend/src/components/AppSidebar.tsx
```diff
@@ -181,19 +181,19 @@ export function AppSidebar() {
                   size="lg"
                   className="data-[state=open]:bg-sidebar-accent data-[state=open]:text-sidebar-accent-foreground"
                 >
-                  {info?.albyAccountConnected && (
```
im-adithya, line 184: Not sure if this is good, or should we do `info?.albyAccountConnected` with ternary to avoid mix up

PR #1316, opened 2025-05-16
chore: filter apps from aotd widget

im-adithya:
It is possible that sometimes apps which already have a standalone widget are featured again in App Of The Day widget, doesn't look that good.

Screenshot 2025-05-16 at 1 42 59 PM


reneaaron (approved):
utACK

PR #1317, opened 2025-05-16
feat: add frequency selector for ZapPlanner payments (#1312)

### LINKED ISSUE #1312: Include Frequency option in Integrted ZapPlanner
At the moment, recurring payments can only be done monthly.

Lots of users could benefit from choosing a more adequate frequency. 
Particularly users who are not currently using lightning so often could benefit from scheduling a weekly or bi-weekly donation to prevent automatic channel closures due to inactivity. 

I propose to add a frequency box similar to what's on https://zapplanner.albylabs.com/create
This would help users create recurring payments more often and prevent channel closures due to inactivity.

Image

NodeDiver:
## What’s new
- Added frequency selector to the ZapPlanner recurring payment form
- Users can now choose the number and unit (`days`, `weeks`, `months`) for scheduling payments

## Why
- Improves parity with [zapplanner.albylabs.com](https://zapplanner.albylabs.com)
- Solves feature request in #1312

### 🔍 How to test
- Go to Internal Apps → ZapPlanner
- Click "New Recurring Payment"
- Submit to create a new scheduled app + subscription with custom frequency

---

Fixes #1312


frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -87,6 +87,8 @@ export function ZapPlanner() {
   const [amount, setAmount] = React.useState("");
   const [comment, setComment] = React.useState("");
   const [senderName, setSenderName] = React.useState("");
+  const [frequencyValue, setFrequencyValue] = React.useState("1");
```
reneaaron, line 90: This means by default the payment is done every day. Is that a good default you think? I'd argue monthly payments are probably more common, also that was the default up until now.

reneaaron (changes requested):
image

The UI of this select looks wrong: 
- Try to move the "Repeat payment every" below the inputs and make it secondary text color (text-muted-foreground) 
- Use the shadcn select instead of the HTML select to apply proper styles that match other forms
- Make the input wider so more than 1 digit fits (if I want to create a payment every 12 weeks)

NodeDiver:
_In short: **Added frequency and USD/BTC selector**_

_List of changes:_
- Added frequency selector: number + unit (days/weeks/months)
- Added USD/BTC toggle with inline conversion preview (using `fiat` tools)
- Updated default values: amount resets to “5” (USD) and frequency to 1 month; BTC option now labeled “BTC (sats)”
- Expanded input boxes for consistency and full visibility of lightning address 
- Calculated dynamic monthly budget cap:  `maxAmount = ceil((sats * periodsPerMonth * 1.01 + 10) * 1.3)`
- Updated `handleSubmit` to always use `satsToSend`and removed legacy `parsedAmount`  
- Converted “months” → days (1 month = 31 days) since month as a unit of time is not compatible (week and days are)
- Fixed dialog‐reopen bug by adding `open` to conversion `useEffect` dependencies

_I'm happy with the results so far 😊, but I know there is room for improvement and more changes. I’m all ears for any suggestions or additional ideas 🙏._

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -87,18 +94,68 @@ export function ZapPlanner() {
...
+  const [frequencyUnit, setFrequencyUnit] = React.useState("months");
+  const [amountCurrency, setAmountCurrency] = React.useState<"BTC" | "USD">(
```
rolznz, line 99: Later we could create a `useCurrencies` hook. The code for fetching currencies is currently in `frontend/src/screens/settings/Settings.tsx`. CC @pavanjoshi914

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -141,14 +224,17 @@ export function ZapPlanner() {
             "Content-Type": "application/json",
           },
           body: JSON.stringify({
-            recipientLightningAddress: recipientLightningAddress,
-            amount: parsedAmount,
+            recipientLightningAddress,
+            amount: satsToSend, // ← always sats
+            // TODO: include other currencies
```
rolznz, line 229: currency needs to be provided here right?

frontend/src/screens/internal-apps/ZapPlanner.tsx
rolznz: This should be a string, not "BTC" | "USD" otherwise we get a type error

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -109,13 +166,32 @@ export function ZapPlanner() {
...
-      if (isNaN(parsedAmount) || parsedAmount < 1) {
-        throw new Error("Invalid amount");
+      // Determine the satoshi amount to send:
+      const satsToSend =
+        satoshiAmount ??
+        (() => {
+          throw new Error("Invalid amount");
+        })();
+      // Determine how many payments in one month
+      let periodsPerMonth: number;
```
rolznz, line 176: we should do Math.ceil() on this otherwise if you select every 2 months it will set the value to 0.5 and this will break the budgeting

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -256,20 +342,85 @@ export function ZapPlanner() {
...
+                      </Label>
+                      <div className="col-span-3 flex flex-col gap-1 w-full max-w-[450px]">
+                        <div className="flex items-center gap-2 w-full">
+                          <Input
+                            id="frequency"
+                            type="number"
+                            min="1"
+                            value={frequencyValue}
+                            onChange={(e) => {
+                              const value = Math.max(
```
rolznz, line 395: This creates poor UX e.g. you can't even clear the input. We shouldn't modify the value here at all. Any manipulation of the value can be done in the submit function

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -141,14 +224,17 @@ export function ZapPlanner() {
             "Content-Type": "application/json",
           },
           body: JSON.stringify({
-            recipientLightningAddress: recipientLightningAddress,
-            amount: parsedAmount,
+            recipientLightningAddress,
+            amount: satsToSend, // ← always sats
```
rolznz, line 228: this is not correct, for fiat you provide the fiat value and fiat currency. ZapPlanner will calculate the sat value for every single payment, therefore the payment amount will be "stable" in fiat terms (which is the whole point of adding this feature)

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -87,18 +94,68 @@ export function ZapPlanner() {
   const [amount, setAmount] = React.useState("");
   const [comment, setComment] = React.useState("");
   const [senderName, setSenderName] = React.useState("");
+  const [frequencyValue, setFrequencyValue] = React.useState("1");
+  const [frequencyUnit, setFrequencyUnit] = React.useState("months");
```
rolznz, line 98: Can you double check the new fields get properly reset when the modal is closed?

frontend/src/screens/internal-apps/ZapPlanner.tsx
NodeDiver: Things done in my latest commit:
- Added support for multicurrency
- 2xbtc entries were shown on the list, I deduplicated BTC entry, only 1 btc on the top.
- Every currency gets converted to sats in the displayed sats preview; 
- Every currency gets converted to sats; in the nwc subscription

I have taken inspiration from the settings and the original zapplanner repo.

frontend/src/screens/internal-apps/ZapPlanner.tsx
NodeDiver: Added this line just below the   periodsPerMonth definition,
`    periodsPerMonth = Math.ceil(periodsPerMonth);`

frontend/src/screens/internal-apps/ZapPlanner.tsx
NodeDiver: Standardized both inputs to prevent wrong characters/negative numbers:

```
       <Input
                            id="frequency"
                            type="number"
                            min="1"
                            step="1"
                            inputMode="numeric"
                            value={frequencyValue}
                            onChange={(e) => setFrequencyValue(e.target.value)}
                            className="col-span-3 w-70"
                          />
```

frontend/src/screens/internal-apps/ZapPlanner.tsx
NodeDiver: Added this to reset currency selected and frecuency:

```
      setFrequencyValue("1");
      setFrequencyUnit("months");
      setAmountCurrency("USD");
```

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -87,35 +94,134 @@ export function ZapPlanner() {
...
+
+  React.useEffect(() => {
+    // fetch the fiat list and prepend sats/BTC
+    async function fetchCurrencies() {
+      try {
+        const res = await fetch("https://getalby.com/api/rates");
+        const data: Record<string, { name: string }> = await res.json();
+        const fiatCodes = Object.keys(data)
+          .map((c) => c.toUpperCase())
+          // drop "BTC" so it only lives in our prepended slot
```
rolznz, line 117: I don't think our API is supposed to return BTC. So we might be able to remove this line

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -87,35 +94,134 @@ export function ZapPlanner() {
...
+          .filter((code) => code !== "BTC")
+          .sort();
+        setCurrencies(["BTC", ...fiatCodes]);
```
rolznz, line 120: It might be weird, but ZapPlanner expects "SATS" as the currency for Bitcoin (see "for developers" on the ZapPlanner page). It seems to return a 400 right now if you set BTC as the currency.

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -87,35 +94,134 @@ export function ZapPlanner() {
   const [amount, setAmount] = React.useState("");
   const [comment, setComment] = React.useState("");
   const [senderName, setSenderName] = React.useState("");
+  const [frequencyValue, setFrequencyValue] = React.useState("1");
+  const [frequencyUnit, setFrequencyUnit] = React.useState("months");
+  const [amountCurrency, setAmountCurrency] = React.useState<"BTC" | "USD">(
```
rolznz, line 99: could this be renamed to just "currency" instead of "amountCurrency" ?

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -131,6 +237,34 @@ export function ZapPlanner() {
...
+          ...(senderName ? { name: senderName } : {}),
+        }),
+        nostrWalletConnectUrl: createAppResponse.pairingUri,
+        sleepDuration,
+      };
+
+      if (amountCurrency === "BTC") {
+        // user entered sats directly
+        subscriptionBody.amount = satsToSend;
+        subscriptionBody.currency = "BTC";
```
rolznz, line 262: maybe just this line can be removed, so it uses the default currency ("SATS")

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -131,6 +237,34 @@ export function ZapPlanner() {
...
+      } else {
+        // user entered fiat → ZapPlanner will convert to sats later
+        subscriptionBody.amount = parseFloat(amount);
```
rolznz, line 265: I think unfortunately only whole numbers are supported right now. I tried 0.5 USD and it returned a 400 error.

rolznz (approved):
tACK

PR #1321, opened 2025-05-20
fix: make input full width for zaplanner and messageboard dialogs

pavanjoshi914:
its because of this instances directly give col-span-3 to the <input> tag but after introduction to end adornment for input we wrap input in a div
we should always use
```
<div class="col-span-3">
<input... />
</div>
```
we do it at most of the places expect two screens which i fixed below. also checked all the input instances should be okay now

### DIFF
--- a/frontend/src/components/home/widgets/LightningMessageboardWidget.tsx
+++ b/frontend/src/components/home/widgets/LightningMessageboardWidget.tsx
@@ -246,27 +246,29 @@ export function LightningMessageboardWidget() {
                 <Label htmlFor="comment" className="text-right">
                   Your Name
                 </Label>
-                <Input
-                  id="sender-name"
-                  value={senderName}
-                  onChange={(e) => setSenderName(e.target.value)}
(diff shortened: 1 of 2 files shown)


reneaaron:
What was changed in your PR that causes this behavior?

rolznz:
@pavanjoshi914 I don't think this is a very good solution. We need to add it every time we add an input, and how will we remember to always add this? is there a better way a fix can be applied to the Input component? doesn't this only apply to password inputs...?

rolznz:
@pavanjoshi914 ok, thanks!

pavanjoshi914:
the only way we can fix i guess is introduce wrapperClass which applies only to div and not input. but its better to use col-span as outdiv instead of introducing another prop. wdyt?

rolznz (approved):
tACK

rolznz:
@pavanjoshi914 I think it's ok.

@reneaaron if you have a better solution let us know

PR #1322, opened 2025-05-20
chore: add metadata to payments done via Send screen

### LINKED ISSUE #1023: Missing transaction details
Transaction failed. I can not see details.

Missing info
- destination (ln address or invoice/destination
- no error information/reason

Image

im-adithya:
Fixes #1023

## Description

Adds recipient whenever a LN address is used.

## Screenshot

Screenshot 2025-05-20 at 2 50 51 PM

### DIFF
--- a/frontend/src/screens/wallet/send/ConfirmPayment.tsx
+++ b/frontend/src/screens/wallet/send/ConfirmPayment.tsx
@@ -5,7 +5,7 @@ import { Label } from "src/components/ui/label";
 import { LoadingButton } from "src/components/ui/loading-button";
 import { useToast } from "src/components/ui/use-toast";
 
-import { Invoice } from "@getalby/lightning-tools";
+import { Invoice, LightningAddress } from "@getalby/lightning-tools";
 import FormattedFiatAmount from "src/components/FormattedFiatAmount";
 import Loading from "src/components/Loading";
 import { PayInvoiceResponse } from "src/types";
@@ -16,8 +16,9 @@ export default function ConfirmPayment() {
   const navigate = useNavigate();
   const { toast } = useToast();
 
-  const invoice = state?.args?.paymentRequest as Invoice;
   const amount = state?.args?.amount as number | undefined;
+  const invoice = state?.args?.paymentRequest as Invoice;
+  const recipient = state?.args?.recipient as LightningAddress;
   const [isLoading, setLoading] = React.useState(false);
 
   const onSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
@@ -30,6 +31,7 @@ export default function ConfirmPayment() {
           method: "POST",
           body: JSON.stringify({
             amount: amount ? amount * 1000 : undefined,
+            metadata: recipient ? { recipient: recipient.address } : undefined,
           }),
           headers: {
             "Content-Type": "application/json",
--- a/frontend/src/screens/wallet/send/LnurlPay.tsx
+++ b/frontend/src/screens/wallet/send/LnurlPay.tsx
@@ -32,7 +32,10 @@ export default function LnurlPay() {
       });
       navigate(`/wallet/send/confirm-payment`, {
         state: {
-          args: { paymentRequest: invoice },
+          args: {
+            paymentRequest: invoice,
+            recipient: lnAddress,
+          },
         },
       });
     } catch (e) {
--- a/api/models.go
+++ b/api/models.go
@@ -38,7 +38,7 @@ type API interface {
 	GetBalances(ctx context.Context) (*BalancesResponse, error)
 	ListTransactions(ctx context.Context, appId *uint, limit uint64, offset uint64) (*ListTransactionsResponse, error)
 	ListOnchainTransactions(ctx context.Context) ([]lnclient.OnchainTransaction, error)
-	SendPayment(ctx context.Context, invoice string, amountMsat *uint64) (*SendPaymentResponse, error)
+	SendPayment(ctx context.Context, invoice string, amountMsat *uint64, metadata map[string]interface{}) (*SendPaymentResponse, error)
 	CreateInvoice(ctx context.Context, amount uint64, description string) (*MakeInvoiceResponse, error)
 	LookupInvoice(ctx context.Context, paymentHash string) (*LookupInvoiceResponse, error)
 	RequestMempoolApi(endpoint string) (interface{}, error)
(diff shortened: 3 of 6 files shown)


rolznz:
@im-adithya nice. Do you think there's a good place to fit it in the transaction list? or only when you view a specific transaction  entry?

im-adithya:
We can do something like this, wdyt?

Screenshot 2025-05-21 at 12 49 00 PM

rolznz:
@im-adithya looks good to me!

im-adithya:
Ready!

bumi:
great

frontend/src/components/TransactionItem.tsx
```diff
@@ -74,14 +74,22 @@ function TransactionItem({ tx }: Props) {
...
+  const payerName = tx.metadata?.payer_data?.name;
+  const from = payerName
+    ? `from ${payerName}`
     : npub
       ? `zap from ${npub.substring(0, 12)}...`
       : undefined;
 
+  const recipientIdentifier = tx.metadata?.recipient_data?.identifier;
+  const to = recipientIdentifier
+    ? `${tx.state === "failed" ? "sending " : ""}to ${recipientIdentifier}`
```
rolznz, line 86: This looks wrong. What's it supposed to say? if it failed it doesn't mean "sending"

im-adithya: "Failed to send to" didn't feel good because of two tos so I changed it to be "Failed sending to" instead

rolznz: Ah, I see "Failed " is in front. Maybe "Failed payment" rather than "Failed sending"?

frontend/src/components/TransactionItem.tsx
```diff
@@ -74,14 +74,22 @@ function TransactionItem({ tx }: Props) {
...
+    : undefined;
+
   const eventId = tx.metadata?.nostr?.tags?.find((t) => t[0] === "e")?.[1];
 
+  const description = tx.description || tx.metadata?.comment;
```
rolznz, line 91: can we add a comment about how the metadata.comment can be set?

this can be set by the payer for LNURL payments

rolznz: Actually, it seems to not be set. I tried to pay to rolznz@getalby.com with a custom comment and it just set it as the invoice description, and didn't add the comment to the metadata. I think we need to add it to the metadata too

frontend/src/screens/wallet/send/ConfirmPayment.tsx
```diff
@@ -30,6 +31,9 @@ export default function ConfirmPayment() {
           method: "POST",
           body: JSON.stringify({
             amount: amount ? amount * 1000 : undefined,
+            metadata: identifier
+              ? { recipient_data: { identifier } }
```
rolznz, line 35: should we include the comment here too? e.g. paying to a WoS address we would not see the comment by default

frontend/src/components/TransactionItem.tsx
```diff
@@ -243,6 +262,15 @@ function TransactionItem({ tx }: Props) {
...
             )}
+            {tx.metadata?.comment && (
+              <div className="mt-6">
+                <p>Comment</p>
+                <p className="text-muted-foreground break-all">
+                  {tx.metadata?.comment}
+                </p>
+              </div>
+            )}
+            {/* We don't include zap request content as that is set as the description itself by the rails app */}
```
rolznz, line 273: This comment took me a minute to understand after reading it. I think here we should refer to "the rails app" as "the getalby.com backend". Also not sure we should even say it because hub runners could use a different lightning address server. But we could say "for Alby lightning addresses the description is automatically extracted from the zap request, and already displayed above"

frontend/src/components/TransactionItem.tsx
im-adithya: Yeah you are correct, we use this comment as description of the invoice :/ Wouldn't it then be redundant in the modal (metadata.comment and description)?

But I think it's good to follow the spec anyways. Should we then change rails app to set the comment to metadata instead of description?

frontend/src/components/TransactionItem.tsx
im-adithya: Sounds better, will change 🙌

frontend/src/screens/wallet/send/ConfirmPayment.tsx
im-adithya: Resolving, discussed above^

frontend/src/components/TransactionItem.tsx
im-adithya: I was also thinking about it, that sounds good, changed!

frontend/src/components/TransactionItem.tsx
rolznz: No, we can't just consider Alby lightning addresses. Other lightning addresses do not set the description. So I'll re-open the other conversation

frontend/src/screens/wallet/send/ConfirmPayment.tsx
rolznz: Can you review this again?

rolznz: We should set the comment here otherwise we rely on the recipient's lightning address server to set the memo on the invoice. At least one large lightning address provider does not do this.

rolznz (approved):
tACK

frontend/src/screens/wallet/send/LnurlPay.tsx
```diff
@@ -34,7 +35,10 @@ export default function LnurlPay() {
         state: {
           args: {
             paymentRequest: invoice,
-            identifier: lnAddress.lnurlpData?.identifier,
+            metadata: {
+              ...(comment && { comment }),
+              ...(identifier && { recipient_data: { identifier } }),
+            },
```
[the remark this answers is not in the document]
rolznz: Ah ok, thanks!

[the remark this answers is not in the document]
rolznz: so we need to change it back to `...(comment && { comment })`?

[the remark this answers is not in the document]
im-adithya: Oops sorry, I changed it for the screenshot and forgot to add it back, done now ✅

PR #1323, opened 2025-05-21
refactor: replace HasPrefix+TrimPrefix with CutPrefix

pullmerge:
Optimize code using a more modern writing style.


pullmerge:
> Hi @pullmerge thanks for the PR!
> 
> Unfortunately I think the change is a bit less readable and does not really improve our codebase in my opinion. You could look through the issues section or try the app and report any bugs you see, this could be a good way to find something to work on. Feel free to tag me before you start if you like so I can review it.

Okay. Thanks for your guidance.

PR #1329, opened 2025-05-22
feat: show success action message or url on payment confirmation

im-adithya:
## Screenshots

<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>


im-adithya:
I hardcoded to check it, but here's a 1 sat working example: https://ereignishorizont.xyz/tickettest/ (couldn't test since it's on mainnet)

rolznz:
@im-adithya I tried pasting `LNURL1DP68GURN8GHJ7V33D45K7TNNWPSKXEF0D3H82UNVWQH4WEMS894HW9ZP4ZS` into the send box and it fails. Or would you expect it to only work with a lightning address, and which lightning address has a success action? if so I think we should test with one before merging this

im-adithya:
Ah, we don't support LNURLs in Hub, will add this to lightning-tools and then use it here :+1:

frontend/src/screens/wallet/send/ConfirmPayment.tsx
```diff
@@ -43,7 +43,7 @@ export default function ConfirmPayment() {
       navigate(`/wallet/send/success`, {
         state: {
           preimage: payInvoiceResponse.preimage,
-          amount: amount || invoice.satoshi,
+          invoice: invoice,
```
rolznz:
```suggestion
          invoice,
```

frontend/src/screens/wallet/send/PaymentSuccess.tsx
```diff
@@ -17,6 +19,8 @@ export default function PaymentSuccess() {
   const navigate = useNavigate();
   const { toast } = useToast();
 
+  const invoice = state?.invoice as Invoice;
```
rolznz: you should not do this, you are casting an optional field to a non-optional type.

Instead we could have a check:
```
if (!state?.invoice) {
  return null;
}
```

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
rolznz: remove this file, you ran yarn install in the parent directory

frontend/src/screens/wallet/send/PaymentSuccess.tsx
im-adithya: Done!

PR #1332, opened 2025-05-23
chore: do not show error log when swaps are not configured

im-adithya:
We currently show an error log like this:
```
{"error":"auto swap not configured","level":"error","msg":"Couldn't enable auto swaps","time":"2025-05-23T21:21:21+07:00"}
```


rolznz (approved):
utACK

PR #1336, opened 2025-05-23
feat: set 0 reserve for self payments

rolznz:


bumi (approved):
for sure is better for self payments. code looks good to me. 
If we don't miss any potential side effects :) :shipit:

PR #1355, opened 2025-05-29
feat(subaccounts-json): enable JSON download & bump export limit to 50 (#1337)

### LINKED ISSUE #1337: Accounting feature: Add Subaccount/NWC Identifier or filters for transaction history
A DIY PRO subscriber who is a bitcoin business user with over 10 subaccounts on their Hub reported a limitation when trying to manage transactions for accounting purposes. One of the subaccounts is dedicated to their business, and they may create more in the future. Currently, there’s no way to:

- Filter transaction history by subaccount or NWC connection inside the Hub
- Download subaccount-specific data
- Identify subaccounts in the downloadable CSV or JSON from getalby.com

This affects not only usability but also business-related accounting, especially for users managing multiple financial streams. I've seen similar requests before from users with business use cases, so while the number of affected users might be small, the impact on them is significant.

**Possible solutions (open to feedback):**

1. Add filtering options by subaccount/NWC connection inside the Hub wallet view, along with a downloadable history feature (potentially as a Pro feature).
2. As a quicker first step, include a new field in the downloadable CSV/JSON on getalby.com showing which subaccount or NWC connection is linked to each transaction. This would allow users to self-filter with spreadsheet tools.

Improving this would give users that need to do accounting on their main/subwallets more control, encourage long-term use, and support word-of-mouth growth among other business owners 🧾📈

I believe option 2 could be easy to implement and would let us evaluate its impact before deciding if something like option 1 is worth pursuing.

Thanks for considering this!

NodeDiver:
This PR adds a JSON download link for subaccount-specific transactions and raises the default export limit from 20 to 50.

- **Frontend** (`ShowApp.tsx`): Inserts a “Download JSON” link that fetches `/api/transactions?appId=<id>&format=json` and saves it as a file.
- **Backend** (`http_service.go`): Increases the default limit to 50 in `listTransactionsHandler` and supports returning the JSON payload as a downloadable file.

Closes #1337

----
I originally planned to mirror getalby.com’s CSV export, but adding CSV support on the backend introduced a lot of complexity for a first iteration. A JSON download still meets the core need: users can retrieve their full subaccount history for precise accounting and convert it to any format they prefer. Many accountants work comfortably with JSON.

Note: The export limit has been raised to 50 transactions as a conservative first step. We may revisit this limit in the future, there may be a reason the original cap was 20 that deserves further investigation.

Thanks for the review!
image

### DIFF
--- a/frontend/src/screens/apps/ShowApp.tsx
+++ b/frontend/src/screens/apps/ShowApp.tsx
@@ -81,6 +81,31 @@ type AppInternalProps = {
 
 function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
   const { toast } = useToast();
+  // Download sub-account’s full transaction list as JSON
+  const handleExportJson = async (e: React.MouseEvent) => {
+    e.preventDefault();
+    try {
+      // fetch entire tx list for this app
+      const { totalCount, transactions } = await request(
+        `/api/transactions?appId=${app.id}`
+      );
+      // if you ever need pagination, you could refetch here with ?limit=totalCount
+      const payload = { totalCount, transactions };
+      const blob = new Blob([JSON.stringify(payload, null, 2)], {
+        type: "application/json",
+      });
+      const url = URL.createObjectURL(blob);
+      const a = document.createElement("a");
+      a.href = url;
+      a.download = `subaccount-${app.id}-transactions.json`;
+      document.body.appendChild(a);
+      a.click();
+      a.remove();
+      URL.revokeObjectURL(url);
+    } catch (error) {
+      toast({ title: "Error exporting JSON", variant: "destructive" });
+    }
+  };
   const navigate = useNavigate();
   const location = useLocation();
   const [isEditingName, setIsEditingName] = React.useState(false);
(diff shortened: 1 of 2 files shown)


http/http_service.go
```diff
@@ -564,37 +564,33 @@ func (httpSvc *HttpService) lookupTransactionHandler(c echo.Context) error {
...
-	limit := uint64(20)
+	// parse limit/offset/appId
+	limit := uint64(50)
 	offset := uint64(0)
 	var appId *uint
 
-	if limitParam := c.QueryParam("limit"); limitParam != "" {
-		if parsedLimit, err := strconv.ParseUint(limitParam, 10, 64); err == nil {
-			limit = parsedLimit
+	if lp := c.QueryParam("limit"); lp != "" {
```
rolznz, line 572: why did you rename the variables?

http/http_service.go
```diff
@@ -564,37 +564,33 @@ func (httpSvc *HttpService) lookupTransactionHandler(c echo.Context) error {
 func (httpSvc *HttpService) listTransactionsHandler(c echo.Context) error {
 	ctx := c.Request().Context()
 
-	limit := uint64(20)
+	// parse limit/offset/appId
+	limit := uint64(50)
```
rolznz, line 568: this will apply on all pages, I'm not sure we want to change this value.

frontend/src/screens/apps/ShowApp.tsx
```diff
@@ -416,6 +441,18 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
               </CardTitle>
             </CardHeader>
             <CardContent>
+              <div className="flex justify-end text-sm text-muted-foreground mb-4">
```
rolznz, line 444: Note: you should try to be consistent with our other buttons in the app, and use existing components where possible

PR #1356, opened 2025-05-29
feat: second ark wallet support (PoC)

rdmitr:


lnclient/second_ark/second_ark.go
```diff
@@ -0,0 +1,373 @@
+package second_ark
```
rolznz, line 1: Minor: should we call it "bark" everywhere rather than "Bark" in some places and "SecondArk" elsewhere?

lnclient/second_ark/second_ark.go
```diff
@@ -0,0 +1,373 @@
...
+				}
+			case <-cctx.Done():
+				return
+			}
+		}
+	}()
+
+	return &SecondArkService{
+		wallet: wallet,
+		cancel: cancel,
```
rolznz, line 108: we need to add the WaitGroup on the service right? it crashes on exit

lnclient/second_ark/second_ark.go
```diff
@@ -0,0 +1,373 @@
...
+		logger.Logger.WithFields(logrus.Fields{
+			"bolt11": invoice,
+		}).WithError(err).Error("Failed to decode bolt11 invoice")
+
+		return nil, err
+	}
+
+	// Bark won't allow setting amount on invoices that already have an amount set.
+	// Bark also does not support msats, offering rounding up or down to the nearest sat.
+	// FIXME: Rounding up here to avoid underpayment; however, this may lead to unexpected spending.
```
rolznz, line 137: We could return an error if it's not a whole sat amount, I think that's sufficient for now

lnclient/second_ark/second_ark.go
rdmitr: Yep, absolutely. I will also rename stuff in the bindings

lnclient/second_ark/second_ark.go
```diff
@@ -0,0 +1,373 @@
...
+			logger.Logger.WithError(err).Error("Failed to check Second Ark database file")
+			return nil, err
+		}
+
+		barkConfig := bindings.Config{
+			Network:              "signet",
+			Birthday:             nil,
+			AspAddress:           "https://ark.signet.2nd.dev",
+			EsploraAddress:       "https://esplora.signet.2nd.dev",
+			VtxoRefreshThreshold: 0,
```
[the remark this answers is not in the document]
rdmitr: > Can we use the default and bark handle which vtxos to refresh?

Yep, we're doing `refresh_all()` anyways. The `vtxo_refresh_threshold` setting should not be exposed here actually, as it is only used by the command-line client. I will remove it.

> we should call Command::Refresh rather than maintenance directly, shouldn't we?

Not really, the `Command` enum is used to define (and parse) command-line commands and args in the CLI app, and it is only used in the CLI app entry point. The entry point (`inner_main()`) is not accessible in the lib interface, and even if it was, it would be fairly awkward to invoke: it is not designed as a library function (i.e. it can shut down the process), and it sends its output to stdout.

lnclient/second_ark/second_ark.go
rdmitr: Great catch, thanks!

lnclient/second_ark/second_ark.go
rdmitr: Sounds good for custom amounts, but what if it's the invoice that has msats? Should we reject it?

rolznz: If the amount is already in the invoice then the amount should not be provided separately right?

rdmitr: Yeah, but I was concerned whether invoices with msat amounts should be rejected on our side — bark handles them by rounding up the amount, too 🤔

lnclient/second_ark/second_ark.go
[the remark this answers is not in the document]
rolznz: Ah right, thanks

lnclient/second_ark/second_ark.go
rolznz: If it's a standard invoice then bark should handle it, that's fine. If it's a 0-amount invoice and the user has to specify an amount, I think it should always be in sats.

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/2374

PR #1370, opened 2025-06-03
chore: convert tailwind config to esm

im-adithya:
## Description

This makes it work in both node v20 and v22

Previously it failed on v22 saying:
```
file:///.../hub/frontend/tailwind.config.js:1
const defaultTheme = require("tailwindcss/defaultTheme");
                     ^

ReferenceError: require is not define
```


rolznz (approved):
tACK

PR #1371, opened 2025-06-03
Swaps improvements (shifting to standalone page + one-time swaps)

### LINKED ISSUE #1368: Use Boltz client for all swaps
Currently we only use it for auto swaps, but it seems most users just do 1-off swaps.

### LINKED ISSUE #1369: Make auto-swaps more visible in the node page
- add auto-swaps as an item in the advanced navigation on the node page
- add auto-swap link in the one off swap out modal

### LINKED ISSUE #1380: Better integration for swaps + recurring swaps
[Figma link](https://www.figma.com/design/2IyilDewDrdNP3G1NixPYd/%F0%9F%8C%93-Alby-Hub?node-id=8643-52341&t=DYeUMNZrAFkyg1as-4)

Image

Image

Image

Image

im-adithya:
Fixes #1368
Fixes #1369
Fixes #1380

## TODOs

# Swaps testing template

### Steps:
- Pay the invoice
- Boltz then locks up funds
- Wait for confirmation
- Claim the payment
- Using this claim's preimage, boltz will claim the htlc

### Edge cases to test:

- [x] Invoice is expired/invalid
- [x] Payment error (route not found, low spending capacity)
- [x] Hub is shut down after initiation of swap
- [x] Hub is shut down after boltz lockup is found in mempool
- [x] Hub is shut down after boltz lockup confirms
- [x] Hub is shut down while lockup is found
- [x] Hub is shut down while lockup is confirmed
- [x] Hub is shut down for a long time while lockup is confirmed and timeout is passed -> Payment fails but it doesn't receive any update - reached out to boltz to add this

### Refunds

Not Applicable as HTLC is cancelled in worst case. (Can't be claimed by boltz till we claim it onchain by revealing the preimage)

### Steps:
- Give boltz an invoice and receive an address + amount
- Lockup amount in given address
- Wait for confirmation
- Boltz will automatically pay the invoice once confirmed

### Edge cases to test:

- [x] Invoice is expired/invalid
- [x] Hub is shut down after initiation of swap
- [x] Hub is shut down after lockup deposit
- [x] Hub is shut down while lockup is deposited

### Refunds

- [x] Rescue key
- [x] Process refund button in debug tools
- [x] Wrong amount locked up
- [x] Refund before swap is expired

## Auto Swap Out:
- [x] Runs in given frequency

# PR TODOs
- [x] Test wails
- [x] Test auto swaps once again
- [x] Test BIP85 mnemonic derivation
- [x] Add tests for BIP85 mnemonic derivation
- [x] Test all failure cases (expired invoice etc)
- [x] Test trying to refund a non-expired invoice
- [x] Check fees are correctly applied
  - [x] swap in
  - [x] swap out
- [x] Test rescue from mnemonic - works
- [x] Check failed swap outs flow
- [x] Process Refund button + api (debug tool)
- [x] List all swaps in frontend (debug tool)
- [x] Reconnect to boltz for updates
- [x] Specify fees while locking up btc for swap ins
- [x] Start swap service similar to how LNClient is done (to solve context.Background issue - this causes swaps to stay in pending state instead of failed)
- [x] Mark swap as settled if corresponding invoice settles
- [x] Move swaps transaction metadata (after creation of payment/invoice) to new swaps table
- [x] Find a way to not use `context.Background()` for one-off swaps - _can be a follow-up PR_
- [x] Ensure swap in failures can be refunded (test it end to end)
- [x] Test wails app
- [x] Re-test all flows (+ auto swaps)
- [x] Publish FAQ guide and share it with Boltz
- [x] ~Add a check for receive limit in Swap Ins (both auto + one-off) and spendable limit in Swap outs (one-off)~
- [x] ~Remove frontend for non LDK/LND users~ the buttons now look bad but low priority

### DIFF
--- a/swaps/swaps_service.go
+++ b/swaps/swaps_service.go
@@ -923,7 +923,6 @@ func (svc *swapsService) startSwapInListener(swap *db.Swap) {
 					Event: "nwc_swap_succeeded",
 					Properties: map[string]interface{}{
 						"swapType": constants.SWAP_TYPE_IN,
-						"swapId":   swap.SwapId,
 					},
 				})
 				return
@@ -1063,7 +1062,6 @@ func (svc *swapsService) startSwapOutListener(swap *db.Swap) {
 						Event: "nwc_swap_succeeded",
 						Properties: map[string]interface{}{
 							"swapType": constants.SWAP_TYPE_OUT,
-							"swapId":   swap.SwapId,
 						},
 					})
 					return


rolznz:
> What should we show for non LDK/LND users?

Can't we show nothing?

frontend/src/screens/channels/Channels.tsx
```diff
@@ -228,6 +206,13 @@ export default function Channels() {
                 </DropdownMenuContent>
               </DropdownMenu>
 
+              <ResponsiveButton
```
rolznz: This does not fit on mobile and to me the icon is not helpful. Can't we just keep the swap options in the advanced dropdown?

frontend/src/screens/wallet/Swaps.tsx
```diff
@@ -0,0 +1,499 @@
...
+          </div>
+          {swapType == "in" ? <SwapInForm /> : <SwapOutForm />}
+        </div>
+        {swapsSettings.enabled && <ActiveSwaps swaps={swapsSettings} />}
+      </div>
+    </>
+  );
+}
+
+// TODO: WE ONLY WISH TO ADD ONE-OFF SWAP-INS AT THE MOMENT
```
rolznz:
```suggestion
```

frontend/src/screens/wallet/Swaps.tsx
```diff
@@ -0,0 +1,499 @@
...
+    return <Loading />;
+  }
+
+  return (
+    <>
+      <AppHeader title="Swaps" />
+      <div className="flex gap-8 lg:gap-12 w-full flex-col lg:flex-row">
+        <div className="w-full max-w-lg">
+          <div className="flex items-center text-center text-foreground font-medium rounded-lg bg-muted p-1 mb-4">
+            <div
```
rolznz: image

I think it's weird that it defaults to swap out but that one is on the right. Which item is selected I think is not obvious either

frontend/src/screens/wallet/Swaps.tsx
```diff
@@ -0,0 +1,499 @@
...
+            Spending balance
+          </Label>
+        </div>
+        <div className="flex items-start space-x-2">
+          <RadioGroupItem value="external" id="external" className="shrink-0" />
+          <Label
+            htmlFor="external"
+            className="text-primary font-medium cursor-pointer"
+          >
+            External lightning wallet
```
rolznz: image

Can we remove the external option? it makes no sense. We want people to use Alby Hub, not another lightning wallet. The reason it's there for swapping out is so you can swap to cold storage.

frontend/src/screens/wallet/Swaps.tsx
```diff
@@ -0,0 +1,499 @@
...
+            <div className="flex justify-between items-center">
+              <span className="font-medium">Swap amount</span>
+              <span className="text-muted-foreground text-right">
+                {new Intl.NumberFormat().format(swaps.swapAmount)} sats
+              </span>
+            </div>
+            <div className="flex justify-between items-center">
+              <span className="font-medium">Fee</span>
+              <span className="text-muted-foreground text-right">
+                {swaps.albyServiceFee + swaps.boltzServiceFee}% + on-chain fees
```
rolznz: Swapping On-Chain -> lightning is cheaper than lightning -> on-chain in boltz.exchange (0.1% vs 0.5%) - we need to use a different endpoint to get the submarine swap fees. (+ implement submarine swaps)

frontend/src/screens/wallet/Swaps.tsx
rolznz: Ah, the figma uses bold for the text on the selected tab

frontend/src/screens/channels/Channels.tsx
```diff
@@ -285,7 +270,7 @@ export default function Channels() {
                     <Link
                       className="underline"
                       to="#"
-                      onClick={() => setSwapOutDialogOpen(true)}
+                      onClick={() => navigate("/wallet/swaps?type=out")}
```
rolznz: we can use a proper link `to` value now rather than `onClick` - this seems to break the query param too

frontend/src/screens/wallet/Swaps.tsx
```diff
@@ -0,0 +1,499 @@
...
+  const [loading, setLoading] = useState(false);
+  const [showAdvanced, setShowAdvanced] = useState(false);
+  const [recurringSwaps, setRecurringSwaps] = useState(false);
+
+  const onSubmit = async (e: React.FormEvent) => {
+    e.preventDefault();
+
+    try {
+      setLoading(true);
+      await request(`/api/wallet/${recurringSwaps ? "swaps" : "swap"}`, {
```
rolznz: I don't really like this, we are using different endpoints AND passing in different arguments. Can we split it up?

frontend/src/screens/wallet/Swaps.tsx
```diff
@@ -0,0 +1,499 @@
...
+        },
+        body: JSON.stringify({
+          swapType: "out",
+          swapAmount: parseInt(swapAmount),
+          destination: swapTo === "internal" ? onchainAddress : destination,
+          balanceThreshold: parseInt(balanceThreshold),
+        }),
+      });
+      toast({
+        title: recurringSwaps ? "Saved successfully" : "Initiated swap",
```
rolznz: I got "initated swap" toast but the swap payment failed. We really need to fix the backend to do a proper sync payment

api/api.go
```diff
@@ -580,6 +580,36 @@ func (api *api) GetAutoSwapsConfig() (*GetAutoSwapsConfigResponse, error) {
 	}, nil
 }
 
+func (api *api) InitiateSwap(ctx context.Context, initiateSwapRequest *InitiateSwapRequest) error {
```
rolznz: I think we should use different endpoints for swapping in and out, to reduce complexity

frontend/src/screens/wallet/Swaps.tsx
rolznz: Ah, from @im-adithya
> No, for swaps the HTLC is settled once we reveal the preimage to them, which happens when we broadcast a tx to claim from the lockup (which we do only after the lockup tx has been confirmed, hence taking 1 block time (or even more if fee increases). That's why we say swap "initiated" and not successful. How do you think we can update the user about the swap status?

frontend/src/screens/wallet/Swaps.tsx
rolznz: Ah, this is to not show "active recurring swap" when Swap out is selected, right?

im-adithya: No I initially added this because it also had option to select auto swap ins, I'll clean this up once we figure out swap outs and start adding support for swap ins

frontend/src/screens/wallet/Swaps.tsx
im-adithya: Yes, the swap-ins are not implemented yet, I added a todo in the SwapInForm to review fees for swap-ins

frontend/src/screens/wallet/Swaps.tsx
im-adithya: Done!

api/api.go
im-adithya: Done, now we have three endpoints
`/api/wallet/swap` -> for auto swaps ("GET", "POST", "DELETE")
`/api/wallet/swap-out` -> "POST"
`/api/wallet/swap-in` -> "POST" (to be added)

swaps/swaps_service.go
```diff
@@ -195,161 +285,400 @@ func (svc *swapsService) ReverseSwap(ctx context.Context, amount uint64, destina
...
-				if err != nil {
-					return err
-				}
+					txHex, err := claimTransaction.Serialize()
+					if err != nil {
+						logger.Logger.WithError(err).Error("Could not serialize claim transaction")
+						return
+					}
+
+					// TODO: Replace with LNClient method
```
rolznz: What's this TODO for? is it still valid?

swaps/swaps_service.go
```diff
@@ -125,46 +144,117 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
-func (svc *swapsService) StopAutoSwaps() {
-	if svc.cancelFn != nil {
-		logger.Logger.Info("Stopping swap service...")
-		svc.cancelFn()
-		logger.Logger.Info("swap service stopped")
+func (svc *swapsService) EnableAutoSwapIn(ctx context.Context, lnClient lnclient.LNClient) error {
+	// stop any existing swap in process
+	svc.StopAutoSwap(true, false)
+
+	// TODO: change threshold keys
```
rolznz: What needs doing here? reverting the change?

swaps/swaps_service.go
```diff
@@ -125,46 +144,117 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
+		}
+	}
+
 	var network, err = boltz.ParseChain(svc.cfg.GetNetwork())
 	if err != nil {
-		return err
+		return "", err
 	}
 
+	// TODO: use own keys
```
rolznz: What's the benefit for using our own keys? this needs to be unique per swap right? which derivation path would we use and even if we did it would it help with recovery?

swaps/swaps_service.go
im-adithya: Yup, it is suggested by boltz because it's good to not trust boltz and do it ourselves. But can be done later.

swaps/swaps_service.go
im-adithya: I wanted to confirm if the keys are good, I made a mistake by not prefixing auto swaps before with the word "Out", so now we have:

Swap-outs:
`AutoSwapBalanceThresholdKey`
`AutoSwapAmountKey`
`AutoSwapDestinationKey`

Swap-ins:
`AutoSwapInBalanceThresholdKey`
`AutoSwapInAmountKey`

swaps/swaps_service.go
im-adithya: I'm not well aware of how we handle the keys, but I thought maybe it's best to use the one we have internally. There's no benefit as such and we're also not relying on any external library like boltz here so it should be good :+1:

swaps/swaps_service.go
```diff
@@ -125,46 +150,118 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
+	svc.swapInCancelFn = cancelFn
+
+	return nil
 }
 
-func (svc *swapsService) ReverseSwap(ctx context.Context, amount uint64, destination string, lnClient lnclient.LNClient) error {
+func (svc *swapsService) SwapOut(ctx context.Context, amount uint64, destination string, lnClient lnclient.LNClient) (*SwapOutResponse, error) {
+	if destination == "" {
+		var err error
+		destination, err = svc.cfg.Get(config.OnchainAddressKey, "")
```
rolznz: I think we should get a new onchain address from the LNClient instead

swaps/swaps_service.go
rolznz: Cool. Could you update the comment to mention this?

swaps/swaps_service.go
rolznz: But we're unlikely to have automated swap in, and this is a minor breaking change that users will need to reconfigure swaps (without notification of it being removed). I'm not sure it's worth it

im-adithya: Cool, removed the comment :+1:

service/start.go
```diff
@@ -305,8 +303,13 @@ func (svc *service) StartApp(encryptionKey string) error {
 	return nil
 }
 
-func (svc *service) StartAutoSwaps() error {
-	return svc.GetSwapsService().EnableAutoSwaps(svc.ctx, svc.lnClient)
+func (svc *service) startSwapsService() {
+	swapsService := swaps.NewSwapsService(svc.ctx, svc.db, svc.cfg, svc.keys, svc.eventPublisher, svc.lnClient, svc.transactionsService)
+	svc.swapsService = swapsService
+	err := swapsService.EnableAutoSwapOut()
```
im-adithya: Should this be pushed inside `NewSwapsService`?

rolznz: I think so. It's currently inconsistent with how the other services are currently initialized

frontend/src/screens/wallet/swap/SwapOutStatus.tsx
```diff
@@ -0,0 +1,199 @@
...
+                      Paying lightning invoice...
+                    </div>
+                    <Divider color="border-green-600 dark:border-emerald-500" />
+                  </>
+                )
+              ) : (
+                <>
+                  <div className="flex items-center text-muted-foreground text-sm">
+                    <CircleCheckIcon className="w-5 h-5 mr-2 text-green-600 dark:text-emerald-500" />
+                    Swap invoice cancelled
```
rolznz: Why is this a success state? how do you get to it? 

And why is it called "Swap invoice cancelled"? this doesn't make sense

frontend/src/assets/suggested-apps/boltz.png
rolznz, line 1: I think this is not used anymore?

frontend/src/screens/channels/Channels.tsx
```diff
@@ -258,31 +234,34 @@ export default function Channels() {
...
-                    >
-                      <div className="mr-2 text-muted-foreground flex flex-row items-center">
-                        <ZapIcon className="w-4 h-4" />
-                        <ArrowRightIcon className="w-4 h-4" />
-                        <LinkIcon className="w-4 h-4" />
-                      </div>
-                      Swap out
-                    </DropdownMenuItem>
-                  </DropdownMenuGroup>
+                  {(info?.backendType === "LDK" ||
```
rolznz: We can use `hasChannelManagement` (search for other cases)

frontend/src/screens/wallet/index.tsx
```diff
@@ -90,28 +121,21 @@ function Wallet() {
...
-              Receive
-            </Button>
-          </Link>
-          <Link to="/wallet/send">
-            <Button size="lg" className="w-full">
-              <ArrowUpIcon className="h-4 w-4 shrink-0 mr-2" />
-              Send
-            </Button>
-          </Link>
+          {(info?.backendType === "LDK" || info?.backendType === "LND") && (
```
rolznz: use `hasChannelManagement` (check all usages)

frontend/src/screens/wallet/index.tsx
```diff
@@ -24,14 +26,34 @@ function Wallet() {
...
+      <AppHeader
+        title="Wallet"
+        description=""
+        contentRight={
+          <>
+            <ResponsiveButton
+              icon={ArrowDownIcon}
+              text="Receive"
+              size="lg"
+              onClick={() => navigate("/wallet/receive")}
```
rolznz: can't these be normal links? let's avoid using navigate for standard navigation as it breaks browser UX

swaps/swaps_service.go
```diff
@@ -125,254 +183,1057 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
+			"paymentHash": paymentHash,
+		}).Error("Failed to save swap")
+		return nil, err
+	}
+
+	logger.Logger.WithField("swapId", swap.Id).Info("Swap created")
+
+	boltzWs := svc.boltzApi.NewWebsocket()
+	if err := boltzWs.Connect(); err != nil {
+		svc.markSwapState(&dbSwap, constants.SWAP_STATE_FAILED)
```
rolznz: sometimes we set this and sometimes not. It should be handled by the deferred function, right? can you double check that?

swaps/swaps_service.go
```diff
@@ -125,254 +183,1057 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
+		return nil, fmt.Errorf("could not connect to Boltz websocket: %w", err)
+	}
+
+	if err := boltzWs.Subscribe([]string{swap.Id}); err != nil {
+		_ = boltzWs.Close()
+		svc.markSwapState(&dbSwap, constants.SWAP_STATE_FAILED)
+		return nil, err
+	}
+
+	go svc.startSwapOutListener(&dbSwap, boltzWs)
```
rolznz: I'm don't think we handle the websocket well in both these listener functions (swap out and in).

Shouldn't we handle reconnections? is there a recommended way to do this?

I see there is a `Connected()` method and a `Reconnect()` method available.

swaps/swaps_service.go
```diff
@@ -125,254 +183,1057 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
+
+	boltzWs := svc.boltzApi.NewWebsocket()
+	if err := boltzWs.Connect(); err != nil {
+		svc.markSwapState(&dbSwap, constants.SWAP_STATE_FAILED)
+		return nil, fmt.Errorf("could not connect to Boltz websocket: %w", err)
+	}
+
+	if err := boltzWs.Subscribe([]string{swap.Id}); err != nil {
+		_ = boltzWs.Close()
+		svc.markSwapState(&dbSwap, constants.SWAP_STATE_FAILED)
```
rolznz: here too, is marking the swap state needed since we have the deferred function?

swaps/swaps_service.go
```diff
@@ -125,254 +183,1057 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
+	}
+
+	network, err := boltz.ParseChain(svc.cfg.GetNetwork())
 	if err != nil {
 		return err
 	}
 
-	tree := swap.SwapTree.Deserialize()
-	if err := tree.Init(boltz.CurrencyBtc, true, ourKeys, boltzPubKey); err != nil {
+	swapTransactionResp, err := svc.boltzApi.GetSwapTransaction(swapId)
```
rolznz, line 574: is it good to rely on Boltz here? shouldn't we know this already from our own swaps DB? (maybe we can add a TODO to revisit this?)

rolznz: can this be moved into the conditional below? (`if swap.LockupTxId == ""`)

swaps/swaps_service.go
```diff
@@ -125,254 +183,1057 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
+		return
+	}
+	if len(swaps) == 0 {
+		return
+	}
+
+	logger.Logger.WithField("count", len(swaps)).Info("Resuming pending swaps...")
+
+	for _, swap := range swaps {
+		var boltzWs *boltz.Websocket
```
rolznz: We should actually only need one websocket in total shouldn't we?

swaps/swaps_service.go
rolznz: Oh, I saw you use `Connect()` already. Is that fine? did you manage to test it somehow?

swaps/swaps_service.go
```diff
@@ -125,254 +183,1057 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
+				}).Error("Swap in context cancelled")
+				return
+			case update, ok := <-updatesCh:
+				if !ok {
+					logger.Logger.WithField("swapId", swap.SwapId).Error("Boltz websocket closed unexpectedly, reconnecting...")
+					if err := boltzWs.Connect(); err != nil {
+						logger.Logger.WithError(err).WithFields(logrus.Fields{
+							"swapId": swap.SwapId,
+						}).Error("Reconnection to Boltz websocket failed")
+						return
```
rolznz: Is it ok that we just return here? should we retry instead?

rolznz: (for both swap directions)

swaps/swaps_service.go
```diff
@@ -125,254 +183,1057 @@ func (svc *swapsService) EnableAutoSwaps(ctx context.Context, lnClient lnclient.
...
-				if err != nil {
-					logger.Logger.WithField("swapId", swap.Id).Info("Couldn't probe invoice payment, terminating swap")
-					return err
+func (svc *swapsService) startSwapInListener(swap *db.Swap, boltzWs *boltz.Websocket) {
+	network, err := boltz.ParseChain(svc.cfg.GetNetwork())
+	if err != nil {
+		logger.Logger.WithError(err).WithFields(logrus.Fields{
+			"swapId": swap.SwapId,
+		}).Error("Failed to parse network")
+		return
```
rolznz, line 815: In both the listeners we just return if there's an error. Shouldn't we also set the swap state? should we use defer like we do elsewhere?

frontend/src/screens/wallet/swap/SwapOutStatus.tsx
im-adithya: 🤔 this is not success state, it's in the false expression of `swapStatus !== "FAILED"` right? And that would mean the swap failed and the hold invoice has been cancelled, hence the name

frontend/src/screens/wallet/index.tsx
im-adithya: They should be, thanks for catching! Updated other instance too (only one, in subwallet page)

swaps/swaps_service.go
im-adithya: I originally used one, but that was causing some weird issues (one state being missed for a swapId when there are two pending swap Ids) so I changed it to be one, shouldn't be an issue since there won't be (m)any pending ones ;)

swaps/swaps_service.go
im-adithya: We mark it when:
- We can't connect to boltz websocket the first time itself (first time meaning initiation, not resubscription after restart)
- When we receive payment error in `paymentErrorCh`

We don't mark it when:
- Context is done after starting `startSwapOutListener` because then we want to resubscribe on restart
- Reconnection or resubscription fails in `startSwapOutListener` (same reason as above)

I missed to update it as failed when errors happen before running the for loop (like parsing network) and after (while constructing txs after lockup confirmation etc.)

Did that now (in https://github.com/getAlby/hub/pull/1371/commits/f5ca03b9eff6777d747a42eb33cfb1a6cf01765f), but in swap outs, marking as failed shows in the swaps page that the hold invoice is cancelled but in reality it will still be in pending and will be cancelled from boltz side after expiry (because we couldn't construct and broadcast claim in time).

These errors shouldn't typically happen, but how should we handle such cases? Can we cancel the hold invoice from our side?

swaps/swaps_service.go
im-adithya: Fixed (see above)

swaps/swaps_service.go
im-adithya: We are mainly doing this to get the lockup transaction hex which is used to construct the tx below (and we can't assume to have it always since it can be done from external wallet too), we can fetch this (if `swap.LockupTxId != ""` using mempool api too somehow but I don't see much difference. But I'll add this as a comment for reference as it looks as if we are doing it for txId

swaps/swaps_service.go
im-adithya: Fixed (see https://github.com/getAlby/hub/pull/1371#discussion_r2192539928)

swaps/swaps_service.go
rolznz: I don't think this answers my question. We have an error here, when we return an error the deferred function should run and update the swap state. So why do we manually update it here?

frontend/src/screens/wallet/swap/SwapOutStatus.tsx
rolznz: You are displaying a `CircleCheckIcon` with green colors. I don't think we should show success colors and content when the swap failed.

I think we should change the title to "Swap failed" too, "swap invoice cancelled" is too complicated

swaps/swaps_service.go
rolznz: @im-adithya hmm, can you verify that and maybe ask Boltz? that doesn't sound good.

swaps/swaps_service.go
rolznz: Ah ok, thanks!

rolznz:
@im-adithya can you check we're on the latest boltz client version in this PR https://github.com/getAlby/hub/pull/1471

swaps/swaps_service.go
im-adithya: Sorry I went on a tangent, addressed now. But still the question I was asking at the end was about how to handle failed swap out cases before broadcasting claim, because then the frontend would say failed but the tx list would still keep showing a pending tx since boltz doesn't know we failed to broadcast claim (till expiry)

im-adithya: Discussed on slack, will implement retries as network issue is a real concern here.

frontend/src/screens/wallet/swap/SwapOutStatus.tsx
im-adithya: Done, looks better now:

Screenshot 2025-07-09 at 1 17 41 PM

swaps/swaps_service.go
im-adithya: Done

swaps/swaps_service.go
im-adithya: Fixed

swaps/swaps_service.go
im-adithya: Resolved (for pending swaps we now only use a single websocket connection), we can do this similarly for individual swaps too but it's not possible currently because we can't chain swap id subscriptions right now on a single boltz websocket connection. Reached out to them on discord.

swaps/swaps_service.go
im-adithya: Done, we keep trying till it connects before starting this go routine

swaps/swaps_service.go
```diff
@@ -743,31 +769,55 @@ func (svc *swapsService) subscribePendingSwaps() {
...
+			go svc.startSwapInListener(&swap, ch)
 		case constants.SWAP_TYPE_OUT:
-			go svc.startSwapOutListener(&swap, boltzWs)
+			go svc.startSwapOutListener(&swap, ch)
 		}
 	}
+
+	go func() {
+		for update := range boltzWs.Updates {
+			if ch, ok := chans[update.Id]; ok {
```
rolznz: why don't you handle if ok is false here?

rolznz: it should reconnect the boltz websocket

swaps/swaps_service.go
```diff
@@ -840,143 +890,123 @@ func (svc *swapsService) startSwapInListener(swap *db.Swap, boltzWs *boltz.Webso
...
-					// since this is not a mandatory step, we redeclare err
-					// to avoid running the defer function in case of failure
-					claimDetails, err := svc.boltzApi.GetSwapClaimDetails(swap.SwapId)
-					if err != nil {
-						logger.Logger.WithError(err).WithFields(logrus.Fields{
-							"swapId": swap.SwapId,
-						}).Error("Could not get claim details from Boltz")
-						return
-					}
+				// since this is not a mandatory step, we redeclare err
```
rolznz: if it's not a mandatory step why does it `return`? this means it stops the whole swap flow

rolznz: why do we write all this code if it's not needed? this doesn't make sense to me.

swaps/swaps_service.go
```diff
@@ -840,143 +890,123 @@ func (svc *swapsService) startSwapInListener(swap *db.Swap, boltzWs *boltz.Webso
...
-						}).Error("Could not process refund")
-					}
+						"swapId":        swap.SwapId,
+						"receiveAmount": amount,
+					}).WithError(err).Error("Failed to save received amount to swap")
 					return
 				}
+				logger.Logger.WithField("swapId", swap.SwapId).Info("Swap succeeded")
+				svc.eventPublisher.Publish(&events.Event{
+					Event: "nwc_swap_succeeded",
```
rolznz, line 923: What is this event for? I can't see it consumed

im-adithya: This is to notify the user via email (I'm not sure if this is how it's done, I followed channel closure case `nwc_channel_closed`)

swaps/swaps_service.go
rolznz: @im-adithya verified that boltz handles reconnects

swaps/swaps_service.go
rolznz: Do we want to notify of successful swaps via email? we need to do some additional steps:
1. Add it to the oauth event whitelist
2. Add the new event to the README
3. Remove fields from the event that invasion of privacy, for example the swap amount

Once we start publishing these events in production, we can setup emails

swaps/swaps_service.go
im-adithya: Removed this, I thought it's good to cooperate with boltz, but even in regtest we don't receive this update (I think they claim it directly using script path without waiting for the partial sig). We now check if we receive `invoice.paid` state and mark the swap as successful ✅ 

> Once the user's lockup transaction is included in a block (or found in the mempool if [0-conf](https://api.docs.boltz.exchange/0-conf.html) applies), Boltz will try to pay the invoice provided by the user.

swaps/swaps_service.go
im-adithya: Done ✅ (we only send swap type and swap id)

swaps/swaps_service.go
```diff
@@ -1062,10 +1061,8 @@ func (svc *swapsService) startSwapOutListener(swap *db.Swap, updateCh chan boltz
 					svc.eventPublisher.Publish(&events.Event{
 						Event: "nwc_swap_succeeded",
 						Properties: map[string]interface{}{
-							"swapType":    constants.SWAP_TYPE_OUT,
-							"swapId":      swap.SwapId,
-							"amount":      swap.ReceiveAmount,
-							"destination": swap.DestinationAddress,
+							"swapType": constants.SWAP_TYPE_OUT,
+							"swapId":   swap.SwapId,
```
rolznz: Should we consider the swap ID private too?

swaps/swaps_service.go
```diff
@@ -926,7 +926,7 @@ func (svc *swapsService) startSwapInListener(swap *db.Swap, updateCh chan boltz.
 					"swapId":     swap.SwapId,
 					"lockupTxId": swap.LockupTxId,
 				}).Info("Lockup transaction confirmed in mempool")
-			case boltz.InvoicePaid:
+			case boltz.InvoicePaid, boltz.TransactionClaimed:
```
rolznz: why do we need both cases?

swaps/swaps_service.go
im-adithya: Removed

rolznz (approved):
tACK

PR #1377, opened 2025-06-04
fix: add nil check for node policy

im-adithya:
If you force-close a channel and refresh your node page, the list of channels doesn't load because of a nil pointer dereference. This PR solves that by adding a `nil` check


im-adithya:
> does it aim to solve it or does it solve it? 😀

Haha, never realised while typing that 😄

PR #1378, opened 2025-06-04
fix: add prop to not render sidebar trigger in header

### LINKED ISSUE #1359: Bug: app crashes after doing migration
User reported when they did a migration from desktop to Alby Hub, the page crashed due to a sidebar error.

Image

im-adithya:
Fixes #1359

Adds `hideTrigger` prop to not render it to avoid crashing

### DIFF
--- a/frontend/src/components/AppHeader.tsx
+++ b/frontend/src/components/AppHeader.tsx
@@ -7,13 +7,19 @@ type Props = {
   description?: string | ReactElement;
   contentRight?: React.ReactNode;
   breadcrumb?: boolean;
+  hideTrigger?: boolean;
 };
 
-function AppHeader({ title, description = "", contentRight }: Props) {
+function AppHeader({
+  title,
+  description = "",
+  contentRight,
+  hideTrigger,
+}: Props) {
   return (
     <>
       <header className="flex flex-row items-center border-b border-border pb-4 gap-2">
-        <SidebarTrigger className="-ml-1 md:hidden" />
+        {!hideTrigger && <SidebarTrigger className="-ml-1 md:hidden" />}
         <Separator orientation="vertical" className="mr-2 h-4 md:hidden" />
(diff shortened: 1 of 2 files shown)


frontend/src/components/AppHeader.tsx
```diff
@@ -7,13 +7,19 @@ type Props = {
...
+  title,
+  description = "",
+  contentRight,
+  hideTrigger,
+}: Props) {
   return (
     <>
       <header className="flex flex-row items-center border-b border-border pb-4 gap-2">
-        <SidebarTrigger className="-ml-1 md:hidden" />
+        {!hideTrigger && <SidebarTrigger className="-ml-1 md:hidden" />}
```
rolznz: minor: this is a double-negative. It would be better to have `showTrigger && ...`

frontend/src/components/AppHeader.tsx
```diff
@@ -7,13 +7,19 @@ type Props = {
   description?: string | ReactElement;
   contentRight?: React.ReactNode;
   breadcrumb?: boolean;
+  hideTrigger?: boolean;
```
rolznz: is this the best way? or can we not use the App header from the CreateNodeMigrationFileSuccess page?

PS. this variable naming is vague - what trigger?

im-adithya: I thought of doing that but went with less code option. Wdyt? `showSidebarTrigger` or adding header?

rolznz: I guess it's ok. I would use `addSidebarTrigger` or `includeSidebarTrigger` rather than `show` because a trigger is not even visible, right?

im-adithya: It is only visible in mobile, because sidebar is present by default in >md widths

rolznz (approved):
tACK

Issue #1376, opened 2025-06-04
When changing unlock password, also regenerate JWT Secret


rolznz:
This will instantly invalidate any existing JWT tokens. e.g. in case you are worried one of the machines you logged in was compromised (or use an experimental hub developer API token).

rolznz:
Hi @achaljhawar sure, go ahead!

PR #1383, opened 2025-06-05
feat: Regenerate JWT Secret on password change

achaljhawar:
fixes: #1376 
The `ChangeUnlockPassword` function in `config/config.go` has been updated to:
- Generate a new random JWT secret.
- Store this new secret in the database.
- Update the active JWT secret in the config.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -269,6 +269,19 @@ func (cfg *config) ChangeUnlockPassword(currentUnlockPassword string, newUnlockP
 		return err
 	}
 
+	newSecret, err := randomHex(32)
+	if err != nil {
+		logger.Logger.WithError(err).Error("failed to generate new JWT secret")
+	} else {
+		err = cfg.SetUpdate("JWTSecret", newSecret, "")
+		if err != nil {
+			logger.Logger.WithError(err).Error("failed to save new JWT secret")
+		} else {
+			cfg.JWTSecret = newSecret
+			logger.Logger.Info("Successfully regenerated JWT secret after password change")
+		}
+	}
+
 	return nil
 }
 


config/config.go
```diff
@@ -269,6 +269,19 @@ func (cfg *config) ChangeUnlockPassword(currentUnlockPassword string, newUnlockP
 		return err
 	}
 
+	newSecret, err := randomHex(32)
```
rolznz, line 272: Could you move all the new code into the DB transaction above underneath the for loop?

config/config.go
```diff
@@ -269,6 +269,19 @@ func (cfg *config) ChangeUnlockPassword(currentUnlockPassword string, newUnlockP
...
+	if err != nil {
+		logger.Logger.WithError(err).Error("failed to generate new JWT secret")
+	} else {
+		err = cfg.SetUpdate("JWTSecret", newSecret, "")
+		if err != nil {
+			logger.Logger.WithError(err).Error("failed to save new JWT secret")
+		} else {
+			cfg.JWTSecret = newSecret
```
rolznz, line 280: An existing issue but I don't like how we set the `JWTSecret` manually here. Actually I think we should remove this from the config struct and inside
`func (cfg *config) GetJWTSecret()` change to fetch from the config just like we do with `func (cfg *config) GetRelayUrl()` below. Then it's consistent with how we fetch other config data

rolznz:
> yes , but should I write separate tests for it?
@achaljhawar it might be a bit difficult. I can have a look if we can make one.

Could you create a new PR, this time on a separate branch? I can't push edits because your PR is from the `master` branch.

achaljhawar:
ohk

PR #1385, opened 2025-06-06
fix: re-queue failed consumed notification events

rolznz:
fixes https://github.com/getAlby/hub/issues/1212

### DIFF
--- a/nip47/nip47_service.go
+++ b/nip47/nip47_service.go
@@ -76,7 +76,14 @@ func (svc *nip47Service) StartNotifier(relay *nostr.Relay) {
 				// relay disconnected
 				return
 			case event := <-svc.nip47NotificationQueue.Channel():
-				nip47Notifier.ConsumeEvent(relay.Context(), event)
+				logger.Logger.WithField("event", event).Debug("Consuming event from notification queue")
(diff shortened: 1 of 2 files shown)


nip47/nip47_service.go
```diff
@@ -76,7 +76,14 @@ func (svc *nip47Service) StartNotifier(relay *nostr.Relay) {
...
 			case event := <-svc.nip47NotificationQueue.Channel():
-				nip47Notifier.ConsumeEvent(relay.Context(), event)
+				logger.Logger.WithField("event", event).Debug("Consuming event from notification queue")
+				err := nip47Notifier.ConsumeEvent(relay.Context(), event)
+				if err != nil {
+					logger.Logger.WithError(err).WithField("event", event).Error("Failed to consume event from notification queue")
+					// wait and then re-add the item to the queue
+					time.Sleep(5 * time.Second)
+					svc.nip47NotificationQueue.AddToQueue(event)
+				}
```
im-adithya, line 86: This would mean if any app in the for loop of `notifySubscribers` fails to send a notif, everything will be sent again. But I think this is better than not sending notifs :+1:

rolznz: hmm, good point. I'm not sure how we can easily solve that, but good to keep in mind.

Issue #1382, opened 2025-06-05
Create invoice for a different app id


rolznz:
This allows the getalby.com connection to be able to make invoices for other apps, enabling multiple lightning addresses.

There should be a way to API call to create a lightning address on getalby.com (including the app ID and the app name)

Issue #1387, opened 2025-06-06
Create lightning address for sub-wallet


rolznz:
Follow-up of https://github.com/getAlby/hub/issues/1382

When creating a sub-wallets you should be able to create a lightning address for it. The Alby account will have a main lightning address + N sub-wallet lightning addresses and will pass the `metadata: { app_id }` in the make_invoice request when the LNURL callback is requested.

- [ ] UI on Sub-wallet page to create lightning address (+ display on created sub wallet - possibly store the lightning address in the app metadata?)
- [ ] execute API call to getalby.com to create a lightning address, providing `app_name` and `app_id`

matjaz:
You can use your own domain for lightning addresses using NWC /w https://github.com/matjaz/btc-service

PR #1388, opened 2025-06-06
feat: add app id override to make invoice

rolznz:
Closes https://github.com/getAlby/hub/issues/1382

Simply pass `metadata: {app_id: 2}` and the invoice will be for app 2 instead, no matter what app initiated the request.

Does not do the lightning address part (see https://github.com/getAlby/hub/issues/1387)

### DIFF
--- a/transactions/transactions_service.go
+++ b/transactions/transactions_service.go
@@ -139,6 +139,16 @@ func NewTransactionsService(db *gorm.DB, eventPublisher events.EventPublisher) *
 }
 
 func (svc *transactionsService) MakeInvoice(ctx context.Context, amount uint64, description string, descriptionHash string, expiry uint64, metadata map[string]interface{}, lnClient lnclient.LNClient, appId *uint, requestEventId *uint) (*Transaction, error) {
+	logger.Logger.WithFields(logrus.Fields{
+		"app_id":           appId,
+		"request_event_id": requestEventId,
+		"amount":           amount,
+		"description":      description,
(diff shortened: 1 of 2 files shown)


bumi:
oh wow. how does it feel? do you think that's good?

rolznz:
@bumi I think it's good, the only thing I worry about is that if a developer/app sets this for some other reason then they could mistakenly activate it. But I think the odds are very small.

transactions/transactions_service.go
```diff
@@ -152,6 +162,16 @@ func (svc *transactionsService) MakeInvoice(ctx context.Context, amount uint64,
 		}
 	}
 
+	if metadata["app_id"] != nil {
+		overwriteAppIdType, ok := metadata["app_id"].(float64)
+		if !ok {
+			return nil, errors.New("failed to overwrite app ID")
+		}
+		overwriteAppId := uint(overwriteAppIdType)
```
bumi, line 170: do we need to validate that this app exists? 
what if an invalid id is set? (or the app is deleted by now)

rolznz: I tested this case and I think it's fine without further validation. It will just be treated as a normal invoice not for a sub-wallet. (Kind of similar if you delete an app, the app will be deleted but transactions will remain).

A very small side effect I found: if you create an invoice for an app ID that doesn't exist yet (e.g. I had 2 local apps but made one for id:3) when the app is created, the app will suddenly have that transaction associated with it.

PR #1389, opened 2025-06-06
feat: Regenerate JWT Secret on password change

achaljhawar:
fixes: #1376 

The `ChangeUnlockPassword` function in `config/config.go` has been updated to:

- Generate a new random JWT secret.
- Store this new secret in the database.
- Update the active JWT secret in the config.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -269,6 +269,19 @@ func (cfg *config) ChangeUnlockPassword(currentUnlockPassword string, newUnlockP
 		return err
 	}
 
+	newSecret, err := randomHex(32)
+	if err != nil {
+		logger.Logger.WithError(err).Error("failed to generate new JWT secret")
+	} else {
+		err = cfg.SetUpdate("JWTSecret", newSecret, "")
+		if err != nil {
+			logger.Logger.WithError(err).Error("failed to save new JWT secret")
+		} else {
+			cfg.JWTSecret = newSecret
+			logger.Logger.Info("Successfully regenerated JWT secret after password change")
+		}
+	}
+
 	return nil
 }
 


achaljhawar:
@rolznz I have created a new branch

config/config.go
```diff
@@ -269,6 +282,18 @@ func (cfg *config) ChangeUnlockPassword(currentUnlockPassword string, newUnlockP
 		return err
 	}
 
+	newSecret, err := randomHex(32)
```
rolznz: This is still not inside the DB transaction

config/config.go
```diff
@@ -105,20 +104,29 @@ func (cfg *config) init(env *AppConfig) error {
...
+		err := cfg.SetIgnore("JWTSecret", cfg.Env.JWTSecret, "")
 		if err != nil {
-			logger.Logger.WithError(err).Error("failed to generate JWT secret")
-			return err
+			logger.Logger.WithError(err).Error("failed to set JWT secret from env")
 		}
-		err = cfg.SetIgnore("JWTSecret", hex, "")
-		if err != nil {
-			return err
+	} else {
```
rolznz: This code got a lot more complicated. We should not have an if-else. Could you see if you could change it?


1. check `cfg.Get("JWTSecret", "")` is empty before doing anything, rather than checking cfg.Env.JWTSecret.
2. create a local variable jwtSecret := cfg.Env.JWTSecret
3. `if jwtSecret == ""` then generate one and update `jwtSecret`.
4. `cfg.SetIgnore("JWTSecret", jwtSecret, "")`
	
This way we do not need any `else` and should be able to remove lots of lines of code.

achaljhawar:
I'll work on simplifying this

config/config.go
achaljhawar: is this problem solved?

rolznz (approved):
tACK, thanks for the PR!

Issue #1025, opened 2025-01-26
Unable to load when reverse proxy is configured with a location other than root


pabpas:
First time trying Alby Hub and I noticed that if I serve it (with nginx) to the outside in `mydomain.com/alby`, the browser will show an error about `malformed jwt` and the main page will not load. Setting the location to / solves the issue (`mydomain.com/`).

rolznz:
Hi, yeah unfortunately I don't think it will work not on the route. `alby.mydomain.com` would work if you are ok with adding another DNS entry

daum3ns:
i am confused because my redirect removes the URI. This is also visible in the AlbyHub logs 

`{"host":"removed","level":"info","msg":"handled API request","remote_ip":"xx.xx.xx.xx","request_id":"wfkXCdKaybzbrkeZKxviGvMBrPjDpqAw","status":200,"time":"2025-02-23T11:58:28Z","uri":"/","user_agent":"Mozilla/5.0 (X11; Linux x86_64; rv:135.0) Gecko/20100101 Firefox/135.0"}`

as you can see the URI is /, the status code is 200, but the page is just blank.. No indication of an error in the logs.. 
when i do a request to the /api  i get the same "missing or malformed jwt"" error as pabpas

rolznz:
Hi, what version are you running? did you start/unlock Alby Hub?

daum3ns:
HI
I use the latest version from github (v1.14.3). No I just started the binary and wanted to access the site but only see a white page. However, i just realized that, when I make a call to the correct port und using plain HTTP (i.e without going through the proxy) i can see the page. 
Do you think i need to go through a set up process of some kind before it can work (i'm new to albyhub)?

rolznz:
@daum3ns what error code do you get on the white page? did you setup the proxy correctly for https and correctly configure your subdomain?

daum3ns:
let me try to explain better :) my idea was that the reverse proxy (nginx) terminates the tls connection. 
i have the following start script for albyhub:

```bash
#!/bin/bash

WORK_DIR="/mnt/opt/albyhub/data" LOG_EVENTS=true LDK_GOSSIP_SOURCE="" PORT="8029" LN_BACKEND_TYPE=LND LND_ADDRESS=127.0.0.1:10019 LND_CERT_FILE=<path_to_cert> LND_MACAROON_FILE=<path_to_macaroon> /mnt/opt/albyhub/bin/albyhub
```

[...]

PR #1393, opened 2025-06-09
feat: add base path to front end

8de2fdb0:
Small MR adding the ability to serve the hub behind a proxy.

- add docker build arg to set base path frontend is served under
- set vite.config.ts base value based on docker build arg
- update http variant request function to inject base path during calls to /api and /images
- update how static assets are loaded from public folder to inject base path during vite build

To use a base path build the docker image with --build-arg BASE_PATH=/hub/. All assets and api calls from the frontend will use that base path as prefix.

The backend service is not using the base path, when using a proxy the base path has to be stripped.

#1025

Let me know what you think.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,12 @@
 FROM node:20-alpine as frontend
+
+# Set the base path for the frontend build
+# This can be overridden at build time with --build-arg BASE_PATH=<url>
+# Allows to build a frontend that can be served from a subpath, e.g. /hub/
+ARG BASE_PATH="/"
 WORKDIR /build
 COPY frontend ./frontend
+RUN echo "Building frontend with base path $BASE_PATH"
 RUN cd frontend && yarn install --network-timeout 3000000 && yarn build:http
 
 FROM golang:1.24 as builder
--- a/frontend/platform_specific/http/src/utils/request.ts
+++ b/frontend/platform_specific/http/src/utils/request.ts
@@ -1,9 +1,24 @@
 import { getAuthToken } from "src/lib/auth";
 import { ErrorResponse } from "src/types";
 
+const BASE_URL = import.meta.env.BASE_URL;
+const PREFIXES = ["/api", "/images"];
+
+function startsWithPrefix(path: string, prefixes: string[]): boolean {
+  return prefixes.some((prefix) => path.startsWith(prefix));
+}
+
 export const request = async <T>(
   ...args: Parameters<typeof fetch>
 ): Promise<T | undefined> => {
+  if (
+    BASE_URL !== "/" &&
+    typeof args[0] === "string" &&
+    startsWithPrefix(args[0], PREFIXES)
+  ) {
+    args[0] = BASE_URL + args[0].slice(1);
+  }
+
   const token = getAuthToken();
   if (token) {
     if (!args[1]) {
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -13,7 +13,7 @@ import routes from "src/routes.tsx";
 import { isHttpMode } from "src/utils/isHttpMode";
 
 const createRouterFunc = isHttpMode() ? createBrowserRouter : createHashRouter;
-const router = createRouterFunc(routes);
+const router = createRouterFunc(routes, { basename: import.meta.env.BASE_URL });
 
 function App() {
   const { data: info } = useInfo();
--- a/frontend/src/components/connections/AlbyConnectionCard.tsx
+++ b/frontend/src/components/connections/AlbyConnectionCard.tsx
@@ -46,6 +46,9 @@ import { useAlbyMe } from "src/hooks/useAlbyMe";
 import { LinkStatus, useLinkAccount } from "src/hooks/useLinkAccount";
 import { App, BudgetRenewalType } from "src/types";
 
+import AlbyAccountDarkSVG from "public/images/illustrations/alby-account-dark.svg";
+import AlbyAccountLightSVG from "public/images/illustrations/alby-account-light.svg";
+
 function AlbyConnectionCard({ connection }: { connection?: App }) {
   const { data: albyMe } = useAlbyMe();
   const { loading, linkStatus, loadingLinkStatus, linkAccount } =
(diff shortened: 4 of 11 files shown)


frontend/platform_specific/http/src/utils/request.ts
```diff
@@ -1,9 +1,24 @@
 import { getAuthToken } from "src/lib/auth";
 import { ErrorResponse } from "src/types";
 
+const BASE_URL = import.meta.env.BASE_URL;
+const PREFIXES = ["/api", "/images"];
+
+function startsWithPrefix(path: string, prefixes: string[]): boolean {
```
rolznz, line 7: I don't think the changes in this file should be needed - requests always go to `/api` (this is a special request function to abstract the differences between HTTP and Wails APIs)

rolznz: Can we just do:
`const BASE_URL = import.meta.env.BASE_URL || "";`
`args[0] = BASE_URL + args[0];`

Dockerfile
```diff
@@ -1,6 +1,12 @@
 FROM node:20-alpine as frontend
+
+# Set the base path for the frontend build
+# This can be overridden at build time with --build-arg BASE_PATH=<url>
+# Allows to build a frontend that can be served from a subpath, e.g. /hub/
+ARG BASE_PATH="/"
```
rolznz, line 6: there is `BASE_PATH` and `BASE_URL`, we only need one, right?

frontend/src/App.tsx
```diff
@@ -13,7 +13,7 @@ import routes from "src/routes.tsx";
 import { isHttpMode } from "src/utils/isHttpMode";
 
 const createRouterFunc = isHttpMode() ? createBrowserRouter : createHashRouter;
-const router = createRouterFunc(routes);
+const router = createRouterFunc(routes, { basename: import.meta.env.BASE_URL });
```
rolznz, line 16: Minor: I don't think we should apply the base URL to the hash router. (Used by the Wails/desktop app)

Dockerfile
rolznz: Currently `BASE_URL` is an already declared property used by the hub go app.
Also note there is a `FRONTEND_URL` used by the go app too.

frontend/src/App.tsx
8de2fdb0: Good point.

I can update it to following snippet;
```ts
const [createRouterFunc, createRouterOpts] = isHttpMode()
  ? [createBrowserRouter, { basename: import.meta.env.BASE_URL }]
  : [createHashRouter, undefined];
const router = createRouterFunc(routes, createRouterOpts);

```

Dockerfile
8de2fdb0: So `BASE_URL` is a vite specific, I think they call it URL since they allow to set the whole URL in cases resources come from a specific server (CDN for example). This feature is focused on allowing to server the frontend behind a proxy using sub path. That case is usually named `base_path` and I wanted to keep the top level variable reflecting that specific use case. URL would suggest to allow to change the whole URL.

I could't see a specific group set up for the `/api` path in the echo framework boilerplate, and a quick search seems to suggest `FRONTEND_URL` and `BASE_PATH` is used when connecting the alby account with the local hub instance ( from default `https://getalby.com/`). 

Yes that has a potential for creating confusion. Maybe the `BASE_PATH` should be called `VITE_FRONTEND_BASE_PATH`. 
Since you have a better understanding of the already existing variables, do you have any suggestion that helps differentiate ?

And I would really not focus on `BASE_URL` since it is a vite internal and is a constant injected into the build artefact. Runtime env variables won't reflect on that, it is just build time relevant.

frontend/platform_specific/http/src/utils/request.ts
8de2fdb0: > I don't think the changes in this file should be needed - requests always go to /api (this is a special request function to abstract the differences between HTTP and Wails APIs)

Running `vite build` will inject the `import.meta.env.BASE_URL` to all resources loaded from the server, but not into the API requests, there is probably a better way to have vite modify the API urls from where they are used, but injecting it here seemed simpler then refactoring all the individual api hooks. 

I only added it to the `platform_specific/http` variant since that feature won't make any sense with `wails`.

> Can we just do:
const BASE_URL = import.meta.env.BASE_URL || "";
args[0] = BASE_URL + args[0];

The const `import.meta.env.BASE_URL` is injected from vite during build,  see https://vite.dev/guide/build.html#public-base-path.
Looking at the docs for the specific [vite config option](https://vite.dev/config/shared-options.html#base) it is default set to "/",  my changes also added that value in the config in case the env variable `BASE_PATH` is unset. So it must be always "/" at that point.

About the path concatenation, yes I think so, but it would need to be some kind of  `path.join`  so `/base_url/` + `/api` will result in `/base_url/api`. The `slice[1]` kind of expects the first char to be `/` and removes it and since only parameters starting with `/api` and `/images` are matched that will always be true. But that will fail in case `BASE_URL` is `/hub`.  

Best is probably to add a small joinPath function.

Dockerfile
rolznz: Thanks for the details, after looking into it I understand this is how vite works. And the FRONTEND_URL we have is somewhat different and only applies to Alby Cloud. So that one can probably be ignored for now.

I made some changes in this PR - could you review?

https://github.com/8de2fdb0/hub/pull/1

frontend/platform_specific/http/src/utils/request.ts
rolznz: I simplified this in https://github.com/8de2fdb0/hub/pull/1 and it works from my testing with caddy. Please let me know if that works for you.

One thing to note is you will need to update to use `/hub` rather than `/hub/` for the `BASE_PATH`.

frontend/src/App.tsx
rolznz: Thanks. After thinking about it again I don't think it's an issue - and I added extra documentation in my PR. I'll resolve this thread.

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/1439

Issue #1400, opened 2025-06-10
Bug: Incorrectly showing conflicting public/private channels warning when user has 1 public channel


rolznz:
![Image](https://github.com/user-attachments/assets/d68b8598-c734-4f39-b733-67fbff0d7bb6)

Advanced options are not shown so it should _know_ to check public.

PR #1401, opened 2025-06-10
fix: default to public channel when opening channel if all existing channels are public

rolznz:
Fixes https://github.com/getAlby/hub/issues/1400


im-adithya (approved):
tACK

PR #1403, opened 2025-06-11
Feat: swap out improvements

rolznz:
- add lockup transaction id to hold payment metadata
- return swap ID and payment hash in swap out response
- retries to get fee rates
- separate swap out success page with status

### DIFF
--- a/frontend/src/screens/wallet/swap/SwapOutStatus.tsx
+++ b/frontend/src/screens/wallet/swap/SwapOutStatus.tsx
@@ -1,9 +1,14 @@
-import { CircleCheckIcon, CopyIcon, ExternalLinkIcon } from "lucide-react";
+import {
+  CircleCheckIcon,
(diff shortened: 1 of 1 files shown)


frontend/src/screens/wallet/swap/SwapOutSuccess.tsx
```diff
@@ -0,0 +1,102 @@
...
+  CardHeader,
+  CardTitle,
+} from "src/components/ui/card";
+import { useToast } from "src/components/ui/use-toast";
+import { useSyncWallet } from "src/hooks/useSyncWallet";
+import { useTransaction } from "src/hooks/useTransaction";
+import { copyToClipboard } from "src/lib/clipboard";
+import { SwapOutResponse } from "src/types";
+
+export default function SwapOutSuccess() {
```
rolznz: I think this component should be renamed

PR #1408, opened 2025-06-11
fix: always pass in fee and handle mempool errors

im-adithya:
## Description

In LND, withdrawing with `TargetConf` set to 1 can make it use very high fees >20-30 sat per vB as it uses mean/median instead of mode (roughly). To avoid that, we set the fees in the frontend itself.

This also adds code to avoid the withdraw page from crashing if the mempool api returns an error, in that case, the user can still input a feeRate or else it will use the `TargetConf: 1`

## Screenshot

Screenshot 2025-06-11 at 6 25 09 PM

### DIFF
--- a/frontend/src/screens/wallet/WithdrawOnchainFunds.tsx
+++ b/frontend/src/screens/wallet/WithdrawOnchainFunds.tsx
@@ -3,7 +3,7 @@ import {
   ChevronDown,
   CopyIcon,
   ExternalLinkIcon,
-  TriangleAlertIcon,
+  InfoIcon,
 } from "lucide-react";
 import React from "react";
 import AppHeader from "src/components/AppHeader";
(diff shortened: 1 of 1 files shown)


frontend/src/screens/wallet/WithdrawOnchainFunds.tsx
```diff
@@ -315,7 +324,13 @@ export default function WithdrawOnchainFunds() {
               onOpenChange={setConfirmDialogOpen}
               open={confirmDialogOpen}
             >
-              <Button>Withdraw</Button>
+              <Button className="w-full">Withdraw</Button>
+              {feeRate && (
+                <div className="mt-2 text-muted-foreground text-sm flex items-center">
+                  <TriangleAlertIcon className="h-4 w-4 mr-1" />
+                  Onchain payment will be made with {feeRate} sat/vB fee
```
rolznz: I was wrong I think.

If we have a mempool error I am now thinking we should require the user to enter a fee rate. You were right, both LND and LDK are susceptible to very high fees if there's a fee spike, which the user likely doesn't want.

rolznz (approved):
tACK

PR #1419, opened 2025-06-17
feat: add swaps table

im-adithya:
# Swaps testing template

### Steps:
- Pay the invoice
- Boltz then locks up funds
- Wait for confirmation
- Claim the payment
- Using this claim's preimage, boltz will claim the htlc

### Edge cases to test:

- [ ] Invoice is expired/invalid
- [ ] Hub is shut down after initiation of swap
- [ ] Hub is shut down after boltz lockup is found in mempool
- [ ] Hub is shut down after boltz lockup confirms
- [ ] Hub is shut down while lockup is found
- [ ] Hub is shut down while lockup is confirmed

### Refunds

Not Applicable as HTLC is cancelled in worst case. (Can't be claimed by boltz till we claim it onchain by revealing the preimage)

### Steps:
- Give boltz an invoice and receive an address + amount
- Lockup amount in given address
- Wait for confirmation
- Boltz will automatically pay the invoice once confirmed

### Edge cases to test:

- [ ] Invoice is expired/invalid
- [ ] Hub is shut down after initiation of swap
- [ ] Hub is shut down after lockup deposit
- [ ] Hub is shut down while lockup is deposited

### Refunds

- [ ] Rescue key
- [ ] Process refund button in debug tools
- [ ] Wrong amount locked up
- [ ] Refund before swap is expired

## Auto Swap Out:
- [ ] Runs in given frequency

# PR TODOs
- [x] Test wails
- [x] Test auto swaps once again
- [x] Test BIP85 mnemonic derivation
- [x] Add tests for BIP85 mnemonic derivation
- [ ] Test all failure cases (expired invoice etc)
- [ ] Test trying to refund a non-expired invoice
- [x] Test rescue from mnemonic - works
- [x] Check failed swap outs flow
- [x] Process Refund button + api (debug tool)
- [x] List all swaps in frontend (debug tool)
- [x] Reconnect to boltz for updates
- [x] Specify fees while locking up btc for swap ins
- [x] Start swap service similar to how LNClient is done (to solve context.Background issue - this causes swaps to stay in pending state instead of failed)
- [x] Mark swap as settled if corresponding invoice settles

### DIFF
--- a/service/keys/keys_test.go
+++ b/service/keys/keys_test.go
@@ -138,6 +138,7 @@ func TestGenerateSwapMnemonic(t *testing.T) {
 	swapMnemonic, err := keys.GetSwapMnemonic(masterKey)
 	require.NoError(t, err)
 
+	// this matches https://iancoleman.io/bip39/ -> check "Show BIP85" and set BIP85 Index to 128260
 	expectedSwapMnemonic := "truth cargo pluck prefer mosquito symptom review kitchen exile fit corn vault"
 	assert.Equal(t, expectedSwapMnemonic, swapMnemonic)
 }


service/keys/keys.go
```diff
@@ -93,6 +98,19 @@ func (keys *keys) Init(cfg config.Config, encryptionKey string) error {
 	}
 	keys.appKey = appKey
 
+	netParams := &chaincfg.MainNetParams
+	network := cfg.GetNetwork()
+	if network == "testnet" {
+		netParams = &chaincfg.TestNet3Params
+	}
+
+	hdMasterKey, err := hdkeychain.NewMaster(bip39.NewSeed(mnemonic, ""), netParams)
```
rolznz: We need to use BIP-85 here to generate a child mnemonic with a custom derivation path so that it's isolated from the recovery phrase

im-adithya: Yeah I couldn't find a go package to do that easily, will add our own implementation and test it against any bip-85 js package :+1:

rolznz: I didn't test it, but: https://github.com/e4coder/bip85

rolznz: Edit: we should probably not include that library, it has some weird dependencies. But we could try the code

im-adithya: Done ✅ Also tested against test vectors in BIP-85 spec

frontend/src/screens/wallet/swap/SwapStatus.tsx
```diff
@@ -0,0 +1,545 @@
...
+import { useToast } from "src/components/ui/use-toast";
+import { useBalances } from "src/hooks/useBalances";
+import { useMempoolApi } from "src/hooks/useMempoolApi";
+import { useSwap } from "src/hooks/useSwaps";
+import { useSyncWallet } from "src/hooks/useSyncWallet";
+import { copyToClipboard } from "src/lib/clipboard";
+import { RedeemOnchainFundsResponse, SwapIn, SwapOut } from "src/types";
+import { request } from "src/utils/request";
+
+export default function SwapStatus() {
```
rolznz: Why do you put both in the same page? now there is a conditional and 99% of the code is in different components. Can you split it again?

frontend/src/screens/wallet/swap/SwapStatus.tsx
```diff
@@ -0,0 +1,545 @@
...
+    <div className="w-full max-w-lg">
+      <Card className="w-full md:max-w-xs">
+        <CardHeader>
+          <CardTitle className="flex justify-center">
+            {swapStatus === "PENDING" && <Loading className="w-4 h-4 mr-2" />}
+            {statusText[swapStatus]}
+          </CardTitle>
+        </CardHeader>
+        <CardContent className="flex flex-col items-center gap-4">
+          {swapStatus === "SUCCESS" && (
```
rolznz: @im-adithya I think the swap out should only be fully successful once there is a confirmation. The transaction is only broadcasted, not confirmed.

image

frontend/src/screens/wallet/swap/SwapStatus.tsx
```diff
@@ -0,0 +1,545 @@
...
+                    <CircleCheckIcon className="w-5 h-5 mr-2 text-green-600 dark:text-emerald-500" />
+                    Funds received via lightning
+                  </div>
+                  <Divider color="border-green-600 dark:border-emerald-500" />
+                  <div className="flex items-center text-muted-foreground text-sm">
+                    <CircleCheckIcon className="w-5 h-5 mr-2 text-green-600 dark:text-emerald-500" />
+                    <div className="flex items-center gap-2">
+                      <p>Lockup confirmed.</p>
+                      <ExternalLink
+                        to={`https://mempool.space/tx/${swap.lockupTxId}`}
```
rolznz: we need to update the hardcoded links

frontend/src/screens/wallet/swap/SwapStatus.tsx
rolznz: Also I think we are way too technical and do not need all these steps / states. Who apart from you will understand it as it is now?

frontend/src/screens/wallet/swap/SwapStatus.tsx
```diff
@@ -0,0 +1,545 @@
...
+                              View
+                            </ExternalLink>
+                            <CircleHelpIcon className="h-4 w-4 text-muted-foreground" />
+                          </div>
+                        </TooltipTrigger>
+                        <TooltipContent className="w-[300px]">
+                          Lockup usually fails when there is an amount mismatch
+                          or if Hub was shut down during the swap process.
+                          {!swap.claimTxId &&
+                            " You can use the Process Refund button to claim the locked up bitcoin."}
```
rolznz: This is not helpful, where is the button, how does the user navigate to it and do a refund? we have a guide I made already.

swaps/swaps_service.go
```diff
@@ -488,228 +545,225 @@ func (svc *swapsService) SwapIn(ctx context.Context, amount uint64, lnClient lnc
...
+	}
+
+	var tree *boltz.SwapTree
+	var ourKeys *btcec.PrivateKey
+	var swap *boltz.CreateSwapResponse
+
+	err = svc.db.Transaction(func(tx *gorm.DB) error {
+		err := tx.Save(&dbSwap).Error
+		if err != nil {
+			return err
```
rolznz, line 380: anywhere we return errors here we are not updating the swap state - so it's still pending

frontend/src/screens/settings/DebugTools.tsx
```diff
@@ -134,6 +135,62 @@ function ProbeKeysendDialogContent({ apiRequest }: Props) {
...
+        description: "" + error,
+      });
+    }
+    setSwapId("");
+  }
+
+  return (
+    <AlertDialogContent>
+      <AlertDialogHeader>
+        <AlertDialogTitle className="capitalize">Swap Refund</AlertDialogTitle>
```
rolznz:
```suggestion
        <AlertDialogTitle className="capitalize">Refund Swap</AlertDialogTitle>
```

db/migrations/202506170342_swaps.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	_ "embed"
+	"text/template"
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+const swapsMigration = `
+CREATE TABLE swaps(
```
rolznz, line 12: Some weird things here, this is the state of a swap after a successful refund: (I sent the wrong amount of 1000 sats to test it)

- The state is "FAILED", it should be "REFUNDED"
- amountSent is named incorrectly - I did not send that amount. Can we rename it to `sendAmount`?
- it's kind of weird `amountReceived` was set to the refund amount. How did this get set?
- destination is empty for swapping in - maybe then it could be renamed e.g. destinationAddress
- for refunds, maybe we can add a `refundAddress` and set it


```
{
      "id": "vAaAMKEAzozP",
      "type": "in",
      "state": "FAILED",
      "address": "bcrt1pzvqyrsdmnepukr9vm7xkjp625tctj2cd5khtz3ct03l0k8ksn77qyc0zhk",
      "amountSent": 50852,
      "amountReceived": 901,
      "paymentHash": "6bb297c0833dc46c013a3b3326fef7d86d5d886d00a19a91bec8b8faaed45747",
      "destination": "",
      "lockupTxId": "b1e28128adeaef3a048e995cb6a2b5cf43c9bc3fd09b3611205720d7f31da3e4",
      "claimTxId": "bdc6cfcdf25d0940685f566e50d62e8aa5e3944eab9cfe89f67a270664fc4536",
      "autoSwap": false,
      "boltzPubkey": "02d9fa65e27683dd2f1924e93508e5e99133e268c578733263eb7a891f6778a43a",
      "createdAt": "2025-06-25T15:18:14+07:00",
      "updatedAt": "2025-06-25T15:18:33+07:00"
    }
```

swaps/swaps_service.go
im-adithya: Fixed ✅

service/keys/keys.go
```diff
@@ -132,3 +165,34 @@ func (keys *keys) DeriveKey(path []uint32) (*bip32.Key, error) {
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return key.ECPrivKey()
+}
+
+// Taken from https://github.com/e4coder/bip85/blob/main/bip85.go
+func (keys *keys) GetSwapMnemonic(key *bip32.Key) (string, error) {
```
rolznz, line 178: Can we you a test for this with a couple of the vectors?

frontend/src/screens/wallet/swap/SwapStatus.tsx
im-adithya: Done ✅

frontend/src/screens/wallet/swap/SwapStatus.tsx
rolznz: I think we should remove the button and only do automatic refunds or refunds through debug tools, otherwise there are 3 different places

service/keys/keys.go
```diff
@@ -88,11 +95,37 @@ func (keys *keys) Init(cfg config.Config, encryptionKey string) error {
 	albyHubIndex := uint32(bip32.FirstHardenedChild + 128029 /* 🐝 */)
 	appKey, err := masterKey.NewChildKey(albyHubIndex)
 	if err != nil {
-		logger.Logger.WithError(err).Error("Failed to create seed from mnemonic")
+		logger.Logger.WithError(err).Error("Failed to derive app key")
 		return err
 	}
 	keys.appKey = appKey
 
+	swapIndex := uint32(bip32.FirstHardenedChild + 128260 /* 🔄 */)
```
rolznz: I wrote down my key into https://iancoleman.io/bip39/ and then entered `128260` as the BIP85 Index. It doesn't return the same mnemonic as what is generated. Do you know why?

service/keys/keys.go
```diff
@@ -88,11 +95,37 @@ func (keys *keys) Init(cfg config.Config, encryptionKey string) error {
...
+	swapMasterKey, err := masterKey.NewChildKey(swapIndex)
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to derive swap master key")
+		return err
+	}
+
+	mnemonic, err = keys.GetSwapMnemonic(swapMasterKey)
```
rolznz: please do not use the same variable name for a different thing. Can you rename this to `swapMnemonic`?

db/migrations/202506170342_swaps.go
rolznz: Another one, when I deposited the wrong amount and disabled auto-refund:

The lockupTxId should not be empty, should it? I deposited funds.
```
{
      "id": "vsp6dsevgbUb",
      "type": "in",
      "state": "FAILED",
      "address": "bcrt1pgewnv98uhz9wrad27vdttg6qm9u8rys9ltgm5j8pdk35gzjrsccqjdqrrx",
      "amountSent": 50852,
      "amountReceived": 50000,
      "paymentHash": "5f6fb10c84b0682723af6225b83a49858842e93fae20cb37e5018dcefe15af2b",
      "destination": "",
      "lockupTxId": "",
      "claimTxId": "",
      "autoSwap": false,
      "boltzPubkey": "0316c9a4ebe84573a3a75802f090ddbe2bd9a4a5088503e1fffc83363139ea371e",
      "createdAt": "2025-06-25T16:41:05+07:00",
      "updatedAt": "2025-06-25T16:41:18+07:00"
    }
```

service/keys/keys.go
im-adithya: Fixed now

frontend/src/screens/wallet/swap/SwapInStatus.tsx
```diff
@@ -85,130 +126,229 @@ export default function SwapInStatus() {
...
-                  !isPaying &&
-                  !hasPaidWithHubFunds && (
+                <CircleXIcon className="w-60 h-60" />
+                <div className="flex flex-col gap-2 items-center">
+                  <p className="text-xl font-bold slashed-zero text-center">
+                    {new Intl.NumberFormat().format(swap.amountSent)} sats
+                  </p>
+                  <FormattedFiatAmount amount={swap.amountSent} />
+                </div>
+                {swap.lockupTxId && !swap.claimTxId && (
```
rolznz: I think we should remove this button, users can go to debug tools to recover a swap

frontend/src/types.ts
```diff
@@ -207,14 +209,38 @@ export type SwapFees = {
...
+  destination?: never;
+  lockupTxId?: string;
+  claimTxId?: string;
+};
+
+export type SwapOut = BaseSwap & {
+  type: "out";
+  destination: string;
+  address?: never;
+  amountReceived?: number;
```
rolznz: some fields here are duplicated and can be moved to BaseSwap

transactions/transactions_service.go
```diff
@@ -1460,3 +1478,31 @@ func (svc *transactionsService) markPaymentFailed(tx *gorm.DB, dbTransaction *db
 	})
 	return nil
 }
+
+// NOTE: we wouldn't need this if we resubscribe to pending swaps
```
rolznz: can we do this instead? it feels wrong to do this from the transactions service. I don't think it should know about swaps at all

swaps/swaps_service.go
```diff
@@ -805,3 +925,195 @@ func (svc *swapsService) getFeeRates() (*FeeRates, error) {
 	}
 	return &rates, nil
 }
+
+func (svc *swapsService) markSwapState(dbSwap *db.Swap, state string) {
+	dbErr := svc.db.Model(&dbSwap).Update("state", state).Error
```
rolznz: here you get a `**db.Swap` - I don't think that's intended?

db/migrations/202506170342_swaps.go
im-adithya: Is it possible that you turned Hub off just after you deposited the amount? If yes it might be possible that the boltz `boltz.TransactionMempool` update wasn't sent yet but the frontend showed the transaction because we set the lockup tx id from the deposit response [here](https://github.com/getAlby/hub/blob/feat/swaps-table/frontend/src/screens/wallet/swap/SwapInStatus.tsx#L117)

im-adithya: > it's kind of weird amountReceived was set to the refund amount. How did this get set?

We do it [here](https://github.com/getAlby/hub/blob/feat/swaps-table/swaps/swaps_service.go#L1092), should we just leave it unset?

im-adithya: > for refunds, maybe we can add a refundAddress and set it

Where can the user set it? I think it wouldn't look good if it's asked before the swap initiation. I think it's best to refund it to the Hub wallet for now, wdyt? (for auto refunds, for manual ones it should be easy to add a field in the debug dialog)

rolznz: > Where can the user set it? I think it wouldn't look good if it's asked before the swap initiation. I think it's best to refund it to the Hub wallet for now, wdyt? (for auto refunds, for manual ones it should be easy to add a field in the debug dialog)

I mean, you need an address to use to receive the refund - so we should save that in the swap table.

> Is it possible that you turned Hub off just after you deposited the amount? If yes it might be possible that the boltz boltz.TransactionMempool update wasn't sent yet but the frontend showed the transaction because we set the lockup tx id from the deposit response [here](https://github.com/getAlby/hub/blob/feat/swaps-table/frontend/src/screens/wallet/swap/SwapInStatus.tsx#L117)

I don't think so - I run a local mempool, it should be there instantly even if I didn't mine a block.

> We do it [here](https://github.com/getAlby/hub/blob/feat/swaps-table/swaps/swaps_service.go#L1092), should we just leave it unset?

Hmm, as long as we have a REFUNDED state maybe it makes sense as-is

swaps/swaps_service.go
```diff
@@ -1077,9 +1087,10 @@ func (svc *swapsService) RefundSwap(swapId string) error {
 		"claimTxId": claimTxId,
 	}).Info("Claim transaction broadcasted for refund")
 
-	err = svc.db.Model(&swap).Updates(map[string]interface{}{
-		"claim_tx_id":     claimTxId,
-		"amount_received": refundAmount,
+	err = svc.db.Model(&swap).Updates(&db.Swap{
+		ClaimTxId:      claimTxId,
+		AmountReceived: refundAmount,
+		State:          constants.SWAP_STATE_REFUNDED,
```
rolznz, line 694: Why do we have two different places where we set the swap state to refunded?

service/keys/keys_test.go
```diff
@@ -109,3 +109,35 @@ func TestGenerateNewMnemonic(t *testing.T) {
...
+	require.Equal(t, mnemonic, mnemonicFromConfig)
+
+	seed := bip39.NewSeed(mnemonicFromConfig, "")
+	masterKey, err := bip32.NewMasterKey(seed)
+	require.NoError(t, err)
+
+	swapMnemonic, err := keys.GetSwapMnemonic(masterKey)
+	require.NoError(t, err)
+
+	expectedSwapMnemonic := "truth cargo pluck prefer mosquito symptom review kitchen exile fit corn vault"
```
rolznz, line 142: We could add a comment: this matches https://iancoleman.io/bip39/ -> check "Show BIP85" and set BIP85 Index to 128260

swaps/swaps_service.go
```diff
@@ -819,6 +819,8 @@ func (svc *swapsService) SwapIn(amount uint64, autoSwap bool) (*SwapResponse, er
 							logger.Logger.WithError(err).WithFields(logrus.Fields{
 								"swapId": swap.Id,
 							}).Error("Could not process refund")
+						} else {
+							swapState = constants.SWAP_STATE_REFUNDED
```
rolznz: This shouldn't be needed actually right? since it's done inside the `RefundSwap` method

api/api.go
```diff
@@ -603,20 +603,20 @@ func (api *api) ListSwaps() (*ListSwapsResponse, error) {
...
-		ClaimTxId:      swap.ClaimTxId,
-		AutoSwap:       swap.AutoSwap,
-		BoltzPubkey:    swap.BoltzPubkey,
-		CreatedAt:      swap.CreatedAt.Format(time.RFC3339),
-		UpdatedAt:      swap.UpdatedAt.Format(time.RFC3339),
+		Id:                 swap.SwapId,
+		Type:               swap.Type,
+		State:              swap.State,
+		SendAmount:         swap.SendAmount,
+		ReceivedAmount:     swap.ReceivedAmount,
```
rolznz, line 611: Shouldn't this be "ReceiveAmount" not "ReceivedAmount"?

swaps/swaps_service.go
```diff
@@ -1036,7 +1045,7 @@ func (svc *swapsService) RefundSwap(swapId string) error {
 		logger.Logger.WithError(err).Error("Failed to fetch fee rate to create claim transaction")
 		return err
 	}
-	// TODO: generate a new key
+	// TODO: Use refundAddress
```
rolznz: You should _set_ the refund address here on the swap item in the DB. It still needs a new generated address

swaps/swaps_service.go
```diff
@@ -1021,12 +1021,15 @@ func (svc *swapsService) RefundSwap(swapId string) error {
 		logger.Logger.WithError(err).Error("Failed to fetch fee rate to create claim transaction")
 		return err
 	}
-	// TODO: Use refundAddress
-	address, err := svc.cfg.Get(config.OnchainAddressKey, "")
-	if err != nil {
-		logger.Logger.WithError(err).Error("Failed to get on-chain address from config")
-		return err
+
+	if address == "" {
```
rolznz, line 627: We need to save the refund address to the swap

swaps/swaps_service.go
im-adithya: This is done because of some hacky code, that is removed after adding resubscribing logic, should be good now!

swaps/swaps_service.go
im-adithya: Addressed in other comment (it is only done here now)

api/api.go
im-adithya: Ah I was wondering why this sounds wrong 😄 Changing!

swaps/swaps_service.go
```diff
@@ -1115,6 +763,448 @@ func (svc *swapsService) ListSwaps() ([]Swap, error) {
...
+	if len(swaps) == 0 {
+		return
+	}
+
+	logger.Logger.WithField("count", len(swaps)).Info("Resuming pending swaps...")
+
+	boltzWs := svc.boltzApi.NewWebsocket()
+	if err := boltzWs.Connect(); err != nil {
+		logger.Logger.WithError(err).Error("Could not connect to Boltz websocket")
+		return
```
rolznz, line 748: do we really want a return here? then the user is stuck if they have pending swaps

im-adithya: Fixed

PR #1426, opened 2025-06-21
feat: allow self-paying invoices generated externally from LND

### LINKED ISSUE #1294: Allow self-paying invoices generated externally from LND
Currently it falls into the self-payment code which checks for an existing transaction and fails:

`"error":"The transaction requested was not found","level":"error","msg":"Failed to send`

This seems to be because self-payment check code is doing:

`selfPayment := paymentRequest.Payee != "" && paymentRequest.Payee == lnClient.GetPubkey()`

We could probably change this to check for the invoice exists in our db instead.

ajaysehwal:
Fixes #1294 

Currently, when users try to pay invoices generated externally from LND (using lncli addinvoice or other wallets), 
Alby Hub incorrectly identifies them as self-payments and fails with the "The transaction requested was not found" error. 
This happens because the self-payment detection logic only checks if the invoice's payee pubkey matches the node's pubkey, without verifying if the invoice actually exists in our database. 
The fix enhances this detection to require both conditions: the payee pubkey must match our node's pubkey, AND a corresponding incoming transaction must exist in our database.
This ensures external invoices with matching pubkey process as regular payments while maintaining the existing self-payment functionality for true self-payments. All existing tests continue to pass, confirming no regression in current behavior.

### DIFF
--- a/transactions/self_payment_detection_test.go
+++ b/transactions/self_payment_detection_test.go
@@ -1,7 +1,6 @@
 package transactions
 
 import (
-	"context"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -13,8 +12,6 @@ import (
 )
 
 func TestSendPaymentSync_SelfPaymentDetection_WithIncomingTransaction(t *testing.T) {
-	ctx := context.TODO()
-
 	svc, err := tests.CreateTestService(t)
 	require.NoError(t, err)
 	defer svc.Remove()
@@ -32,7 +29,7 @@ func TestSendPaymentSync_SelfPaymentDetection_WithIncomingTransaction(t *testing
 	})
 
 	transactionsService := NewTransactionsService(svc.DB, svc.EventPublisher)
-	transaction, err := transactionsService.SendPaymentSync(ctx, tests.MockInvoice, nil, nil, svc.LNClient, nil, nil, nil)
+	transaction, err := transactionsService.SendPaymentSync(tests.MockInvoice, nil, nil, svc.LNClient, nil, nil)
 
 	assert.NoError(t, err)
 	assert.NotNil(t, transaction)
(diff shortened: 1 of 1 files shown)


ajaysehwal:
Hi @frnandu and @im-adithya, could you please review this PR and let me know if any further changes are needed?

rolznz:
@ajaysehwal any chance you could add a couple of tests to cover the new logic?

transactions/self_payment_detection_test.go
```diff
@@ -0,0 +1,214 @@
...
+	transactionsService := NewTransactionsService(svc.DB, svc.EventPublisher)
+	transaction, err := transactionsService.SendPaymentSync(ctx, tests.MockInvoice, nil, nil, svc.LNClient, nil, nil, nil)
+
+	assert.NoError(t, err)
+	assert.NotNil(t, transaction)
+	assert.False(t, transaction.SelfPayment)
+	assert.Equal(t, constants.TRANSACTION_STATE_SETTLED, transaction.State)
+}
+
+func TestSendPaymentSync_SelfPaymentDetection_MultipleIncomingTransactions(t *testing.T) {
```
rolznz: What is the purpose of this test? can we remove it?

transactions/self_payment_detection_test.go
```diff
@@ -0,0 +1,214 @@
...
+	transactionsService := NewTransactionsService(svc.DB, svc.EventPublisher)
+	transaction, err := transactionsService.SendPaymentSync(ctx, tests.MockInvoice, nil, nil, svc.LNClient, nil, nil, nil)
+
+	assert.NoError(t, err)
+	assert.NotNil(t, transaction)
+	assert.False(t, transaction.SelfPayment)
+	assert.Equal(t, constants.TRANSACTION_STATE_SETTLED, transaction.State)
+}
+
+func TestSendPaymentSync_SelfPaymentDetection_DatabaseError(t *testing.T) {
```
rolznz: What is the purpose of this test? can we remove it?

transactions/self_payment_detection_test.go
```diff
@@ -0,0 +1,214 @@
...
+	sqlDB.Close()
+
+	transactionsService := NewTransactionsService(svc.DB, svc.EventPublisher)
+	transaction, err := transactionsService.SendPaymentSync(ctx, tests.MockInvoice, nil, nil, svc.LNClient, nil, nil, nil)
+
+	assert.Error(t, err)
+	assert.Nil(t, transaction)
+}
+
+func TestSendPaymentSync_SelfPaymentDetection_WithApp(t *testing.T) {
```
rolznz: What is the purpose of this test? can we remove it? (does it matter if there's an app or not? the code path is the same?)

PR #1430, opened 2025-06-23
feat: specify max channel saturation power of half from env

### LINKED ISSUE #1425: LDK: allow configuring max_channel_saturation_power_of_half via env
> given channels on the network are often in the few-million sat range, sending 3+M sats in a single payment is not all that commonly doable
that said, ldk does try to be a bit conservative about what % of a channel it will try to use in its MPP split heuristic. Given your 2M sat payments worked across 5-10 MPP parts, I'd say its doing pretty well, but you could try playing with https://docs.rs/lightning/latest/lightning/routing/router/struct.PaymentParameters.html#structfield.max_channel_saturation_power_of_half somewhat. It defaults to never sending more than 1/4 of a channel at a time, but you could try 1/8 or 1/2

im-adithya:
Fixes #1425

Adds an env var `LDK_MAX_CHANNEL_SATURATION` which is defaulted to `2` (same as LDK)

### DIFF
--- a/config/models.go
+++ b/config/models.go
@@ -15,40 +15,41 @@ const (
 )
 
 type AppConfig struct {
-	Relay                    string `envconfig:"RELAY" default:"wss://relay.getalby.com/v1"`
-	LNBackendType            string `envconfig:"LN_BACKEND_TYPE"`
-	LNDAddress               string `envconfig:"LND_ADDRESS"`
-	LNDCertFile              string `envconfig:"LND_CERT_FILE"`
-	LNDMacaroonFile          string `envconfig:"LND_MACAROON_FILE"`
-	Workdir                  string `envconfig:"WORK_DIR"`
-	Port                     string `envconfig:"PORT" default:"8080"`
-	DatabaseUri              string `envconfig:"DATABASE_URI" default:"nwc.db"`
-	JWTSecret                string `envconfig:"JWT_SECRET"`
-	LogLevel                 string `envconfig:"LOG_LEVEL" default:"4"`
-	LogToFile                bool   `envconfig:"LOG_TO_FILE" default:"true"`
-	Network                  string `envconfig:"NETWORK"`
-	LDKNetwork               string `envconfig:"LDK_NETWORK"`
-	LDKEsploraServer         string `envconfig:"LDK_ESPLORA_SERVER" default:"https://electrs.getalbypro.com"` // TODO: remove LDK prefix
-	LDKGossipSource          string `envconfig:"LDK_GOSSIP_SOURCE"`
-	LDKLogLevel              string `envconfig:"LDK_LOG_LEVEL" default:"3"`
-	LDKVssUrl                string `envconfig:"LDK_VSS_URL" default:"https://vss.getalbypro.com/vss"`
-	LDKListeningAddresses    string `envconfig:"LDK_LISTENING_ADDRESSES" default:"0.0.0.0:9735,[::]:9735"`
-	LDKTransientNetworkGraph bool   `envconfig:"LDK_TRANSIENT_NETWORK_GRAPH" default:"false"`
-	MempoolApi               string `envconfig:"MEMPOOL_API" default:"https://mempool.space/api"`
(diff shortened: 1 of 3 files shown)


config/models.go
```diff
@@ -15,40 +15,41 @@ const (
...
+	DatabaseUri                        string `envconfig:"DATABASE_URI" default:"nwc.db"`
+	JWTSecret                          string `envconfig:"JWT_SECRET"`
+	LogLevel                           string `envconfig:"LOG_LEVEL" default:"4"`
+	LogToFile                          bool   `envconfig:"LOG_TO_FILE" default:"true"`
+	Network                            string `envconfig:"NETWORK"`
+	LDKNetwork                         string `envconfig:"LDK_NETWORK"`
+	LDKEsploraServer                   string `envconfig:"LDK_ESPLORA_SERVER" default:"https://electrs.getalbypro.com"` // TODO: remove LDK prefix
+	LDKGossipSource                    string `envconfig:"LDK_GOSSIP_SOURCE"`
+	LDKLogLevel                        string `envconfig:"LDK_LOG_LEVEL" default:"3"`
+	LDKMaxChannelSaturationPowerOfHalf string `envconfig:"LDK_MAX_CHANNEL_SATURATION" default:"2"`
```
rolznz: can this be the proper type instead of a string?

im-adithya: Should I also change this for other fields? For example LDKLogLevel?

im-adithya: Fixed now ✅

config/models.go
```diff
@@ -15,40 +15,42 @@ const (
...
+	JWTSecret                          string `envconfig:"JWT_SECRET"`
+	LogLevel                           string `envconfig:"LOG_LEVEL" default:"4"`
+	LogToFile                          bool   `envconfig:"LOG_TO_FILE" default:"true"`
+	Network                            string `envconfig:"NETWORK"`
+	LDKNetwork                         string `envconfig:"LDK_NETWORK"`
+	LDKEsploraServer                   string `envconfig:"LDK_ESPLORA_SERVER" default:"https://electrs.getalbypro.com"` // TODO: remove LDK prefix
+	LDKGossipSource                    string `envconfig:"LDK_GOSSIP_SOURCE"`
+	LDKLogLevel                        string `envconfig:"LDK_LOG_LEVEL" default:"3"`
+	LDKMaxChannelSaturationPowerOfHalf uint8  `envconfig:"LDK_MAX_CHANNEL_SATURATION" default:"2"`
+	LDKMaxPathCount                    uint8  `envconfig:"LDK_MAX_PATH_COUNT" default:"10"`
```
rolznz:
```suggestion
	LDKMaxPathCount                    uint8  `envconfig:"LDK_MAX_PATH_COUNT" default:"5"`
```

lnclient/ldk/ldk.go
```diff
@@ -111,7 +111,10 @@ func NewLDKService(ctx context.Context, cfg config.Config, eventPublisher events
 	}
 
 	ldkConfig.ListeningAddresses = &listeningAddresses
-	logLevel, _ := strconv.Atoi(cfg.GetEnv().LDKLogLevel)
+	logLevel, err := strconv.Atoi(cfg.GetEnv().LDKLogLevel)
+	if err != nil {
+		logLevel = int(ldk_node.LogLevelDebug)
```
rolznz, line 117:
```suggestion
		logLevel = int(ldk_node.LogLevelInfo)
```

im-adithya: I did this because it is the same as the default (3) and in the next line we add `LogLevelGossip` (1) which brings it to `LogLevelInfo` (4)
```go
const (
	LogLevelGossip LogLevel = 1
	LogLevelTrace  LogLevel = 2
	LogLevelDebug  LogLevel = 3
	LogLevelInfo   LogLevel = 4
	LogLevelWarn   LogLevel = 5
	LogLevelError  LogLevel = 6
)
```

rolznz: can you add a comment for that?

im-adithya: Done

Issue #1412, opened 2025-06-16
Request: Allow us to input a custom mempool explorer URL


Ks2wAu89jnEgCR0bGUAA:
Many of us are running a self hosted instance of mempool.space in order to take advantage of the privacy and self sovereignty benefits. Clicking an on chain transaction in Alby Hub takes you to the public mempool.space instance. I would like to request a setting where we can paste the URL for our own instance of mempool so that can be used instead.

rolznz:
Hi, currently it can be set via an environment variable `MEMPOOL_API`. (Note: it's mentioned [here](https://github.com/getAlby/hub?tab=readme-ov-file#mutinynet) but it should be mentioned as a general environment variable)

Ks2wAu89jnEgCR0bGUAA:
I used the quick start script (https://github.com/getAlby/hub/tree/master/scripts/linux-x86_64) how can I configure the environment variables if installed in this manner?

Ks2wAu89jnEgCR0bGUAA:
Yes, I did try http too, the result was the same. Looking forward to the future versions <3

PR #1431, opened 2025-06-23
fix: use custom mempool url

im-adithya:
Fixes https://github.com/getAlby/hub/issues/1412

This allows users to set a custom mempool URL for the frontend. But I'm thinking if we can change the backend env var from `MEMPOOL_API` to `MEMPOOL_URL` and use it everywhere (not sure how to pass it to frontend though + it would also be a breaking change)

### DIFF
--- a/frontend/src/constants.ts
+++ b/frontend/src/constants.ts
@@ -19,3 +19,6 @@ export const SUPPORT_ALBY_CONNECTION_NAME = `ZapPlanner - Alby Hub`;
 export const SUPPORT_ALBY_LIGHTNING_ADDRESS = "hub@getalby.com";
 
 export const SUBWALLET_APPSTORE_APP_ID = "uncle-jim";
+
+export const MEMPOOL_URL =
+  import.meta.env.VITE_MEMPOOL_URL || "https://mempool.space";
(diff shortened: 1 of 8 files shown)


rolznz:
> This allows users to set a custom mempool URL for the frontend. But I'm thinking if we can change the backend env var from MEMPOOL_API to MEMPOOL_URL and use it everywhere (not sure how to pass it to frontend though + it would also be a breaking change)

We could return it in the info endpoint like we do with a few other environment variables

To avoid the breaking change we could just calculate the mempool URL from the mempool api (take the substring on the left hand side of `/api`) - this should work in all environments I think (testnet, mutinynet, mainnet, should be the same with self-hosted mempool instances?)

Then we don't need to add any new variables. What do you think?

frontend/src/constants.ts
```diff
@@ -19,3 +19,5 @@ export const SUPPORT_ALBY_CONNECTION_NAME = `ZapPlanner - Alby Hub`;
 export const SUPPORT_ALBY_LIGHTNING_ADDRESS = "hub@getalby.com";
 
 export const SUBWALLET_APPSTORE_APP_ID = "uncle-jim";
+
+export const DEFAULT_MEMPOOL_URL = "https://mempool.space";
```
rolznz: Can this be removed? we always have a mempool URL set in the backend

im-adithya: Done, ~but should we add a check in the backend to throw an error if required env variables are not set? I don't think we do this atm~ MempoolApi has a default so it should be good :+1:

PR #1438, opened 2025-06-24
docs: use local update script for pi hubs

rolznz:
We don't want to recommend downloading the update script every time as it renders the verify step useless.

See https://github.com/getAlby/hub/pull/1220#issuecomment-2997162063

### DIFF
--- a/scripts/pi-aarch64/README.md
+++ b/scripts/pi-aarch64/README.md
@@ -6,15 +6,13 @@ You should have some basic Linux understanding to install and operate it.
 Have a look at our [installation guide](https://github.com/getAlby/hub/tree/master/scripts/pi-arm) for more details and inspiration.
 
 SSH into your Pi and run:
+
 ```shell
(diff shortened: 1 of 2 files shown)


rolznz:
Ah, right. An update script should be downloaded as part of the install script. I need to update that as well to match our other scripts.

For now you can download the update script manually from https://github.com/getAlby/hub/tree/master/scripts/ (open the pi folder according to what device you have, and then open the update.sh script, and click "download raw file", then save it in /opt/albyhub on your pi).

You will then need to make it executable (chmod +x ./update.sh)

Sakrecoer:
Nicely done @rolznz ! One thing, you are running `wget` without any flags, meaning if the script is already present in the system, it will not overwrite it, instead it will create `update.sh.1`

To have `wget` overwrite the existing file, use `-O`. See: https://stackoverflow.com/a/30418815

PR #1441, opened 2025-06-24
feat: add expired state to nip47 transactions

### LINKED ISSUE #1429: Add expired state to transactions response
See https://github.com/nostr-protocol/nips/pull/1933

im-adithya:
Fixes #1429 

Simpler implementation to just return this expired state in NWC, internally we don't change anything as it would complicate unnecessarily.

### DIFF
--- a/nip47/models/transactions.go
+++ b/nip47/models/transactions.go
@@ -3,6 +3,7 @@ package models
 import (
 	"encoding/json"
 	"strings"
+	"time"
 
 	"github.com/getAlby/hub/logger"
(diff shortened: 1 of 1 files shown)


rolznz:
@im-adithya it goes through the same mapping function doesn't it? why are more changes needed?

im-adithya:
Whoops sorry, yes I didn't test `lookup_invoice` so I marked it for myself, works perfect as it's the same `toNip47Transaction` method ✅

Issue #1436, opened 2025-06-23
Add a debug tool to dump LDK-node scorer data


rolznz:
Add a debug tool to dump/export local scorer data that can be analyzed if payments are repeatedly failing

achaljhawar:
Hi can I work on this issue?

PR #1442, opened 2025-06-24
Expose ExportPathfindingScores LDK method as custom command

achaljhawar:
fixes: #1436 
- A new custom command `export_pathfinding_scores` is defined and implemented in the LDKService.
- The command calls the underlying `node.ExportPathfindingScores()` method and returns the file path of the exported scores.


achaljhawar:
sorry for the oversight I added the button for the command

Issue #1437, opened 2025-06-24
Auto unlock for LND does not always work


rolznz:
If the LND sync takes longer than our timeout (currently 1 minute?) then Alby Hub will mark the node startup as failed and the user will have to re-enter their password. This should be increased.

Examples where this happens:
- Lower end devices that sync slower (e.g. raspberry pi)
- After a longer time shutdown (e.g. after a powercut)

PR #1444, opened 2025-06-24
chore: increase lnd wait time before retrying

im-adithya:
Fixes https://github.com/getAlby/hub/issues/1437 by increasing wait time to 30s (should it be 1 minute?) cc @rolznz

And also addresses the issue where hitting `Ctrl+C` doesn't stop the app immediately because the os signal goroutine is registered after service creation.


rolznz:
@im-adithya nice fix. Maybe we can increase the `maxRetries` instead e.g. to 60 (would be equal to retrying up to 10 minutes?)

im-adithya:
Sounds good, changed ✅

rolznz (approved):
utACK

Issue #774, opened 2024-11-04
Improve Deposit Page UX


rolznz:
On the /channels/onchain/deposit-bitcoin page there are some issues:
- The address display is formatted with spaces - if you manually select and copy it you will copy with spaces which is an invalid bitcoin address. (the visual spacing should be done with css instead)
- The page should trigger sync of the wallet using the useSync hook.
- The page should check the mempool for this address and show incoming / confirmed funds.
- Once the funds are confirmed, show a success state and a button to go to the node page

![Image](https://github.com/user-attachments/assets/bd0bda94-dcbe-4385-ab2f-a274b695c707)

PR #1446, opened 2025-06-24
feat: show deposit status in onchain deposit page

im-adithya:
Fixes https://github.com/getAlby/hub/issues/774

Few things:
- I added option to specify custom `refreshInterval` to useMempoolApi as it is the same hook for multiple endpoints
- The deposit page only records the first incoming transaction and shows it's success screen, any other pending transactions coming in at the same time will be missed

## Screenshots
Screenshot 2025-06-24 at 6 33 39 PM
Screenshot 2025-07-02 at 6 04 12 PM
Screenshot 2025-06-24 at 6 38 10 PM


frontend/src/screens/onchain/DepositBitcoin.tsx
```diff
@@ -43,50 +86,82 @@ export default function DepositBitcoin() {
...
-              href={`bitcoin:${onchainAddress}`}
-              target="_blank"
-              className="flex justify-center"
-            >
-              <QRCode value={onchainAddress} />
-            </a>
+        {confirmedAmount ? (
+          <DepositSuccess amount={confirmedAmount} />
+        ) : (
+          <Card>
```
rolznz, line 100: I don't think we should show the QR code when the transaction is in the mempool. Can we change it to show a pending status?

frontend/src/screens/onchain/DepositBitcoin.tsx
```diff
@@ -1,30 +1,73 @@
...
+      const utxo = mempoolAddressUtxos.find((utxo) => utxo.txid === txId);
+      if (utxo?.status.confirmed) {
+        setConfirmedAmount(utxo.value);
+      }
+    } else {
+      const unconfirmed = mempoolAddressUtxos.find(
+        (utxo) => !utxo.status.confirmed
+      );
+      if (unconfirmed) {
+        toast({
```
rolznz: Maybe if we have a pending state then a toast is not really needed

im-adithya: Fixed

frontend/src/types.ts
```diff
@@ -340,6 +340,19 @@ export type OnchainBalanceResponse = {
   pendingSweepBalancesDetails: PendingBalancesDetails[];
 };
 
+// from https://mempool.space/docs/api/rest#get-address-utxo
+export type MempoolUtxo = {
```
rolznz, line 344: There's a `MempoolTransaction` on line 366. Can we use the same type for both?

rolznz: Ah, it is actually a different response type

PR #1452, opened 2025-06-27
feat: create sub-wallet lightning address

rolznz:
Closes https://github.com/getAlby/hub/issues/1387

TODO:
- [x] Review UI / flow
- [x] Review metadata (saving address as lud16 on the app)
- [x] Ability to add to existing sub-wallets (on individual app page)
- [x] Wails handler
- [x] Way to remove lightning address
  - [x] Test once getalby.com endpoint is available
- [x] Improve duplicate lightning address error message

image

image

image

image

image

### DIFF
--- a/alby/alby_oauth_service.go
+++ b/alby/alby_oauth_service.go
@@ -288,8 +288,8 @@ func (svc *albyOAuthService) GetInfo(ctx context.Context) (*AlbyInfo, error) {
 
 	if res.StatusCode >= 300 {
 		logger.Logger.WithFields(logrus.Fields{
-			"body":       string(body),
-			"statusCode": res.StatusCode,
+			"body":        string(body),
+			"status_code": res.StatusCode,
 		}).Error("info endpoint returned non-success code")
 		return nil, fmt.Errorf("info endpoint returned non-success code: %s", string(body))
 	}
@@ -379,7 +379,7 @@ func (svc *albyOAuthService) GetVssAuthToken(ctx context.Context, nodeIdentifier
 	}
 
 	if vssResponse.Token == "" {
-		logger.Logger.WithField("vssResponse", vssResponse).WithError(err).Error("No token in API response")
+		logger.Logger.WithField("vss_response", vssResponse).WithError(err).Error("No token in API response")
 		return "", errors.New("no token in vss response")
 	}
 
@@ -529,8 +529,8 @@ func (svc *albyOAuthService) GetMe(ctx context.Context) (*AlbyMe, error) {
 
 	if res.StatusCode >= 300 {
 		logger.Logger.WithFields(logrus.Fields{
-			"body":       string(body),
-			"statusCode": res.StatusCode,
+			"body":        string(body),
+			"status_code": res.StatusCode,
 		}).Error("users endpoint returned non-success code")
 		return nil, fmt.Errorf("users endpoint returned non-success code: %s", string(body))
 	}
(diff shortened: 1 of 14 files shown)


alby/alby_oauth_service.go
```diff
@@ -386,6 +386,118 @@ func (svc *albyOAuthService) GetVssAuthToken(ctx context.Context, nodeIdentifier
 	return vssResponse.Token, nil
 }
 
+func (svc *albyOAuthService) CreateLightningAddress(ctx context.Context, address string, appId uint) error {
+	logger.Logger.WithFields(logrus.Fields{
+		"address": address,
+		"app_id":  appId,
```
im-adithya, line 392: We have to align this inconsistency in logs, we keep mixing up `snake_case` and `camelCase` (unrelated to this PR)

frontend/src/hooks/useDeleteSubwalletLightningAddress.ts
```diff
@@ -0,0 +1,38 @@
...
+    } catch (error) {
+      toast({
+        title: "Failed to delete lightning address",
+        description: (error as Error).message,
+        variant: "destructive",
+      });
+    }
+    setDeletingLightningAddress(false);
+  }
+  return { deleteSubwalletLightningAddress, deletingLightningAddress };
```
im-adithya:
```suggestion
  return { deleteLightningAddress, deletingLightningAddress };
```

(minor) and similarly in imports

alby/alby_oauth_service.go
```diff
@@ -386,6 +386,118 @@ func (svc *albyOAuthService) GetVssAuthToken(ctx context.Context, nodeIdentifier
...
+		lightningAddressErrors := &createLightningAddressErrors{}
+		err = json.Unmarshal(responseBody, lightningAddressErrors)
+		if err != nil {
+			logger.Logger.WithError(err).Error("failed to unmarshal errors response")
+			return err
+		}
+		if len(lightningAddressErrors.Address) == 0 {
+			return errors.New("unknown error occurred")
+		}
+		return errors.New(lightningAddressErrors.Address[0])
```
im-adithya: This looks wrong (the response struct), can we change this in rails app to `Reason`?

rolznz: I think this is the standard errors response from the getalby.com backend when there is a validation error. Is there a different endpoint that returns reason?

api/api.go
```diff
@@ -251,6 +251,79 @@ func (api *api) DeleteApp(userApp *db.App) error {
...
+	}
+
+	err = api.albyOAuthSvc.CreateLightningAddress(ctx, createLightningAddressRequest.Address, createLightningAddressRequest.AppId)
+
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to create lightning address for app")
+		return err
+	}
+
+	metadata["lud16"] = createLightningAddressRequest.Address + "@getalby.com"
```
bumi: I think there is also a `full_address` property

frontend/src/screens/apps/ShowApp.tsx
```diff
@@ -284,6 +293,60 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
...
+                                size="lg"
+                                variant="secondary"
+                                loading={creatingLightningAddress}
+                                onClick={() =>
+                                  createLightningAddress(
+                                    intendedLightningAddress
+                                  )
+                                }
+                              >
+                                Create
```
bumi: Do we have a check for a subscription here, too? like on the create screen?

alby/alby_oauth_service.go
rolznz: I've renamed the ones in this file

api/api.go
rolznz: Updated

frontend/src/hooks/useDeleteSubwalletLightningAddress.ts
rolznz: Thanks, fixed

frontend/src/screens/apps/ShowApp.tsx
rolznz: Added

reneaaron (changes requested):
image

As a self-hosting user I still seem to see the UI to add a lightning address, however it fails with an oAuth error.

rolznz:
@reneaaron I just fixed it, sorry about that!

Issue #1460, opened 2025-07-03
Sqlite -> Postgres migration fails with "violates foreign key constraint" for production hub when inserting app permissions


rolznz:
I have some app_permissions entries for an app that does not exist. They were created in `2024-09-01` but we should have [solved this issue](https://github.com/getAlby/hub/pull/274) earlier in `2024-07-22`.

PR #1462, opened 2025-07-04
Fix: sqlite connection parameters, add migrations to delete orphaned records and change postgres transaction amount to bigint

rolznz:
Fixes https://github.com/getAlby/hub/issues/1460

From @rdmitr 
> I have a hunch that multiple sqlite "connections" are opened implicitly, and the PRAGMA ends up being applied to the initial DB connection only. Thing is, Gorm uses the sql package from the Go's standard library, which does internal connection pooling. When it needs a new connection, it uses the driver to open it

I tested this and it turns out to be true. After the change, cascade deletes are working properly again.

I'm not sure why they were working before - maybe a change in gorm since we added the pragma statements.

Applies the SQLITE pragmas to the connection string instead.

Then the existing data needs to be cleaned up to remove orphans (that's the first migration)

I found another issue while migrating my own hub DB to postgres - I had a transaction with an amount_msat bigger than a 4 byte integer. Postgres unlike sqlite treats integers as 4 bytes. (second migration)

I also found some places we use the DB directly within a DB transaction, rather than the transaction itself. Those have been fixed.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -1,7 +1,6 @@
 package db
 
 import (
-	"errors"
 	"fmt"
 	"strings"
 
@@ -50,9 +49,6 @@ func NewDBWithConfig(cfg *Config) (*gorm.DB, error) {
 		}
 	} else {
 		sqliteURI := cfg.URI
-		if strings.Contains(sqliteURI, "?") {
-			return nil, errors.New("sqlite query parameters should not be passed as they will be configured by Alby Hub")
-		}
 
 		// apply pragma if we're not running the tests
 		if !strings.Contains(sqliteURI, "?mode=memory") {
--- a/db/db.go
+++ b/db/db.go
@@ -1,6 +1,7 @@
 package db
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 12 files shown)


frnandu (approved):
utACK

db/db.go
```diff
@@ -80,46 +90,16 @@ func newSqliteDB(sqliteConfig sqlite.Config, gormConfig *gorm.Config) (*gorm.DB,
...
-	if err != nil {
-		return nil, err
-	}
 
-	// moves temporary tables from disk into RAM, speeds up performance a lot
-	err = gormDB.Exec("PRAGMA temp_store = memory", nil).Error
-	if err != nil {
-		return nil, err
-	}
+	// NOTE: PRAGMA command may not apply to all db connections
```
rdmitr: Guess the only way to achieve this is to make a custom wrapper SQL driver on top of go-sqlite3 that would execute the PRAGMA statement on each connection 🤔

rolznz: Sounds good - I will make a separate issue for this one. Maybe if you have time, it would be great if you could look into it!

rdmitr: Will do! 👍

PR #1466, opened 2025-07-07
Add spending recommendations

MoritzKa:
Users can also increase their receiving capacity by spending funds.

Should we align both notifications? One of them also mentions swaps as a way to increase receiving capacity. We could mention that on [frontend/src/screens/wallet/receive/ReceiveInvoice.tsx](https://github.com/getAlby/hub/blob/66a1be6d162f6ffb22f801a10f4c023ae6a93059/frontend/src/screens/wallet/receive/ReceiveInvoice.tsx#L116) too.


rolznz:
Adding swaps in the other place sounds good to me. I'll do that before merging

rolznz (approved):
tACK

PR #1364, opened 2025-06-01
feat: rebalance into channel (WIP)

rolznz:
TODOs:
- [ ] improve frontend to provide good default amounts and do not allow receiving through a counterparty with more spending capacity than others
- [ ] update default LSP service URL to mainnet
- [ ] loading spinner on rebalance dialog
- [ ] improve dialog copy
- [ ] update guide content

image

image

PR #1484, opened 2025-07-10
chore: minor copy changes, remove unnecessary metadata field

rolznz:
Addresses feedback in https://github.com/getAlby/hub/pull/1364

### DIFF
--- a/api/rebalance.go
+++ b/api/rebalance.go
@@ -113,7 +113,7 @@ func (api *api) RebalanceChannel(ctx context.Context, rebalanceChannelRequest *R
 		return nil, err
 	}
 
-	if paymentRequest.MSatoshi > int64(float64(rebalanceChannelRequest.AmountSat)*float64(1000)*float64(1.03)+1 /*fees*/) {
+	if paymentRequest.MSatoshi > int64(float64(rebalanceChannelRequest.AmountSat)*float64(1000)*float64(1.003)+1 /*0.3% fees*/) {
 		return nil, errors.New("rebalance payment is more expensive than expected")
 	}
 
--- a/api/rebalance.go
+++ b/api/rebalance.go
@@ -22,7 +22,6 @@ func (api *api) RebalanceChannel(ctx context.Context, rebalanceChannelRequest *R
 
 	receiveMetadata := map[string]interface{}{
 		"receive_through": rebalanceChannelRequest.ReceiveThroughNodePubkey,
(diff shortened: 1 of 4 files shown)


api/rebalance.go
```diff
@@ -22,7 +22,6 @@ func (api *api) RebalanceChannel(ctx context.Context, rebalanceChannelRequest *R
 
 	receiveMetadata := map[string]interface{}{
 		"receive_through": rebalanceChannelRequest.ReceiveThroughNodePubkey,
-		"amount_sat":      rebalanceChannelRequest.AmountSat,
```
im-adithya, line 25: Similarly in `payMetadata ` below

frontend/src/components/channels/ChannelDropdownMenu.tsx
```diff
@@ -70,7 +70,7 @@ export function ChannelDropdownMenu({
                   onClick={() => setDialog("rebalance")}
                 >
                   <ScaleIcon className="h-4 w-4" />
-                  Rebalance In
+                  Rebalance
```
im-adithya, line 73: Just marking that we have to change this in the guides too

api/rebalance.go
rolznz: I think this one's different - this is how much we want to swap. The payment will be more expensive because it includes the service fee

im-adithya: Ah okay got it, thanks 👍

frontend/src/screens/wallet/swap/SwapOutStatus.tsx
```diff
@@ -130,7 +130,7 @@ export default function SwapOutStatus() {
                       <div className="flex items-center text-muted-foreground text-sm">
                         <Loading className="w-5 h-5 mr-2" />
                         <div className="flex items-center gap-2">
-                          <p>Waiting for onchain confirmation...</p>
+                          <p>Waiting for 2 onchain confirmations...</p>
```
im-adithya: Whoops, this should be 2/1 based on whether claimTxId is available I'll change that

rolznz: Thanks!

frontend/src/components/channels/ChannelDropdownMenu.tsx
rolznz: @im-adithya I'm undoing this change just here - now it's completely unclear what the direction of the rebalance is

PR #1495, opened 2025-07-12
feat: standalone lightning address support

matjaz:
For standalone HTTP instance add support for Lightning addresses.
Currently username is app_pubkey, but custom username would be better option.

Supported are following:
- LUD-06 - payRequest
- LUD-12 - comment
- LUD-16 - lightning address (static internet identifier)
- LUD-21 - verify


rolznz:
@matjaz thanks for the PR.

However, I am not sure LNURL fits well in the hub:
- It won't work for locally-hosted hubs that are not publicly accessible
- It won't work for Alby Cloud because requests require a proxy to go to the correct user's hub
- It won't work for desktop apps as they don't have an http interface

I think it seems better to keep it separate (such as the service you implemented, or Alby Lite). Also it's already been merged that you can now create getalby.com lightning addresses for sub-wallets.

matjaz:
Hey,

that's all correct, however it's a perfect fit for public self-hosted hub with custom domain, not requiring additional app. 
Separate app requires maintenance, duplicates data and communicates via relay.

rolznz:
@matjaz I asked the team. Although it is a nice idea, this will not work for the majority of users and we already have a solution that works no matter what way you run Alby Hub.

The getalby.com sub-wallet lightning addresses do require a relay, but they use the getalby.com relay which we spend a lot of time to make it reliable. Lightning addresses are a core feature of the Alby Account. We will ensure that they work reliably.

I will mark this as a draft for now, but please feel free to maintain a fork and if anyone who self-hosts on a VPS asks for this feature I will let them know you have implemented this.

PR #1497, opened 2025-07-13
fix: check reserved app name

matjaz:


rolznz (approved):
Nice improvement! utACK

PR #1498, opened 2025-07-14
feat: add app last used time

rolznz:
Now that we delete excess request and response events, they are not reliable to be used to show the last used time for the app.

Instead we store it as a timestamp column on the app itself.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -333,9 +333,6 @@ func (api *api) DeleteLightningAddress(ctx context.Context, appId uint) error {
 
 func (api *api) GetApp(dbApp *db.App) *App {
 
-	var lastEvent db.RequestEvent
-	lastEventResult := api.db.Where("app_id = ?", dbApp.ID).Order("id desc").Limit(1).Find(&lastEvent)
-
 	paySpecificPermission := db.AppPermission{}
 	appPermissions := []db.AppPermission{}
 	var expiresAt *time.Time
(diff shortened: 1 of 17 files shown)


api/api.go
```diff
@@ -461,8 +454,8 @@ func (api *api) ListApps() ([]App, error) {
 		var lastEvent db.RequestEvent
 		lastEventResult := api.db.Where("app_id = ?", dbApp.ID).Order("id desc").Limit(1).Find(&lastEvent)
 		if lastEventResult.RowsAffected > 0 {
-			apiApp.LastEventAt = &lastEvent.CreatedAt
 		}
+		apiApp.LastUsed = dbApp.LastUsed
```
im-adithya: All this can be removed now, instead we can add this directly at L437

api/api.go
```diff
@@ -389,18 +386,14 @@ func (api *api) GetApp(dbApp *db.App) *App {
 		Metadata:           metadata,
 		WalletPubkey:       walletPubkey,
 		UniqueWalletPubkey: uniqueWalletPubkey,
+		LastUsed:           dbApp.LastUsed,
```
im-adithya: Wondering if we should also call this `LastUsedAt` as we use the suffix for time related fields everywhere

api/api.go
rolznz: oops, thanks

api/api.go
rolznz: True, thanks!

PR #1503, opened 2025-07-15
fix: remove event deletion from migration

rolznz:
Deleting all at once is too slow for hubs that have many events. Instead we will use the background task to slowly remove excess events in batches.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -326,4 +326,15 @@ func (svc *service) removeExcessEvents() {
 		"amount":   numEventsToDelete,
 		"below_id": deleteEventsBelowId,
 	}).Info("Removed excess events")
+
+	// TODO: REMOVE AFTER 2025-01-01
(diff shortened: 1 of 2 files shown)


service/service.go
```diff
@@ -326,4 +326,15 @@ func (svc *service) removeExcessEvents() {
 		"amount":   numEventsToDelete,
 		"below_id": deleteEventsBelowId,
 	}).Info("Removed excess events")
+
+	// TODO: REMOVE AFTER 2025-01-01
```
im-adithya, line 330: 2026*, also this reminds me that we can now remove [version tag](https://github.com/getAlby/hub/blob/master/nip47/event_handler.go#L110) too. `GetFirst` also looks deprecated there, ~I'll create an issue~ opened https://github.com/getAlby/hub/issues/1505

Issue #764, opened 2024-10-29
rename LOG_EVENTS environment variable


rolznz:
This is exclusively used to publish events to the Alby API (e.g. to send notifications of payments or channel opens/closures).

PR #1509, opened 2025-07-17
chore: rename LOG_EVENTS env variable

matjaz:
Fixes https://github.com/getAlby/hub/issues/764

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
 # only enable event logging in production
-LOG_EVENTS_TO_ALBY=false
+SEND_EVENTS_TO_ALBY=false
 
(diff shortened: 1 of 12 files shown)


.env.example
```diff
@@ -1,5 +1,5 @@
 # only enable event logging in production
-LOG_EVENTS=false
+LOG_EVENTS_TO_ALBY=false
```
rolznz:
```suggestion
SEND_EVENTS_TO_ALBY=false
```

rolznz:
Thanks for this PR, and the cleanup too!

rolznz (approved):
utACK

matjaz:
maybe note, it's a braking change for someone set to false.

PR #1510, opened 2025-07-17
fetch peer alias on backend

matjaz:


lnclient/models.go
```diff
@@ -173,10 +174,12 @@ type OnchainBalanceResponse struct {
...
-	NodeId      string `json:"nodeId"`
-	Address     string `json:"address"`
-	IsPersisted bool   `json:"isPersisted"`
-	IsConnected bool   `json:"isConnected"`
+	NodeId           string  `json:"nodeId"`
+	NodeAlias        *string `json:"nodeAlias"`
+	Address          string  `json:"address"`
+	IsPersisted      bool    `json:"isPersisted"`
+	IsConnected      bool    `json:"isConnected"`
+	HasOpenedChannel bool    `json:"hasOpenedChannel"`
```
rolznz: I don't think we should return this here

frontend/src/screens/peers/Peers.tsx
```diff
@@ -21,51 +21,19 @@ import {
...
-  React.useEffect(() => {
-    loadNodeStats();
-  }, [loadNodeStats]);
-
   async function checkDisconnectPeer(peer: Peer) {
     try {
-      if (!channels) {
-        throw new Error("channels not loaded");
-      }
-      if (channels.some((channel) => channel.remotePubkey === peer.nodeId)) {
```
rolznz, line 68: in this case I think this simple check is better than the complexity introduced in the API GetPeers method

lnclient/models.go
```diff
@@ -173,10 +174,12 @@ type OnchainBalanceResponse struct {
 }
 
 type PeerDetails struct {
-	NodeId      string `json:"nodeId"`
-	Address     string `json:"address"`
-	IsPersisted bool   `json:"isPersisted"`
-	IsConnected bool   `json:"isConnected"`
+	NodeId           string  `json:"nodeId"`
+	NodeAlias        *string `json:"nodeAlias"`
```
rolznz: Can you make a new API model and set it there instead - unfortunately I don't think we can get the alias from the LNClient yet

matjaz: LND has info method, not sure about LDK and others.

lnclient/models.go
matjaz: what do you suggest? new model?

frontend/src/screens/peers/Peers.tsx
matjaz: if done on backend we don't need channels list here.
maybe some client caching would be nice?

lnclient/models.go
```diff
@@ -157,6 +157,7 @@ type CloseChannelResponse struct {
 type PendingBalanceDetails struct {
 	ChannelId     string `json:"channelId"`
 	NodeId        string `json:"nodeId"`
+	NodeAlias     string `json:"nodeAlias"`
```
rolznz, line 160: Can we remove the NodeAlias here too?

frontend/src/screens/channels/Channels.tsx
```diff
@@ -600,9 +573,7 @@ export default function Channels() {
                         to={`https://amboss.space/node/${details.nodeId}`}
                         className="underline"
                       >
-                        {nodes.find(
-                          (node) => node.public_key === details.nodeId
-                        )?.alias || "Unknown"}
+                        {details.nodeAlias || "Unknown"}
```
rolznz, line 576: How is this set?

rolznz:
@matjaz I am now thinking this is a complicated way to do it (especially with the alias for details of pending balances from channel closures).

I now think instead it would be easier and cleaner to add a new hook on the frontend which uses SWR to fetch node details for a single node pubkey.

SWR is easy to configure caching and we can restructure the components to use SWR.

Then we don't need to use a single method which fetches all the aliases in a loop on the frontend, which is not in-line with how the rest of the app data fetching really works right now.

rolznz:
@matjaz yeah, for example on the Peers page instead of having a function like this:

```
const loadNodeStats = React.useCallback(async () => {
    if (!peers) {
      return [];
    }
    const nodes = await Promise.all(
      peers?.map(async (peer): Promise<MempoolNode | undefined> => {
        try {
          const response = await request<MempoolNode>(
            `/api/mempool?endpoint=/v1/lightning/nodes/${peer.nodeId}`
          );
          return response;
        } catch (error) {
          console.error(error);
          return undefined;
        }
      })
    );
    setNodes(nodes.filter((node) => !!node) as MempoolNode[]);
  }, [peers]);
  ```
  
  I would make a new component e.g. 
  
  ```
  export function PeerTableRow(props) {
     const alias = usePeerAlias(props.peerId);
     // render the table row here
     return (
       <TableRow key={peer.nodeId}>
         ...
       </TableRow>
      );
  }
  ```
  
  Then the usePeerAlias hook can cache for a long time (e.g. the whole session) per node ID since it probably won't change.
  
  But what do you mean by the second stage?

PR #1525, opened 2025-07-21
feat: include fees with amount in transaction lists

### LINKED ISSUE #464: Include fees in the balance that left a wallet listed on transaction list
Currently, they are only visible in details.

Aditya-PS-05:
fixes #464 

Changes:

Showed amount + fees in the transactions lists and once user clicks on a particular transaction, it will be shown like this 

Screenshot from 2025-07-22 02-09-15

### DIFF
--- a/frontend/src/components/TransactionItem.tsx
+++ b/frontend/src/components/TransactionItem.tsx
@@ -196,17 +196,31 @@ function TransactionItem({ tx }: Props) {
                   {type == "outgoing" ? "-" : "+"}
                   <span className="font-medium">
                     {new Intl.NumberFormat().format(
-                      Math.floor(tx.amount / 1000)
+                      Math.floor(
+                        type === "outgoing" && tx.state !== "failed"
+                          ? (tx.amount + tx.feesPaid) / 1000
+                          : tx.amount / 1000
+                      )
                     )}
                   </span>
                 </p>
                 <p className="text-muted-foreground">
-                  {Math.floor(tx.amount / 1000) == 1 ? "sat" : "sats"}
+                  {Math.floor(
+                    type === "outgoing" && tx.state !== "failed"
+                      ? (tx.amount + tx.feesPaid) / 1000
+                      : tx.amount / 1000
+                  ) == 1
+                    ? "sat"
+                    : "sats"}
                 </p>
               </div>
               <FormattedFiatAmount
                 className="text-xs md:text-base"
-                amount={Math.floor(tx.amount / 1000)}
+                amount={Math.floor(
+                  type === "outgoing" && tx.state !== "failed"
+                    ? (tx.amount + tx.feesPaid) / 1000
+                    : tx.amount / 1000
+                )}
               />
             </div>
           </div>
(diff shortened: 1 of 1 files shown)


frontend/src/components/TransactionItem.tsx
```diff
@@ -276,13 +308,38 @@ function TransactionItem({ tx }: Props) {
...
             {tx.state != "failed" && type == "outgoing" && (
               <div className="mt-6">
-                <p>Fee</p>
-                <p className="text-muted-foreground">
-                  {new Intl.NumberFormat().format(
-                    Math.floor(tx.feesPaid / 1000)
-                  )}{" "}
-                  {Math.floor(tx.feesPaid / 1000) == 1 ? "sat" : "sats"}
-                </p>
+                <p>Amount Breakdown</p>
```
rolznz: I would make this simpler by just adding a single new "Amount" entry above the fee one. I don't think we need a full breakdown like this

frontend/src/components/TransactionItem.tsx
```diff
@@ -196,17 +196,31 @@ function TransactionItem({ tx }: Props) {
                   {type == "outgoing" ? "-" : "+"}
                   <span className="font-medium">
                     {new Intl.NumberFormat().format(
-                      Math.floor(tx.amount / 1000)
+                      Math.floor(
+                        type === "outgoing" && tx.state !== "failed"
```
rolznz: I don't really like these conditionals and duplication

Can you make a single `amountSatsWithFees` constant and use it in all places.

`const amountSatsWithFees = (tx.amount + tx.feesPaid || 0) / 1000`

Aditya-PS-05: Well, I defined it as 

```
const amountSatsWithFees = Math.floor(
    type === "outgoing" && tx.state !== "failed"
      ? (tx.amount + tx.feesPaid) / 1000
      : tx.amount / 1000
);
```

So that, we don't need `|| 0`.

rolznz: @Aditya-PS-05 there will not be a fee if the payment failed. So there is no need for a conditional

frontend/src/components/TransactionItem.tsx
```diff
@@ -52,6 +52,10 @@ function TransactionItem({ tx }: Props) {
   const [showDetails, setShowDetails] = React.useState(false);
   const type = tx.type;
 
+  const amountSatsWithFees = Math.floor(
+    type === "outgoing" ? (tx.amount + tx.feesPaid) / 1000 : tx.amount / 1000
```
rolznz: I already asked to remove the conditional, and you removed the Math.floor too. Please can you check again

Aditya-PS-05: I misunderstood it.

Aditya-PS-05:
@rolznz , please review it.

rolznz:
@Aditya-PS-05 I've marked for draft now, we have not got consensus within the team that we would like to change the display format.

PR #1535, opened 2025-07-23
fix: ignore sigpipe signal

rolznz:
It is unclear right now where this is coming from but it's causing the app to be restarted for some of our cloud instances. Possibly it could be due to log forwarding or something on the LDK side (I also saw this with one of our users who had a Breez backend). By ignoring it, it should hopefully trigger an error which can be caught and hopefully we get extra logs as to what caused it in the first place

### DIFF
--- a/cmd/http/main.go
+++ b/cmd/http/main.go
@@ -22,7 +22,7 @@ func main() {
 	// Create a channel to receive OS signals.
 	osSignalChannel := make(chan os.Signal, 1)
 	// Notify the channel on os.Interrupt, syscall.SIGTERM. os.Kill cannot be caught.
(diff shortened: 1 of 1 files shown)


rolznz:
@bumi no. It is probably from a network disconnect but I'm not sure what (if it's related to logging as LDK suggested it might be, or any LDK/upstream dependency that does network requests)

bumi:
From how I understand the SIGPIPE signal there must be something that reads/writes from a process. There should no other process be involved I guess. Interesting. 

We don't process any piped writes, do we?

rolznz:
I don't think so, except for our log shipper but it runs as a separate app.

PR #1538, opened 2025-07-25
fix: remove deprecated version tags and replace GetFirst with Find

### LINKED ISSUE #1505: Remove version tag from info event
## Description

Version tag can now be removed

Refer to:

https://github.com/getAlby/hub/blob/e1d9c7963ebb354cab572b5066b9f6495c47a0aa/nip47/event_handler.go#L110

https://github.com/getAlby/hub/blob/e1d9c7963ebb354cab572b5066b9f6495c47a0aa/nip47/publish_nip47_info.go#L69

## Other

`nostr.Tags` `GetFirst` method is also deprecated, so it's good to replace that with `Find` or `FindWithValue` as well.

Aditya-PS-05:
fixes #1505 

Changes: 
Removed the version tag logic and replaced all Tags.GetFirst() calls with Tags.Find() in event handlers and tests.


Aditya-PS-05:
I am concerned that Find("d")[1] can be unsafe, I am thinking to create an helper function may be something like this:

```
func getTagValue(tags nostr.Tags, key string) string {
	if tag := tags.Find(key); tag != nil && len(*tag) > 1 {
		return (*tag)[1]
	}
	return ""
}
```

nip47/event_handler.go
```diff
@@ -102,20 +102,8 @@ func (svc *nip47Service) HandleEvent(ctx context.Context, relay nostrmodels.Rela
...
-	// TODO: Remove version tag after 01-06-2025
-	if encryptionTag == nil {
-		vTag := event.Tags.GetFirst([]string{"v"})
-		if vTag != nil && vTag.Value() != "" {
-			version := vTag.Value()
-			if version == "1.0" {
-				encryption = constants.ENCRYPTION_TYPE_NIP44_V2
-			}
-		}
+	if encryptionTag := event.Tags.Find("encryption"); encryptionTag != nil {
```
rolznz:
```suggestion
	if encryptionTag := event.Tags.Find("encryption"); encryptionTag != nil && len(encryptionTag) > 1 {
```

rolznz: Can we also extract the `encryptionTag := event.Tags.Find("encryption");` to a separate line, I don't like this complicated `if` statement

Aditya-PS-05: Tags.Find() already ensures that the returned tag has at least 2 items (tag name + at least one value). So I guess that `len(encryptionTag) > 1` is reduntant.

rolznz: Cool, thanks for checking!

rolznz: Could you then just split that `if`?

Aditya-PS-05: Done

PR #1539, opened 2025-07-26
feat: add convert to sub-wallet button for isolated apps

### LINKED ISSUE #1533: Allow converting an isolated app to a subwallet
On the app detail page, maybe in some advanced dropdown menu from the topbar (similar to the channels page?) it could be possible to change an isolated app to a sub-wallet by updating the app's metadata.

Note: this should only be possible for users with a paid plan to avoid bypassing the limit

Aditya-PS-05:
fixes #1533 

Changes:
Added the dropdown menu and `handleConvertToSubwallet` function to convert the isolated app to sub-wallet for the paid users.

### DIFF
--- a/frontend/src/screens/apps/ShowApp.tsx
+++ b/frontend/src/screens/apps/ShowApp.tsx
@@ -18,6 +18,7 @@ import {
   AlertCircleIcon,
   PencilIcon,
   Settings2Icon,
+  SquareStackIcon,
   Trash2Icon,
 } from "lucide-react";
 import AppAvatar from "src/components/AppAvatar";
@@ -172,7 +173,7 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
     try {
       const updateAppRequest: UpdateAppRequest = {
         name: app.name,
-        scopes: Array.from(app.scopes),
+        scopes: app.scopes,
         budgetRenewal: app.budgetRenewal,
         expiresAt: app.expiresAt,
         maxAmount: app.maxAmount,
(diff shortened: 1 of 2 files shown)


frontend/src/screens/apps/ShowApp.tsx
```diff
@@ -156,6 +168,39 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
     }
   };
 
+  const handleConvertToSubwallet = async () => {
+    try {
+      const updateAppRequest: UpdateAppRequest = {
+        name: app.name,
```
matjaz, line 175: I recommend all `...app` props are copied.

rolznz: Hmm. I'm a bit hesitant with `...app` as the request payload is not necessarily the same and we end up sending extra properties to the backend.

Although I don't like how it is now. In my opinion ideally we only patch the single field (the metadata field) rather than passing all the data. But we would need more changes for that. I would leave it as-is for now

frontend/src/screens/apps/ShowApp.tsx
```diff
@@ -203,40 +248,66 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
...
-                      disabled={isDeleting}
-                    >
-                      Continue
-                    </AlertDialogAction>
-                  </AlertDialogFooter>
-                </AlertDialogContent>
-              </AlertDialog>
+              <div className="flex gap-2 items-center">
+                {app.isolated &&
+                  app.metadata?.app_store_app_id !==
```
rolznz: I think we should only allow apps that don't have a app store app id. Otherwise they will lose that association

frontend/src/screens/apps/ShowApp.tsx
rolznz: I created a separate issue: https://github.com/getAlby/hub/issues/1546

PR #1540, opened 2025-07-27
Modified MakeInvoice in phoenixd.go 

thebullishbitcoiner:
Description should be checked first (instead of descriptionHash).


bumi:
We have to be careful this might have implications on LNURL (but I also think by now most wallets have removed the description hash requirement)

rolznz:
@bumi in LDK we check the description first https://github.com/getAlby/hub/blob/master/lnclient/ldk/ldk.go#L750

I think we decided enough wallets have dropped this validation

PR #1541, opened 2025-07-27
feat: add JSON export for subaccount transaction history

### LINKED ISSUE #1337: Accounting feature: Add Subaccount/NWC Identifier or filters for transaction history

Aditya-PS-05:
fixes #1337 

## Summary
- Added the download JSON link for subaccount-specific transactions. 
- Works for both Wails and http.


rolznz:
@Aditya-PS-05 ok, thank you - please tag me when it is ready for review

Aditya-PS-05:
> @Aditya-PS-05 ok, thank you - please tag me when it is ready for review

@rolznz , you can review it.

frontend/src/components/TransactionsList.tsx
```diff
@@ -53,6 +117,26 @@ function TransactionsList({
...
+                    <MoreHorizontalIcon className="h-4 w-4" />
+                  </Button>
+                </DropdownMenuTrigger>
+                <DropdownMenuContent align="end">
+                  <DropdownMenuItem
+                    className="flex flex-row items-center gap-2 cursor-pointer"
+                    onClick={handleDownloadTransactions}
+                  >
+                    <DownloadIcon className="h-4 w-4" />
+                    Export CSV (up to 20)
```
rolznz: What I mean is we should download all the transactions by paginating until there is none left

frontend/src/components/TransactionsList.tsx
```diff
@@ -53,6 +117,26 @@ function TransactionsList({
         />
       ) : (
         <>
+          {appId && transactions.length > 0 && (
```
rolznz: why only for a subaccount? can it also be for the main transaction list?

frontend/src/components/TransactionsList.tsx
```diff
@@ -53,6 +117,26 @@ function TransactionsList({
...
+            <div className="mb-4 flex justify-end">
+              <DropdownMenu>
```
rolznz: This dropdown menu should be inline with the "Transactions" title so we don't add extra vertical space. Probably best to be moved to where the title is rendered

frontend/src/components/TransactionsList.tsx
```diff
@@ -53,6 +117,26 @@ function TransactionsList({
...
+                <DropdownMenuTrigger asChild>
+                  <Button size="icon" variant="ghost">
+                    <MoreHorizontalIcon className="h-4 w-4" />
+                  </Button>
+                </DropdownMenuTrigger>
+                <DropdownMenuContent align="end">
+                  <DropdownMenuItem
```
rolznz: We should make this a subsciber feature - unfortunately for dropdown items it needs to be rendered conditionally - see how the `Set Node Alias` menu item is done

Aditya-PS-05: Done

frontend/src/components/TransactionsList.tsx
Aditya-PS-05: Done

frontend/src/components/TransactionsList.tsx
Aditya-PS-05: Done

frontend/src/components/TransactionsList.tsx
Aditya-PS-05: Done

Aditya-PS-05:
@rolznz , please review it.

frontend/src/components/TransactionsList.tsx
```diff
@@ -1,12 +1,129 @@
...
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
+    window.URL.revokeObjectURL(url);
+  } catch (error) {
+    console.error("Error downloading transactions:", error);
+  }
+};
+
+export const TransactionsExportMenu = ({ appId }: { appId?: number }) => {
```
rolznz, line 92: Can you move this to a separate file (+ all the export function code)

frontend/src/components/TransactionsList.tsx
```diff
@@ -53,6 +170,11 @@ function TransactionsList({
         />
       ) : (
         <>
+          {!appId && transactions.length > 0 && (
+            <div className="mb-4 flex justify-end">
+              <TransactionsExportMenu />
```
rolznz, line 175: This is still in the wrong place. It should be rendered next to the swap and buy bitcoin buttons, otherwise there is a huge gap introduced.

image

frontend/src/components/TransactionsList.tsx
Aditya-PS-05: Sure

frontend/src/components/TransactionsList.tsx
```diff
@@ -1,12 +1,129 @@
...
+      if (data.transactions && data.transactions.length > 0) {
+        allTransactions = [...allTransactions, ...data.transactions];
+        offset += LIST_TRANSACTIONS_LIMIT;
+      } else {
+        hasMoreTransactions = false;
+      }
+    }
+
+    // Convert to CSV and create download
+    const csvString = convertToCSV(allTransactions);
```
rolznz, line 74: Did you also test this in Wails mode? does it work fine?

frontend/src/components/TransactionsList.tsx
```diff
@@ -1,12 +1,129 @@
...
+
+  return [csvHeaders, ...csvRows].join("\n");
+};
+
+const handleDownloadTransactions = async (appId?: number) => {
+  try {
+    // Fetch all transactions by paginating through all pages
+    let allTransactions: Transaction[] = [];
+    let offset = 0;
+    let hasMoreTransactions = true;
```
rolznz, line 55: no need for this variable, you can just do a `while(true)` and `break`

frontend/src/components/TransactionsList.tsx
```diff
@@ -1,12 +1,129 @@
...
+
+    while (hasMoreTransactions) {
+      let url = `/api/transactions?limit=${LIST_TRANSACTIONS_LIMIT}&offset=${offset}`;
+      if (appId) {
+        url += `&appId=${appId}`;
+      }
+
+      const data = (await request(url)) as ListTransactionsResponse;
+
+      if (data.transactions && data.transactions.length > 0) {
```
rolznz, line 65: here I would just do:
```
if (!data.transactions.length) {
  break;
}
```

This will also remove a redundant request if the last page is not full.

frontend/src/components/TransactionsList.tsx
```diff
@@ -1,12 +1,129 @@
...
+    let offset = 0;
+    let hasMoreTransactions = true;
+
+    while (hasMoreTransactions) {
+      let url = `/api/transactions?limit=${LIST_TRANSACTIONS_LIMIT}&offset=${offset}`;
+      if (appId) {
+        url += `&appId=${appId}`;
+      }
+
+      const data = (await request(url)) as ListTransactionsResponse;
```
rolznz, line 63: can you check how we do this elsewhere (don't use brackets around the await, and throw an exception if the response is not truthy)

frontend/src/components/TransactionsList.tsx
```diff
@@ -1,12 +1,129 @@
...
+import { ListTransactionsResponse, Transaction } from "src/types";
+import { request } from "src/utils/request";
+
+const convertToCSV = (transactions: Transaction[]) => {
+  if (!transactions.length) {
+    return "";
+  }
+
+  // Get headers from the first transaction
+  const headers = Object.keys(transactions[0]);
```
rolznz, line 27: This is not a reliable way, because some fields are optional.

For example doing it this way, the `metadata` field is missing in the CSV.

Maybe here we need to manually maintain a list of columns we want.

frontend/src/components/TransactionsList.tsx
```diff
@@ -1,12 +1,129 @@
...
+  const csvHeaders = headers.join(",");
+
+  // Convert each transaction to CSV row
+  const csvRows = transactions.map((tx) => {
```
rolznz, line 31: Here we do csv generation manually, isn't there a library we can use? I think this will be safer to ensure we don't make a mistake with the encoding

rolznz:
Marking as draft for now, needs some additional work

frontend/src/components/TransactionsList.tsx
rolznz: That or we merge the headers from all the rows?

Aditya-PS-05:
> @Aditya-PS-05 do you still have time to work on this?

I was bit busy.. Yeah, I have time. I will fix it shortly.

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/1672

Issue #1527, opened 2025-07-22
Add pagination/filtering on api/apps endpoint


rolznz:
Currently we fetch all apps. This is too inefficient once a user has >1000 apps and can take over 5 seconds to fetch, blocking the database and causing other queries to fail. It's especially slow because it also fetches all app permissions, and for every isolated app also calculates the balance.

For 6000 apps this takes 20 seconds to return the response.

PR #1543, opened 2025-07-28
chore: add apps pagination

rolznz:
Closes https://github.com/getAlby/hub/issues/1527

This PR significantly improves performance for hubs with many apps, by fetching only a single page of apps rather than all. A lot of places in the hub we fetched all the apps which was often unnecessary - all those places have been fixed.

App and subwallet pages now use the same pagination component as the transaction list, which has been extracted.

Tested on a hub with 5000 apps


im-adithya (approved):
tACK, this is a HUGE improvement 🚀 💪

PR #1550, opened 2025-07-29
chore: useNodeDetails hook for loading node alias

matjaz:
based on ideas from #1510

### DIFF
--- a/frontend/src/screens/peers/Peers.tsx
+++ b/frontend/src/screens/peers/Peers.tsx
@@ -118,7 +118,7 @@ function PeerTableRow(props: PeerTableRowProps) {
       <TableCell>{peer.nodeId}</TableCell>
       <TableCell>{peer.address}</TableCell>
       <TableCell>
-        {peer.isConnected && !hasOpenedChannels(peer) && (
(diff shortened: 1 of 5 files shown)


frontend/src/screens/peers/Peers.tsx
```diff
@@ -118,7 +118,7 @@ function PeerTableRow(props: PeerTableRowProps) {
       <TableCell>{peer.nodeId}</TableCell>
       <TableCell>{peer.address}</TableCell>
       <TableCell>
-        {peer.isConnected && !hasOpenedChannels(peer) && (
```
matjaz: why?

rolznz: @matjaz I had a disconnected peer I didn't want any more.

Alby Hub will continually try to reconnect to its peers.

matjaz: Got it. So actually that action is "Remove peer".

rolznz: Yeah, maybe that's better copy. Although there is no reason just to disconnect a peer and keep it :thinking:

Issue #1551, opened 2025-07-30
Bug: 401 errors after changing unlock password


rolznz:
After changing the unlock password we generate a new token, but the http service still expects the old one. The hub runs but all secured endpoints will return 401.

https://github.com/getAlby/hub/blob/master/http/http_service.go#L114

A workaround is to completely restart your hub (for Alby Cloud - go to https://getalby.com/hub_deployment/edit)

PR #1552, opened 2025-07-30
fix: use dynamic JWT key func

rolznz:
Fixes https://github.com/getAlby/hub/issues/1551

This was introduced when we changed the JWT secret when the password was changed. I think I didn't test it well, by only testing I could change my password and then change it back, without properly testing using the hub with the new password :disappointed: 

An invalid token correctly returns 401 on all endpoints and redirects the user to the unlock screen.

The problem before was the middleware returned 401, but we return an `unlocked` boolean in the info endpoint, which is based on comparing against the correct token. This is why the hub was in a weird state ("unlocked" but all endpoints were returning 401)

### DIFF
--- a/http/http_service.go
+++ b/http/http_service.go
@@ -111,7 +111,10 @@ func (httpSvc *HttpService) RegisterSharedRoutes(e *echo.Echo) {
 		NewClaimsFunc: func(c echo.Context) jwt.Claims {
 			return new(jwtCustomClaims)
 		},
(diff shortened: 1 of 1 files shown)


bumi:
great find. so this dynamically loads the key? instead of once?

rolznz:
> great find. so this dynamically loads the key? instead of once?

Thanks. Yeah, it loads it at request time

PR #1553, opened 2025-07-30
fix: signal handling after ignoring sigpipe signal

rolznz:
Currently if a SIGPIPE signal is captured, we cannot gracefully exit the application.

Tested by sending 2 SIGPIPES and then a SIGINT


im-adithya (approved):
LGTM, couldn't test it though

PR #1560, opened 2025-08-01
chore: check password before node start

matjaz:


http/http_service.go
```diff
@@ -269,6 +269,12 @@ func (httpSvc *HttpService) startHandler(c echo.Context) error {
 		})
 	}
 
+	if !httpSvc.cfg.CheckUnlockPassword(startRequest.UnlockPassword) {
```
rolznz, line 272: Internally it also checks, is this to return a better error message?

matjaz: yeah, instant feedback + no need to call node start

rolznz: Note: we could do the same with the Wails version. But I don't think it's that important.

Thanks!

Issue #1565, opened 2025-08-04
Bug: swaps service accesses LNClient after it has been shut down


rolznz:
See https://github.com/getAlby/hub/blob/master/swaps/swaps_service.go#L199 - it will cause a panic.

I think this can be reproduced by stopping the node from the settings page.

Aditya-PS-05:
@rolznz , can I work on this issue?

im-adithya:
Hey Aditya, the setup is quite complicated and I'm already taking a look, feel free to take up any other issue!

PR #1572, opened 2025-08-05
fix: stop autoswaps on shutting down ln client

im-adithya:
Fixes https://github.com/getAlby/hub/issues/1565

Tested on LDK

### DIFF
--- a/service/start.go
+++ b/service/start.go
@@ -287,7 +287,7 @@ func (svc *service) StartApp(encryptionKey string) error {
 		return err
 	}
 
-	svc.swapsService = swaps.NewSwapsService(svc.ctx, svc.db, svc.cfg, svc.keys, svc.eventPublisher, svc.lnClient, svc.transactionsService)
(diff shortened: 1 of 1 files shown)


service/stop.go
```diff
@@ -36,6 +36,9 @@ func (svc *service) stopLNClient() {
 		})
 		return
 	}
+
+	svc.swapsService.StopAutoSwapOut()
```
rolznz: minor: shouldn't this be done before the lnclient is shutdown?

rolznz: Edit: I don't think this is the right place, and isn't the responsibility of this method (`stopLNClient`)

rolznz: I'm wondering how this is happening actually, is it a race condition? because we only call `stopLNClient` when the context is cancelled. And the swaps service should also stop when that context is cancelled

im-adithya: Whoops I found the actual fix, changed

rolznz: Nice find!

PR #1575, opened 2025-08-05
feat: autoswaps using xpubs

frnandu:
fixes 1480

### DIFF
--- a/swaps/swaps_service.go
+++ b/swaps/swaps_service.go
@@ -17,6 +17,9 @@ import (
 
 	"github.com/BoltzExchange/boltz-client/v2/pkg/boltz"
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/btcutil/hdkeychain"
+	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/getAlby/hub/config"
 	"github.com/getAlby/hub/constants"
 	"github.com/getAlby/hub/db"
@@ -46,6 +49,7 @@ type swapsService struct {
 	boltzWs             *boltz.Websocket
 	swapListeners       map[string]chan boltz.SwapUpdate
 	swapListenersLock   sync.Mutex
+	autoSwapXpub        string // Decrypted xpub for auto swaps
 }
 
 type SwapsService interface {
@@ -191,7 +195,7 @@ func (svc *swapsService) EnableAutoSwapOut() error {
 	logger.Logger.Info("Starting auto swap workflow")
 
 	go func() {
-		ticker := time.NewTicker(1 * time.Hour)
+		ticker := time.NewTicker(1 * time.Minute)
 		for {
 			select {
 			case <-ticker.C:
@@ -207,13 +211,35 @@ func (svc *swapsService) EnableAutoSwapOut() error {
 					logger.Logger.Info("Threshold requirements not met for swap, ignoring")
 					continue
 				}
+
+				actualDestination := swapDestination
+				var addressIndex uint32
+				if swapDestination != "" {
+					if err := svc.validateXpub(swapDestination); err == nil {
+						actualDestination, addressIndex, err = svc.getNextUnusedAddressFromXpub()
+						if err != nil {
+							logger.Logger.WithError(err).Error("Failed to get next address from xpub")
+							actualDestination = "" // Fallback to empty address if we can't get one
+						}
+					}
+				}
+
 				logger.Logger.WithFields(logrus.Fields{
 					"amount":      amount,
-					"destination": swapDestination,
+					"destination": actualDestination,
 				}).Info("Initiating swap")
-				_, err = svc.SwapOut(amount, swapDestination, true)
+				swapResponse, err := svc.SwapOut(amount, actualDestination, true)
 				if err != nil {
 					logger.Logger.WithError(err).Error("Failed to swap")
+				} else if addressIndex > 0 {
+					err = svc.cfg.SetUpdate(config.AutoSwapXpubIndexStart, strconv.FormatUint(uint64(addressIndex), 10), "")
+					if err != nil {
+						logger.Logger.WithError(err).Error("Failed to update xpub index start after creating swap")
+					}
+					logger.Logger.WithFields(logrus.Fields{
+						"swapId":    swapResponse.SwapId,
+						"nextIndex": addressIndex,
+					}).Info("Updated xpub index start for swap address")
 				}
 			case <-ctx.Done():
 				logger.Logger.Info("Stopping auto swap workflow")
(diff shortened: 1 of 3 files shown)


swaps/swaps_service.go
```diff
@@ -207,13 +210,35 @@ func (svc *swapsService) EnableAutoSwapOut() error {
...
 				}
+
+				actualDestination := swapDestination
+				var addressIndex uint32
+				if swapDestination != "" {
+					if err := svc.validateXpub(swapDestination); err == nil {
+						actualDestination, addressIndex, err = svc.getNextUnusedAddressFromXpub()
+						if err != nil {
+							logger.Logger.WithError(err).Error("Failed to get next address from xpub")
+							actualDestination = "" // Fallback to empty address if we can't get one
```
rolznz: I think we should fail the swap instead. And we should verify the xpub when the user submits the auto swap details

swaps/swaps_service.go
```diff
@@ -207,13 +210,35 @@ func (svc *swapsService) EnableAutoSwapOut() error {
...
 				logger.Logger.WithFields(logrus.Fields{
 					"amount":      amount,
-					"destination": swapDestination,
+					"destination": actualDestination,
 				}).Info("Initiating swap")
-				_, err = svc.SwapOut(amount, swapDestination, true)
+				swapResponse, err := svc.SwapOut(amount, actualDestination, true)
 				if err != nil {
 					logger.Logger.WithError(err).Error("Failed to swap")
+				} else if addressIndex > 0 {
```
rolznz: We try not to use if/else where avoidable. To be consistent inside the error check it should call `continue` like the other conditionals above

swaps/swaps_service.go
rolznz: We can just log the error and then do `continue`

swaps/swaps_service.go
rolznz: Also why do we check if `addressIndex > 0`?

Can't we just always do `SetUpdate` with `addressIndex + 1`?

rolznz: I think we should bump the address index immediately after getting an unused one too, rather than waiting for the swap to succeed (just as a precaution) - currently it will do it after the swap is completely finished

rolznz:
@frnandu tested using Sparrow wallet with a `tpub` connected to regtest (boltz setup) and it works well! Just minor feedback

swaps/swaps_service.go
frnandu: > Also why do we check if `addressIndex > 0`?

This is was is used for deciding if an xpub is being used and whether we need to update storage.
We could also remember the result of  `svc.validateXpub(swapDestination)` if you think that is better.

> Can't we just always do `SetUpdate` with `addressIndex + 1`?

I had it like that before, but if for some reason the swap creation fails (boltz is down for example) but `err == nil` it could fall into a retry cycle and increment it on each try until it finally is successful, which would make a lot of addresses between the last used and the swap address empty, and with a smal gap limit wallets will not see the funds.
Maybe if we check the `SwapResponse` and increment only when a paymentHash is properly returned?

frnandu: > I think we should bump the address index immediately after getting an unused one too, rather than waiting for the swap to succeed (just as a precaution) - currently it will do it after the swap is completely finished

Precaution against parallel auto swaps or what specifically?
Are parallel auto swaps execution possible to occur?
I guess if they are trigger every 1h it could happen that we still miss the first onchain confirmation for an ongoing swap when another is triggered to start. Is that what you mean?

frnandu: Added swap response checking for filled `paymentHash` and `swapId` and only then  `SetUpdate` with `addressIndex + 1`

rolznz: Ah ok, good point about the gap limit.

> Precaution against parallel auto swaps or what specifically? Are parallel auto swaps execution possible to occur?

Yeah, I managed locally to have 2 swaps use the same address. I'm not sure how it happened because the swaps should have executed sequentially. CC @im-adithya

swaps/swaps_service.go
```diff
@@ -207,13 +210,37 @@ func (svc *swapsService) EnableAutoSwapOut() error {
...
-					"destination": swapDestination,
+					"destination": actualDestination,
 				}).Info("Initiating swap")
-				_, err = svc.SwapOut(amount, swapDestination, true)
+				swapResponse, err := svc.SwapOut(amount, actualDestination, true)
 				if err != nil {
 					logger.Logger.WithError(err).Error("Failed to swap")
+					continue
+				}
+				if addressIndex > 0 && swapResponse != nil && swapResponse.SwapId != "" && swapResponse.PaymentHash != "" {
```
rolznz: Why do we need all these checks? they should all be set if `err` is nil. We should rely on the method to either return an error or a valid response.

swaps/swaps_service.go
frnandu: How did you trigger the swaps? did you changed ticker frequency or somehow manually?
I guess without incrementing `AutoSwapXpubIndexStart` after swap creation and if second swap is triggered before first chain confirmation, the function to get next address from xpub will not see still any transactions on that index address from esplora and would return still the same address as previous swap.
Now with storing +1 after checking that swap has paymentHash (meaning should be in progress to confirm) this should not be a problem anymore I think.

swaps/swaps_service.go
frnandu: `addressIndex > 0` is for detecting xpub usage, otherwise the autoswap will be using a single address and we don't need to update `config.AutoSwapXpubIndexStart`.
Rest is not necessary you're right. 
I initially was thinking that swapResponse could contain something else with reason for failed swap or something.

swaps/swaps_service.go
frnandu: Actually checking `swapResponse` for `paymentHash` is not a bullet proof way of making sure the swap will finish, because it's just creating it and will return a `paymentHash` even if later it will fail the swap for some reason.
I was testing with mutinynet and got this:
```
{"level":"info","msg":"Swap created","swapId":"xmwgdVsgcZRJ","time":"2025-08-13T23:55:35+02:00"}
{"level":"info","msg":"Updated xpub index start for swap address","nextIndex":5,"swapId":"xmwgdVsgcZRJ","time":"2025-08-13T23:55:35+02:00"}
{"level":"error","msg":"Failed to receive update from boltz","swap_id":"xmwgdVsgcZRJ","time":"2025-08-13T23:55:36+02:00"}
{"level":"info","msg":"Subscribed to boltz websocket","swapId":"xmwgdVsgcZRJ","time":"2025-08-13T23:55:36+02:00"}
{"error":"Network signet not supported","level":"error","msg":"Failed to parse network","swapId":"xmwgdVsgcZRJ","time":"2025-08-13T23:55:36+02:00"}
```
Which means that it is incrementing the `AutoSwapXpubIndexStart` after swap initialization, but it is not finishing the swap, which means those addresses will be empty. And if the reason for swap failure in boltz is temporary (like node down or lack of liquidity) and later after few retries it will succeed, we still can get the issue of too big gap of empty addresses right?

Is the only way to avoid that to make sure we only increment `AutoSwapXpubIndexStart` when we get the confirmation that swap is finished (and funds should be in the address confirmed) combined with making sure we don't ever start auto swaps in parallel. Do you see other solution?

swaps/swaps_service.go
rolznz: How does it work if `addressIndex` is 0? which should be the case when you use a brand new xpub right? then after the swap succeeds this check will not be hit? (But it seems to be hit :question: )

frnandu: you're right, it could potentially be 0
changed it to use `usedXpubDerivation` bol

rolznz: Cool

swaps/swaps_service.go
rolznz: I think it should be impossible to trigger 2 swaps at once. I'm confused as to how this can happen. Maybe I restarted the hub while the swap was happening. I think that could be why. Then it would process the original swap while creating a new one and both use the same address. I set the timer to 1 minute from 1 hour to test the swaps, which is probably why both swaps happened in parallel. But I think in production this will be a super rare edge case.

rolznz: Edit: the swap runs in a goroutine so I was wrong.

swaps/swaps_service.go
```diff
@@ -1317,3 +1346,134 @@ func (svc *swapsService) doMempoolRequest(endpoint string, result interface{}) e
...
+	destination, _ := svc.cfg.Get(config.AutoSwapDestinationKey, "")
+	if destination == "" {
+		return "", 0, errors.New("no destination configured")
+	}
+
+	if err := svc.validateXpub(destination); err != nil {
+		return "", 0, errors.New("destination is not a valid XPUB")
+	}
+
+	indexStr, _ := svc.cfg.Get(config.AutoSwapXpubIndexStart, "0")
```
rolznz: the 2nd parameter is the encryption key, this should be empty. I will fix it in a following commit

swaps/swaps_service.go
rolznz: Resolved in https://github.com/getAlby/hub/pull/1609

swaps/swaps_service.go
rolznz: resolved in https://github.com/getAlby/hub/pull/1609

swaps/swaps_service.go
```diff
@@ -1317,3 +1346,134 @@ func (svc *swapsService) doMempoolRequest(endpoint string, result interface{}) e
...
+	index, _ := strconv.ParseUint(indexStr, 10, 32)
```
rolznz: Here also the parsing will fail if the key did not exist (trying to parse an empty string) - resolved in https://github.com/getAlby/hub/pull/1609

PR #1576, opened 2025-08-06
fix: improve readability of qr code

reneaaron:


frontend/src/screens/wallet/Receive.tsx
```diff
@@ -43,6 +43,7 @@ export default function Receive() {
             <CardContent className="flex flex-col items-center gap-6 pt-6">
               <div className="relative flex items-center justify-center">
                 <QRCode
+                  level="Q"
```
im-adithya: Can you also add this to other instances as well where there's an image overlay, ex: in `frontend/src/screens/apps/AppCreated.tsx` etc.?

rolznz: We had problems in general for lightning addresses. In Alby Go we don't show the avatar on lightning addresses. Should we remove it here?

reneaaron: Can you remember what exactly the problems were we had back then? One would think that with enough error correction that shouldn't be a problem, right?

rolznz: Yeah, that + the avatar was too big. But it still has a negative effect on scanning, especially I think if the lightning address is shorter

reneaaron: Okay, so that means we just remove it everywhere for now? It's definitely better in terms of readability if there are no overlays.

rolznz: I would probably remove it also because this is a personal app and is not shown to others unlike Alby Go for example. I don't think there's any benefit to show an icon here. What do you think?

im-adithya: We agreed to remove the avatar for now to improve readability

Issue #1547, opened 2025-07-29
Add swaps to send/receive flows


rolznz:
See figma "send to on-chain (swap)" https://www.figma.com/design/2IyilDewDrdNP3G1NixPYd/%F0%9F%8C%93-Alby-Hub?node-id=8708-130741&t=NqAbqIYjwkQmyapX-0

See figma "receive from on-chain (swap)" https://www.figma.com/design/2IyilDewDrdNP3G1NixPYd/%F0%9F%8C%93-Alby-Hub?node-id=8708-130742&t=NqAbqIYjwkQmyapX-0

PR #1582, opened 2025-08-07
feat: add swap out and onchain withdraw to send flow

im-adithya:
Closes https://github.com/getAlby/hub/issues/1547

## TODOs

- [ ] Option to set fee rate for swap outs -> reached out to boltz, can be done in a follow-up?
- [x] I think fee rate while sending to Onchain addresses can be improved
- [x] We have a warning which is based on NextMaxSpendableMPP in ConfirmPayment screen, but acc to the design we show the spending balance below the confirm button, how to proceed here?
- [x] Check if `PendingPaymentAlert` is used enough, should we also keep it in ZeroAmount screen and Send screen (my reasoning to not include it in the initial screen is because we also have onchain now)
- [x] I used different forms for Onchain and Swap screens, and that causes amount to reset on switching, is that okay?
- [x] What to do with Deposit and Withdraw pages in Node screen?
- [x] Do we have to show the red text for spending balance with increase button while sending to onchain address too (via swap)?
- [x] Make swap out the default for sending on-chain payments
- [x] How can we notify people of onchain transactions once confirmed?
- [x] Change Swap Out description in tx list `Swap to bcrt1q57g52nzxsat7hfshw9dtf4swr3tv50hdax52nd`
- [x] Show onchain fee information correctly
- [x] Recipient looks weird when sending to BTC address (design input)
- [x] Pay Invoice and 0-amount screen don't look good (design input)

### Do we want?

- [x] Confirm payment before sending to ln address in LnurlPay screen (design input)
- [x] Should we show balance card with increase button (design input)
- [x] Split forms into different components

## Description

This PR improves Send screen to match the Alby Hub 2.0 design and mainly adds option to send to BTC addresses in the Send screen by Withdrawing/Swapping. Part 1 of #1547 

For swaps, we currently just redirect the user to `SwapOutStatus` page that we already have to reduce complexity. We can maybe tackle that in a follow up step where we improve both `SwapOutStatus` and `OnchainSwapSend` screens

## Video

https://github.com/user-attachments/assets/38576d4e-3ee5-4667-ae6f-242de82e9749

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ _To configure via env, the following parameters must be provided:_
 - `LDK_ESPLORA_SERVER`: By default the optimized Alby esplora is used. You can configure your own esplora server (note: the public blockstream one is slow and can cause onchain syncing and issues with opening channels)
 - `LDK_VSS_URL`: Use VSS (encrypted remote storage) rather than local sqlite store for lightning and bitcoin data. Currently this feature only works for brand new Alby Hub instances that are connected to Alby Accounts with an active subscription plan.
 - `LDK_LISTENING_ADDRESSES`: configure listening addresses, required for public channels, and ideally reachable if you would like others to be able to initiate peering with your node.
+- `LDK_ANNOUNCEMENT_ADDRESSES`: configure announcement addresses (only required if you use a VPN)
 - `LDK_MAX_CHANNEL_SATURATION`: Sets the maximum portion of a channel's total capacity that may be used for sending a payment, expressed as a power of 1/2. See `max_channel_saturation_power_of_half` in [LDK docs](https://docs.rs/lightning/latest/lightning/routing/router/struct.PaymentParameters.html#structfield.max_channel_saturation_power_of_half).
 - `LDK_MAX_PATH_COUNT`: Maximum number of paths that may be used by MPP payments.
 
--- a/api/api.go
+++ b/api/api.go
@@ -423,6 +423,11 @@ func (api *api) ListApps(limit uint64, offset uint64, filters ListAppsFilters, o
 	if orderBy == "" {
 		orderBy = "last_used_at"
 	}
+	if orderBy == "last_used_at" {
+		// when ordering by last used at, apps with last_used_at is NULL should be ordered last
+		orderBy = "last_used_at IS NULL, " + orderBy
+	}
+
 	query = query.Order(orderBy + " DESC")
 
 	if limit == 0 {
@@ -780,7 +785,7 @@ func (api *api) InitiateSwapOut(ctx context.Context, initiateSwapOutRequest *Ini
 		return nil, errors.New("invalid swap amount")
 	}
 
-	swapOutResponse, err := api.svc.GetSwapsService().SwapOut(amount, destination, initiateSwapOutRequest.UseExactReceiveAmount, false)
+	swapOutResponse, err := api.svc.GetSwapsService().SwapOut(amount, destination, initiateSwapOutRequest.UseExactReceiveAmount, false, false)
 	if err != nil {
 		logger.Logger.WithFields(logrus.Fields{
 			"amount":      amount,
--- a/config/models.go
+++ b/config/models.go
@@ -12,6 +12,7 @@ const (
 	AutoSwapBalanceThresholdKey = "AutoSwapBalanceThreshold"
 	AutoSwapAmountKey           = "AutoSwapAmount"
 	AutoSwapDestinationKey      = "AutoSwapDestination"
+	AutoSwapXpubIndexStart      = "AutoSwapXpubIndexStart"
 )
 
 type AppConfig struct {
@@ -35,6 +36,7 @@ type AppConfig struct {
 	LDKMaxPathCount                    uint8  `envconfig:"LDK_MAX_PATH_COUNT" default:"5"`
 	LDKVssUrl                          string `envconfig:"LDK_VSS_URL" default:"https://vss.getalbypro.com/vss"`
 	LDKListeningAddresses              string `envconfig:"LDK_LISTENING_ADDRESSES" default:"0.0.0.0:9735,[::]:9735"`
+	LDKAnnouncementAddresses           string `envconfig:"LDK_ANNOUNCEMENT_ADDRESSES"`
 	LDKTransientNetworkGraph           bool   `envconfig:"LDK_TRANSIENT_NETWORK_GRAPH" default:"false"`
 	RebalanceServiceUrl                string `envconfig:"REBALANCE_SERVICE_URL" default:"https://megalithic.me"`
 	LDKBitcoindRpcHost                 string `envconfig:"LDK_BITCOIND_RPC_HOST"`
--- a/db/migrations/202508151405_swap_xpub.go
+++ b/db/migrations/202508151405_swap_xpub.go
@@ -1,0 +1,25 @@
+package migrations
+
+import (
+	_ "embed"
+
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// NOTE: This was actually 2025-07-21
+var _202508151405_swap_xpub = &gormigrate.Migration{
+	ID: "202508151405_swap_xpub",
+	Migrate: func(tx *gorm.DB) error {
+
+		err := tx.Exec("ALTER TABLE swaps ADD COLUMN used_xpub BOOLEAN;").Error
+		if err != nil {
+			return err
+		}
+
+		return nil
+	},
+	Rollback: func(tx *gorm.DB) error {
+		return nil
+	},
+}
--- a/db/migrations/migrate.go
+++ b/db/migrations/migrate.go
@@ -35,6 +35,7 @@ func Migrate(gormDB *gorm.DB) error {
 		_202508041737_postgres_amount_bigint,
 		_202508041738_app_last_used,
 		_202508041739_response_events_index,
+		_202508151405_swap_xpub,
 	})
 
 	return m.Migrate()
--- a/db/models.go
+++ b/db/models.go
@@ -106,6 +106,7 @@ type Swap struct {
 	LockupTxId         string
 	ClaimTxId          string
 	AutoSwap           bool
+	UsedXpub           bool
 	TimeoutBlockHeight uint32
 	BoltzPubkey        string
 	SwapTree           datatypes.JSON
--- a/frontend/src/screens/wallet/index.tsx
+++ b/frontend/src/screens/wallet/index.tsx
@@ -14,7 +14,6 @@ import ExternalLink from "src/components/ExternalLink";
 import FormattedFiatAmount from "src/components/FormattedFiatAmount";
 import Loading from "src/components/Loading";
 import LowReceivingCapacityAlert from "src/components/LowReceivingCapacityAlert";
-import ResponsiveButton from "src/components/ResponsiveButton";
 import TransactionsList from "src/components/TransactionsList";
 import {
   Alert,
@@ -44,12 +43,22 @@ function Wallet() {
         description=""
         contentRight={
           <>
-            <Link to="/wallet/receive">
-              <ResponsiveButton icon={ArrowDownIcon} text="Receive" size="lg" />
-            </Link>
-            <Link to="/wallet/send">
-              <ResponsiveButton icon={ArrowUpIcon} text="Send" size="lg" />
-            </Link>
+            <LinkButton
+              to="/wallet/receive"
+              size="lg"
+              className="hidden xl:inline-flex !px-12"
+            >
+              <ArrowDownIcon />
+              Receive
+            </LinkButton>
+            <LinkButton
+              to="/wallet/send"
+              size="lg"
+              className="hidden xl:inline-flex !px-12"
+            >
+              <ArrowUpIcon />
+              Send
+            </LinkButton>
           </>
         }
       />
@@ -90,8 +99,8 @@ function Wallet() {
           </AlertDescription>
         </Alert>
       )}
-      <div className="flex flex-col xl:flex-row justify-between xl:items-start gap-5">
-        <div className="flex flex-col gap-1 text-center xl:text-left">
+      <div className="flex flex-col xl:flex-row justify-between xl:items-start gap-3">
+        <div className="flex flex-col gap-1 p-6 xl:p-0 text-center xl:text-left">
           <div className="text-5xl font-medium balance sensitive slashed-zero">
             {new Intl.NumberFormat().format(
               Math.floor(balances.lightning.totalSpendable / 1000)
@@ -102,6 +111,14 @@ function Wallet() {
             className="text-xl"
             amount={balances.lightning.totalSpendable / 1000}
           />
+        </div>
+        <div className="grid grid-cols-2 items-center gap-3 xl:hidden">
+          <LinkButton to="/wallet/receive" size="lg">
+            Receive
+          </LinkButton>
+          <LinkButton to="/wallet/send" size="lg">
+            Send
+          </LinkButton>
         </div>
         <div className="grid grid-cols-2 items-center gap-3">
           <ExternalLink to="https://www.getalby.com/topup">
--- a/frontend/src/screens/wallet/swap/AutoSwap.tsx
+++ b/frontend/src/screens/wallet/swap/AutoSwap.tsx
@@ -64,6 +64,9 @@ function AutoSwapOutForm() {
   const [balanceThreshold, setBalanceThreshold] = useState("");
   const [swapAmount, setSwapAmount] = useState("");
   const [destination, setDestination] = useState("");
+  const [externalType, setExternalType] = useState<"address" | "xpub">(
+    "address"
+  );
   const [loading, setLoading] = useState(false);
 
   const onSubmit = async (e: React.FormEvent) => {
@@ -194,23 +197,78 @@ function AutoSwapOutForm() {
         </RadioGroup>
       </div>
       {!isInternalSwap && (
-        <div className="grid gap-1.5">
-          <Label>Receiving on-chain address</Label>
-          <div className="flex gap-2 mb-4">
-            <Input
-              placeholder="bc1..."
-              value={destination}
-              onChange={(e) => setDestination(e.target.value)}
-              required
-            />
-            <Button
-              type="button"
-              variant="outline"
-              className="px-2"
-              onClick={paste}
+        <div className="grid gap-4">
+          <div className="flex flex-col gap-3">
+            <Label>Destination Type</Label>
+            <RadioGroup
+              value={externalType}
+              onValueChange={(value) => {
+                setExternalType(value as "address" | "xpub");
+                setDestination("");
+              }}
+              className="flex gap-4 flex-row"
             >
-              <ClipboardPasteIcon className="w-4 h-4" />
-            </Button>
+              <div className="flex items-start space-x-2">
+                <RadioGroupItem
+                  value="address"
+                  id="address"
+                  className="shrink-0"
+                />
+                <div className="grid gap-1.5">
+                  <Label
+                    htmlFor="address"
+                    className="text-sm font-medium cursor-pointer"
+                  >
+                    Single Address
+                  </Label>
+                  <p className="text-xs text-muted-foreground">
+                    Send to the same address each time
+                  </p>
+                </div>
+              </div>
+              <div className="flex items-start space-x-2">
+                <RadioGroupItem value="xpub" id="xpub" className="shrink-0" />
+                <div className="grid gap-1.5">
+                  <Label
+                    htmlFor="xpub"
+                    className="text-sm font-medium cursor-pointer"
+                  >
+                    XPUB
+                  </Label>
+                  <p className="text-xs text-muted-foreground">
+                    Generate new addresses from extended public key
+                  </p>
+                </div>
+              </div>
+            </RadioGroup>
+          </div>
+          <div className="grid gap-1.5">
+            <Label>
+              {externalType === "address"
+                ? "Receiving on-chain address"
+                : "Extended Public Key (XPUB)"}
+            </Label>
+            <div className="flex gap-2">
+              <Input
+                placeholder={externalType === "address" ? "bc1..." : "xpub..."}
+                value={destination}
+                onChange={(e) => setDestination(e.target.value)}
+                required
+              />
+              <Button
+                type="button"
+                variant="outline"
+                className="px-2"
+                onClick={paste}
+              >
+                <ClipboardPasteIcon className="w-4 h-4" />
+              </Button>
+            </div>
+            <p className="text-xs text-muted-foreground">
+              {externalType === "address"
+                ? "Enter a Bitcoin address to receive swapped funds"
+                : "Enter an XPUB to automatically generate new addresses for each swap"}
+            </p>
           </div>
         </div>
       )}
--- a/lnclient/ldk/ldk.go
+++ b/lnclient/ldk/ldk.go
@@ -73,7 +73,6 @@ func NewLDKService(ctx context.Context, cfg config.Config, eventPublisher events
 	}
 
 	ldkConfig := ldk_node.DefaultConfig()
-	listeningAddresses := strings.Split(cfg.GetEnv().LDKListeningAddresses, ",")
 
 	ldkConfig.TrustedPeers0conf = []string{
 		lsp.OlympusLSP().Pubkey,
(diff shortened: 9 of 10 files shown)


frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+      setLoading(true);
+      if (isSwap) {
+        const swapOutResponse = await request<SwapResponse>("/api/swaps/out", {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+          },
+          body: JSON.stringify({
+            swapAmount: +amount,
+            destination: address,
```
im-adithya: Should we add the destination to swap description or payment metadata?

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+          Swap from Spending Balance
+        </Label>
+        <Switch id="swap" checked={isSwap} onCheckedChange={setSwap} />
+      </div>
+      <div className="grid gap-2 text-sm border-t pt-4">
+        <div className="flex items-center justify-between">
+          <p className="text-muted-foreground">On-chain Fee</p>
+          {swapFees ? (
+            <p>
+              ~{new Intl.NumberFormat().format(swapFees.boltzNetworkFee)} sats
```
im-adithya: Change this to reflect actual onchain fee when withdrawing from onchain balance

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+      <div className="flex gap-2">
+        <LinkButton to="/wallet/send" variant="outline">
+          Back
+        </LinkButton>
+        <LoadingButton
+          loading={isLoading}
+          type="submit"
+          className="w-full md:w"
+        >
+          Send
```
im-adithya, line 460: Add validation if amount is greater than spending balance or onchain balance

frontend/src/screens/wallet/send/LnurlPay.tsx
```diff
@@ -68,54 +87,76 @@ export default function LnurlPay() {
...
+      <div className="flex gap-2">
+        <LinkButton to="/wallet/send" variant="outline">
+          Back
+        </LinkButton>
+        <LoadingButton
+          loading={isLoading}
+          type="submit"
+          className="w-full md:w"
+        >
+          Send
```
im-adithya: Add validation checks here too

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+  if (!balances || !address || (!recommendedFees && !mempoolError)) {
+    return <Loading />;
+  }
+
+  return (
+    <form onSubmit={onSubmit} className="grid gap-6 md:max-w-lg">
+      <div className="grid gap-2">
+        <Label>Recipient</Label>
+        <div className="flex flex-wrap gap-2 items-center">
```
rolznz: image

This looks kind of weird at first impression. Not sure what we can improve there

frontend/src/screens/wallet/send/OnchainSuccess.tsx
```diff
@@ -0,0 +1,100 @@
...
+            className="w-full"
+          >
+            <ExternalLinkIcon className="w-4 h-4 mr-2" />
+            View on Mempool
+          </ExternalLinkButton>
+          <LinkButton to="/wallet/send" variant="outline" className="w-full">
+            <HandCoinsIcon className="w-4 h-4 mr-2" />
+            Make Another Payment
+          </LinkButton>
+          <LinkButton to="/wallet" variant="link" className="w-full">
```
rolznz: in the design there is no underline

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+        )}
+      </div>
+      <div className="flex gap-2">
+        <LinkButton to="/wallet/send" variant="outline">
+          Back
+        </LinkButton>
+        <LoadingButton
+          loading={isLoading}
+          type="submit"
+          className="w-full md:w"
```
rolznz: w-full should be removed to be consistent with the design

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+            <div className="flex justify-between text-muted-foreground text-xs sensitive slashed-zero">
+              <div>Minimum: 50000 sats</div>
+              <FormattedFiatAmount className="text-xs" amount={50000} />
+            </div>
+          )}
+        </div>
+      </div>
+      <div className="flex items-center justify-between">
+        <Label htmlFor="swap" className="font-medium text-sm cursor-pointer">
+          Swap from Spending Balance
```
rolznz, line 429: when this switch is off, I think it would be cool to show the spending balance below the switch

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+  const [feeRate, setFeeRate] = React.useState("");
+  const [isLoading, setLoading] = React.useState(false);
+
+  React.useEffect(() => {
+    if (recommendedFees?.fastestFee) {
+      setFeeRate(recommendedFees.fastestFee.toString());
+    }
+  }, [recommendedFees]);
+
+  const onSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
```
rolznz, line 134: can we have different submit functions for the different buttons?

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+      }
+      setLoading(true);
+      if (isSwap) {
+        const swapOutResponse = await request<SwapResponse>("/api/swaps/out", {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+          },
+          body: JSON.stringify({
+            swapAmount: +amount,
```
rolznz: I think we have a problem here: I wanted to send 100K sats but I sent 98K sats. I think in this case the fee should be applied on top

rolznz: If you pay for a product or service this won't work at all because you have to send **exactly** the correct amount to the destination address

frontend/src/screens/wallet/send/LnurlPay.tsx
```diff
@@ -68,54 +87,76 @@ export default function LnurlPay() {
...
+          }
+        />
+        <div className="flex justify-between text-muted-foreground text-xs sensitive slashed-zero">
           <div>
-            <Label>Description</Label>
-            <p className="text-muted-foreground">
-              {lnAddress.lnurlpData.description}
-            </p>
+            Spending Balance:{" "}
+            {new Intl.NumberFormat().format(
```
rolznz: We need to floor the value everywhere

And the pay invoice screen now doesn't show your balance and kind of looks broken.

image

rolznz: Same for paying 0-amount invoices, we need a change there

frontend/src/components/layouts/SendLayout.tsx
```diff
@@ -45,19 +42,7 @@ export default function SendLayout() {
...
       )}
-      <div className="flex gap-12 w-full">
-        <div className="w-full max-w-lg">
-          <Outlet />
-        </div>
-        <BalanceCard
-          balance={balances.lightning.totalSpendable}
-          title="Spending Balance"
-          buttonTitle="Top Up"
-          buttonLink="/channels/outgoing"
```
rolznz, line 54: We lose a feature by removing this card, are you sure we want to do this?

frontend/src/screens/wallet/Send.tsx
```diff
@@ -1,3 +1,4 @@
+import { validate } from "bitcoin-address-validation";
```
rolznz:
```suggestion
import { validate: validateBitcoinAddress } from "bitcoin-address-validation";
```

frontend/src/screens/wallet/send/LnurlPay.tsx
```diff
@@ -37,14 +39,31 @@ export default function LnurlPay() {
         ...(comment && { comment }),
         ...(identifier && { recipient_data: { identifier } }),
       };
-      navigate(`/wallet/send/confirm-payment`, {
-        state: {
-          args: {
-            paymentRequest: invoice,
+      const payInvoiceResponse = await request<PayInvoiceResponse>(
```
rolznz, line 47: We don't want to confirm the payment any more?

frontend/src/screens/wallet/send/Onchain.tsx
rolznz: Isn't it in the swap table entry?

rolznz: But on the transaction list it's missing, I think it would be good to easily see it there

frontend/src/screens/wallet/send/Onchain.tsx
rolznz: I saw the Framer one from @stackingsaunter which is two lines, but not sure why it's longer?

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+                  {word}
+                </span>
+              );
+            }
+          })}
+        </div>
+      </div>
+      <div className="grid gap-2">
+        <Label htmlFor="amount">Amount</Label>
+        <Input
```
rolznz: we could add max validation here based on the user's balance

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+          id="amount"
+          type="number"
+          value={amount}
+          placeholder="Amount in Satoshi..."
+          onChange={(e) => {
+            setAmount(e.target.value.trim());
+          }}
+          min={isSwap ? 50000 : 1}
```
rolznz:
```suggestion
          min={isSwap ? 50000 : 330}
```

rolznz: We should probably have a warning under 10K sats that the value could become unspendable when the fee rate rises and recommend to use lightning instead

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,234 @@
...
+          autoFocus
+          className="[appearance:textfield] [&::-webkit-outer-spin-button]:appearance-none [&::-webkit-inner-spin-button]:appearance-none"
+          endAdornment={
+            <FormattedFiatAmount amount={Number(amount)} className="mr-2" />
+          }
+        />
+        <div className="grid">
+          <div className="flex justify-between text-muted-foreground text-xs sensitive slashed-zero">
+            <div>
+              {isSwap ? "Spending Balance: " : "On-chain Balance: "}
```
rolznz: We have a lot of conditionals e.g. `isSwap`

Can we split the two into different components?

frontend/src/screens/wallet/Send.tsx
```diff
@@ -75,31 +85,33 @@ export default function Send() {
...
+    <div className="w-full md:max-w-lg">
+      <form onSubmit={onSubmit}>
+        <Label htmlFor="recipient">Recipient</Label>
+        <div className="flex gap-2 mb-4">
+          <Input
+            id="recipient"
+            type="text"
+            value={recipient}
+            autoFocus
+            placeholder="Invoice, lightning address, on-chain"
```
rolznz:
```suggestion
            placeholder="Invoice, lightning address, on-chain address"
```

frontend/src/screens/wallet/send/OnchainSuccess.tsx
```diff
@@ -0,0 +1,100 @@
...
+            <p className="text-2xl font-medium slashed-zero">
+              {new Intl.NumberFormat().format(amount)} sats
+            </p>
+            <FormattedFiatAmount amount={amount} className="text-xl" />
+          </div>
+          {!mempoolTx?.status.confirmed && (
+            <p className="text-muted-foreground text-center">
+              You can now leave this page.
+              <br />
+              You'll be notified when transaction arrives.
```
rolznz: how?

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: I think he just used a random example, I added a comment in Figma

frontend/src/screens/wallet/send/OnchainSuccess.tsx
im-adithya: Yeah I was thinking of changing the button component classes directly

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: And then it would disappear and only show under the input on toggling? I like how it is now better

frontend/src/screens/wallet/send/OnchainSuccess.tsx
rolznz: Can you check how we will do this after @reneaaron 's redesign? I guess we don't edit shadcn classes directly any more?

frontend/src/screens/wallet/send/Onchain.tsx
rolznz: hmm. The problem I see right now we can't see the balance until we toggle the switch.

Or could we add it to "Swap from Spending Balance" e.g. "Swap from Spending Balance (300k sats)"

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: Can we say `Swapped out to <BTC Address>`?

rolznz: Maybe even `Swap to <BTC Address>`? since it should be obvious in this case

frontend/src/screens/wallet/send/OnchainSuccess.tsx
im-adithya: Ah okay, I think [shadcn button](https://ui.shadcn.com/docs/components/button#examples) itself has changed, it now starts without an underline and adds it on hover

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: Makes sense, but you mean only when done from Send screen? If we choose to display this for all swap outs then it would be good to think of a similar swap in message to align it (maybe a simple `Swap in` would suffice?)

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: I'm using `ONCHAIN_DUST_SATS` instead of `330` but it is set to `1000`, is that okay?

frontend/src/screens/wallet/send/LnurlPay.tsx
im-adithya: I think so, atleast from the design it looks this way

frontend/src/components/layouts/SendLayout.tsx
im-adithya: I think Jakub's reasoning is that we now show the balances under input fields

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: I did that, but I still feel what we have already is better:

https://github.com/user-attachments/assets/e3d18176-85cc-4602-b81f-188b9cd534f4

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: Oh so you mean two entirely different form components rendered like this: `isSwap ? <OnchainForm/> : <SwapForm/>`

If the current one looks good, I can make two functions and call them here conditionally

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: Fixed. So now, when a 50k sat swap is done from
- Send screen, received amount is 50k
- Swap screen (including auto), sent amount (from lightning) is 50k

frontend/src/screens/wallet/send/Onchain.tsx
rolznz: I guess that's ok. There should be no incentive to spend so little on-chain

frontend/src/components/layouts/SendLayout.tsx
rolznz: Ok. But we need to make sure it shows on every screen now as it did before.

frontend/src/screens/wallet/send/Onchain.tsx
rolznz: hmm. Yeah it is kind of ugly like this. @stackingsaunter @reneaaron any thoughts?

frontend/src/screens/wallet/send/Onchain.tsx
rolznz: Yeah maybe 2 components? because there are a lot of conditionals currently

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: I changed this a bit to make it simpler, can you have a look?

frontend/src/screens/wallet/send/ConfirmPayment.tsx
```diff
@@ -87,36 +94,72 @@ export default function ConfirmPayment() {
...
+            </div>
+            {invoice.description && (
+              <p className="text-lg text-muted-foreground">
+                {invoice.description}
+              </p>
+            )}
+          </CardContent>
+          <CardFooter className="flex flex-col gap-2 pt-2">
+            {exceedsBalance && (
+              <SpendingAlert className="mb-2" maxSpendable={maxSpendable} />
```
im-adithya: This shows the `nextMaxSpendableMPP` amount while the balance below the button is `totalSpendable`

frontend/src/screens/wallet/send/OnchainSuccess.tsx
im-adithya: Sorry I blindly followed the design, should we remove this?

rolznz: I don't think we will do that unless we send emails for all successful swaps, so I think we should remove it

rolznz:
@im-adithya 

> Check if PendingPaymentAlert is used enough, should we also keep it in ZeroAmount screen and Send screen (my reasoning to not include it in the initial screen is because we also have onchain now)

Yes, I think we need it to avoid the user paying twice by accident. This is again what I don't like that we removed the confirmation screen because we can consistently put all this stuff in one place... :confused: 

> I used different forms for Onchain and Swap screens, and that causes amount to reset on switching, is that okay?

I don't think that's good UX. We could use a state variable in the parent component and pass the getter and setter to the child forms

> I think fee rate while sending to Onchain addresses can be improved

Sounds good, I think we should allow setting a fee rate and default to highest fee. **both for swaps and normal on-chain payment**

It's weird right now we show a guessed sat amount for swaps, but we show a vByte amount for on-chain. I think we should use vByte for both.

>  What to do with Deposit and Withdraw pages in Node screen?

Why does anything need to be done? I think we should just leave them there, just like we do for swaps too.

> We have a warning which is based on NextMaxSpendableMPP in ConfirmPayment screen, but acc to the design we show the spending balance below the confirm button, how to proceed here?

Below in red it shows: Not Enough Spending Balance: 990,000 sats. I think the existing warning alert is better: (**Maximum Spendable Balance Too Low** Your payment will likely fail because your maximum spendable balance for the next payment is currently 970,200 sats.)

> Do we have to show the red text for spending balance with increase button while sending to onchain address too (via swap)?

Right now we don't show a warning at all and only have a max amount for form validation. But I think it would be nice to clearly show a warning like we already do with lightning

> How can we notify people of onchain transactions once confirmed?

We discussed this already, I don't think we can easily do this right now, let's not add it.

frontend/src/screens/wallet/send/ConfirmPayment.tsx
rolznz: Please add `<PendingPaymentAlert/>` and `<SpendingAlert/>` to all pages that can submit a lightning payment, otherwise we are losing functionality on those screens.

I think SpendingAlert internally should calculate `exceedsBalance` and only render if necessary. Then we can easily add this alert to the other screens.

> This shows the nextMaxSpendableMPP amount while the balance below the button is totalSpendable

I think this is OK. We have clear copy on the warning about the next spendable balance

api/models.go
```diff
@@ -147,6 +147,7 @@ type CreateLightningAddressRequest struct {
 type InitiateSwapRequest struct {
 	SwapAmount  uint64 `json:"swapAmount"`
 	Destination string `json:"destination"`
+	IsSending   bool   `json:"isSending"`
```
rolznz: This is weird naming in my opinion: `IsSending`.

Also `InitiateSwapRequest` should be `InitiateSwapOutRequest` right?

rolznz: Should `IsSending` be `UseExactReceiveAmount` or something like that?

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,360 @@
...
+          <p className="text-muted-foreground">On-chain Fee</p>
+          {feeRate ? <p>{feeRate} sat/vB</p> : <Loading className="w-4 h-4" />}
+        </div>
+      </div>
+      {amount && +amount < 10_000 && (
+        <Alert>
+          <InfoIcon className="h-4 w-4" />
+          <AlertTitle>Amount not ideal for On-chain transaction</AlertTitle>
+          <AlertDescription>
+            Small amounts can become unspendable during fee spikes. Consider
```
rolznz:
```suggestion
            Small amounts can become unspendable when mempool fees increase. Consider
```

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,360 @@
...
+            using Lightning instead.
+          </AlertDescription>
+        </Alert>
+      )}
+      <div className="flex gap-2">
```
rolznz, line 306: We need an alert about spending anchor reserves too. Can you make sure we have the same functionality as the existing redeem onchain funds page?

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,360 @@
...
+              )}{" "}
+              sats
+            </div>
+            <FormattedFiatAmount
+              className="text-xs"
+              amount={Math.floor(balances.lightning.totalSpendable / 1000)}
+            />
+          </div>
+          <div className="flex justify-between text-muted-foreground text-xs sensitive slashed-zero">
+            <div>Minimum: 50000 sats</div>
```
rolznz: I noticed this is hard coded but in Boltz right now the minimum is 25,000. And maybe if fees increase then the minimum would be more than 50,000. Can we fetch these limits from Boltz?

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -0,0 +1,360 @@
...
+                    </span>
+                  );
+                }
+              })}
+            </div>
+            <Link to="/wallet/send">
+              <XIcon className="w-4 h-4 cursor-pointer text-muted-foreground" />
+            </Link>
+          </div>
+        </div>
```
rolznz, line 81: We should show the existing mempool alert here (applies to both swaps and normal on-chain)

Please check if we do other stuff in the redeem onchain funds page too.

rolznz:
Regarding outgoing on-chain payments, they do not show on the wallet screen. Maybe rather than merging lightning and on-chain  transactions we could show an alert with a link to the node page on the transaction page, if you had an outgoing on-chain transaction within the last 24 hours?

image

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: Added ✅

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: Done ✅ 

And yes we already have this and anchor reserves alert in redeem onchain funds page

api/models.go
im-adithya: Changed to `UseExactReceiveAmount `

frontend/src/screens/wallet/send/Onchain.tsx
rolznz: Sorry, I meant **FROM** the redeem on-chain funds page. The functionality should be the same here

im-adithya:
Ready for re-review!

frontend/src/screens/wallet/send/Onchain.tsx
stackingsaunter: You don't need to see onchain balance when you swap from lightning right? So it should be where the Amount is.

Maybe the switch should be above
CleanShot 2025-08-21 at 15 12 57@2x

CleanShot 2025-08-21 at 15 13 18@2x

@rolznz @reneaaron @im-adithya

Issue #1534, opened 2025-07-23
Add a link to a guide on failed payment / transaction detail modal of failed payments


rolznz:
The guide can give a few suggestions why payments might fail and how to fix them

Aditya-PS-05:
@rolznz , can I work on this issue?

PR #1593, opened 2025-08-11
feat: payment failed alert

rolznz:
Closes https://github.com/getAlby/hub/issues/1534

We need to review payment sending UI in https://github.com/getAlby/hub/pull/1582 - Ideally we don't need to duplicate this on many pages (update: it's on 3 pages).

When a payment fails (or when you view a failed payment in the transaction list) we give a link to a guide and also a button to send details about the failed payment to Alby, so we can review it.

image

image


im-adithya (approved):
tACK

Issue #1548, opened 2025-07-29
Implement new Connections page design


rolznz:
App store and connections merged

See https://www.figma.com/design/W8IlNozcOeYdCun6A7Qk8C/%F0%9F%86%95-Alby-Hub-2.0?node-id=69-511749&t=KljWplEHQ1h6xZP3-0

rolznz:
Hi @Aditya-PS-05 this is a bigger task, an Alby team member will work on it.

Aditya-PS-05:
I can work and create a pr and if you disliked it, you can close it.

PR #1596, opened 2025-08-12
feat: merge app store and connections page

rolznz:
Closes https://github.com/getAlby/hub/issues/1548

TODOs:
- [x] review component inconsistencies with design (tabs, badges) ok for now
- [x] do we need to merge app detail with the app store page (I think it adds complexity) - **added** but editing is still done on the show app page
- [x] implement improved app detail page
- [x] Re-add the links card (even though you're already connected to the app you might want to visit the site)
- [x] How to remove the duplication between the app store: and the connection detail:
  - [x] On the app store, if you click on an app that is already installed, go to the app detail of the first connection for that app.
  - [x] If there are multiple connections show a dropdown on the detail page (with the currently selected connection)
- [x] On app detail page, hide almost everything from the connection details in advanced (id, pubkeys etc)
- [x] Payment tools and merchant tools should be moved below social media
- [x] For isolated apps, add a "Balance" card next to total spent, total received, which can have the buttons for top up and draw down
- [x] For subwallets, add a lightning address card
- [x] Change top up / draw down to increase/decrease
- [x] Merge with tw fixes
  - [x] Does dropdown selection work for entire item?
- [x] Remove unused code in app detail header

### DIFF
--- a/frontend/src/components/connections/DisconnectApp.tsx
+++ b/frontend/src/components/connections/DisconnectApp.tsx
@@ -16,7 +16,13 @@ import { SUBWALLET_APPSTORE_APP_ID } from "src/constants";
 import { useDeleteApp } from "src/hooks/useDeleteApp";
 import { App } from "src/types";
 
-export function DisconnectApp({ app }: { app: App }) {
+export function DisconnectApp({
+  app,
+  onClose,
+}: {
+  app: App;
+  onClose: () => void;
+}) {
   const navigate = useNavigate();
 
   const { deleteApp, isDeleting } = useDeleteApp(() => {
@@ -28,7 +34,7 @@ export function DisconnectApp({ app }: { app: App }) {
   });
 
   return (
-    <AlertDialog>
+    <AlertDialog open>
       <AlertDialogTrigger asChild>
         <Button variant="outline">
           <UnplugIcon className="size-4" /> Disconnect
@@ -51,12 +57,12 @@ export function DisconnectApp({ app }: { app: App }) {
           </AlertDialogDescription>
         </AlertDialogHeader>
         <AlertDialogFooter>
-          <AlertDialogCancel>Cancel</AlertDialogCancel>
+          <AlertDialogCancel onClick={onClose}>Cancel</AlertDialogCancel>
           <AlertDialogAction
             onClick={() => deleteApp(app.appPubkey)}
             disabled={isDeleting}
           >
-            Continue
+            Confirm
           </AlertDialogAction>
         </AlertDialogFooter>
       </AlertDialogContent>
--- a/frontend/src/screens/apps/AppDetails.tsx
+++ b/frontend/src/screens/apps/AppDetails.tsx
@@ -19,6 +19,7 @@ import {
   PlusIcon,
   SquarePenIcon,
   SquareStackIcon,
+  UnplugIcon,
 } from "lucide-react";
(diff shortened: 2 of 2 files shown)


reneaaron:
Image

The app details page looks a bit odd with the connection card on top and the space to the right, doesn't it?

frontend/src/components/AppHeader.tsx
```diff
@@ -6,6 +6,7 @@ type Props = {
   title: string | ReactElement;
   description?: string | ReactElement;
   contentRight?: React.ReactNode;
+  contentRightOnNewLine?: boolean;
```
reneaaron: This isn't used?

frontend/src/components/connections/AppTransactionList.tsx
```diff
@@ -0,0 +1,24 @@
...
+  CardHeader,
+  CardTitle,
+} from "src/components/ui/card";
+
+export function AppTransactionList({ appId }: { appId: number }) {
+  return (
+    <Card>
+      <CardHeader>
+        <CardTitle>
+          <div className="flex flex-row justify-between items-center">
```
reneaaron: Do we really need this layout, should be a default <CardTitle>?

frontend/src/components/connections/ConnectedApps.tsx
```diff
@@ -54,35 +53,47 @@ function AppList() {
...
+                  />
+                </Link>
+              )}
+              {/* <Link to="/apps/new">
               <ResponsiveButton icon={CirclePlusIcon} text="Add Connection" />
-            </Link>
-          </>
-        }
-      />
+            </Link> */}
```
reneaaron:
```suggestion
```

frontend/src/components/Permissions.tsx
```diff
@@ -95,28 +98,34 @@ const Permissions: React.FC<PermissionsProps> = ({
         />
       ) : (
         <>
-          <p className="text-sm font-medium mb-2">This app can:</p>
-          <div className="flex flex-col mb-2">
+          {/* {permissions.isolated && (
```
reneaaron: This isn't needed anymore?

frontend/src/screens/internal-apps/AlbyGo.tsx
```diff
@@ -222,50 +217,50 @@ export function AlbyGo() {
           </Card>
         </div>
         <div className="flex flex-col w-full gap-6">
-          {(app.appleLink ||
-            app.playLink ||
-            app.zapStoreLink ||
-            app.chromeLink ||
-            app.firefoxLink) && (
+          {(appStoreApp.appleLink ||
```
reneaaron: Could we use the AppLinksCard here?

reneaaron (commented):
Overall it feels really nice, great job! 💯 Just some minor feedback:

1) The BuzzPay app looks a bit odd currently: 
Image

2) I still think the "disconnect" button is way too prominent and could be moved into the dropdown

3) Could we make the permission details (and collapse them by default)? What do you think? 

Image

frontend/src/components/Permissions.tsx
rolznz: It's not needed any more, now we have a card above titled "Isolated Balance"

rolznz:
@reneaaron good catch - buzzpay now looks like this:

image

Not that great but maybe ok for a first iteration.

PR #1585, opened 2025-08-07
feat: add swap in and onchain deposit to receive flow

im-adithya:
## Description

This PR improves Receive screen to match the Alby Hub 2.0 design.

⚠️ WIP PR

PR #1598, opened 2025-08-13
fix: use actual swap limits from boltz api

im-adithya:
This PR adds fetching swap min and max amounts from the boltz api instead of hardcoding to `50_000` sats

Addresses this [comment](https://github.com/getAlby/hub/pull/1585#discussion_r2272384741)

### DIFF
--- a/swaps/swaps_service.go
+++ b/swaps/swaps_service.go
@@ -88,6 +88,8 @@ type SwapFees struct {
 	AlbyServiceFee  float64 `json:"albyServiceFee"`
 	BoltzServiceFee float64 `json:"boltzServiceFee"`
 	BoltzNetworkFee uint64  `json:"boltzNetworkFee"`
+	MinAmount       uint64  `json:"minAmount"`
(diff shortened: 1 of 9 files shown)


swaps/swaps_service.go
```diff
@@ -517,11 +519,14 @@ func (svc *swapsService) CalculateSwapOutFee() (*SwapFees, error) {
 
 	fees := pairInfo.Fees
 	networkFee := fees.MinerFees.Lockup + fees.MinerFees.Claim
+	limits := pairInfo.Limits
 
 	return &SwapFees{
 		AlbyServiceFee:  AlbySwapServiceFee,
 		BoltzServiceFee: fees.Percentage,
 		BoltzNetworkFee: networkFee,
+		MinAmount:       limits.Minimal,
```
rolznz, line 528: it's weird we put these in SwapFees. I know that's to not have to do another API call but these are not fees right? they are limits as named above

PR #1601, opened 2025-08-14
fix: ui fixes after upgrading to tw4

reneaaron:
Follow-up PR to fix any inconsistencies after the tailwind 4 upgrade.


im-adithya:
> Also your update banner is too tall

Ah no, that's my macbook's bezel which comes in every screenshot

rolznz:
@reneaaron do you know why there are so many conflicts on this PR? weird

reneaaron:
There were just a lot of import conflicts as I've updated the toasts to use sonner (instead of the deprecated shadcn toaster).

PR #1629, opened 2025-08-20
fix: set explicit timeouts of oauth client requests

rolznz:
I noticed a request can get stuck and potentially hang forever because the default timeout is **no timeout**.

### DIFF
--- a/alby/alby_oauth_service.go
+++ b/alby/alby_oauth_service.go
@@ -333,6 +333,7 @@ func (svc *albyOAuthService) GetVssAuthToken(ctx context.Context, nodeIdentifier
 	}
 
 	client := svc.oauthConf.Client(ctx, token)
+	client.Timeout = 10 * time.Second
 
 	type vssAuthTokenRequest struct {
 		Identifier string `json:"identifier"`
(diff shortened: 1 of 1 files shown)


alby/alby_oauth_service.go
```diff
@@ -333,6 +333,7 @@ func (svc *albyOAuthService) GetVssAuthToken(ctx context.Context, nodeIdentifier
 	}
 
 	client := svc.oauthConf.Client(ctx, token)
+	client.Timeout = 10 * time.Second
```
im-adithya, line 336: I think now is a good time to also change `NewRequest` to `NewRequestWithContext` everywhere instead of using `context.Background()`

```go
// NewRequest wraps [NewRequestWithContext] using [context.Background].
func NewRequest(method, url string, body io.Reader) (*Request, error) {
	return NewRequestWithContext(context.Background(), method, url, body)
}
```

im-adithya (approved):
utACK, left a minor suggestion

alby/alby_oauth_service.go
rolznz: Nice! I will add a separate issue for this. I guess it should be applied everywhere, not just when the OAuth client is used

PR #1632, opened 2025-08-21
app rename improvements

matjaz:


frontend/src/screens/apps/ShowApp.tsx
```diff
@@ -228,6 +228,14 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
                       value={name}
                       id="name"
                       onChange={(e) => setName(e.target.value)}
+                      onKeyDown={(e) => {
```
rolznz, line 231: I think we should use proper accessibility by wrapping the `Input` in a `form`, then we at least don't need the Enter handling. Not sure about escape too. Can't we tab to a cancel button?

PR #1636, opened 2025-08-21
chore: set user agent while connecting to relay

im-adithya:
Fixes https://github.com/getAlby/hub/issues/1611


rolznz (approved):
utACK,

Nice! this is a better way to do it since we don't dig into internal code. We should update http-nostr to do it this way too CC @frnandu

Issue #1531, opened 2025-07-22
Remove "contact alby support" for channel offline warning on Node page


rolznz:
See https://github.com/getAlby/hub/blob/master/frontend/src/components/channels/ChannelWarning.tsx#L22

With non-LSP channels we often cannot help much and we should recommend a user to have 2 channel partners for redundancy. Normally the channel will come online again by itself. This is already in the guide linked at the top of the page. So we just need to update this warning.

NodeDiver:
I'll take this task. I'll test the implementation first on regtest with polar locally and then test it on mutininet.

rolznz:
@NodeDiver no need to test on two different networks, it's a simple copy change in the UI

PR #1640, opened 2025-08-22
Fix/leftover work from 1531

NodeDiver:
This pr is an enhancement to the #1531 original pr: There is an external link for the user to click and to "Learn more" about this issue, this should send the user to one of our FAQs in our guides that was specifically made to better inform the user what this offline channel is all about.

I know there was no particular issue about this. And there are quite a few changes so that the logic for adding a link could be added. I hope everything is as it should be.

image


NodeDiver:
I need to review this more.

PR #1650, opened 2025-08-26
feat: stories

reneaaron:
- Provide helpful videos / guides in context (possibly could also be added on other pages, need to see how that feels from a user perspective)
 - Improve awareness of new features (such as auto-swaps :100: )


frontend/src/screens/Home.tsx
```diff
@@ -51,16 +64,85 @@ function Home() {
   const { data: balances } = useBalances();
   const { data: albyMe } = useAlbyMe();
   const [isNerd, setNerd] = React.useState(false);
+  const [isStoriesOpen, setIsStoriesOpen] = React.useState(true);
   /* eslint-disable  @typescript-eslint/no-explicit-any */
   const extensionInstalled = (window as any).alby !== undefined;
 
+  // State for dialog
```
rolznz: Could all this code be moved to the stories component?

frontend/src/components/ui/custom/stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+        className={cn(
+          "rounded-full border-2 flex items-center justify-center cursor-pointer p-0.5",
+          story.seen
+            ? "bg-muted" // Seen state
+            : "border-primary" // Unseen state with animated primary color border
+        )}
+        onClick={handleClick}
+      >
+        <Avatar className={cn("border-background size-14")}>
+          {typeof story.avatar === "string" ? (
```
rolznz: It would be good to just be consistent here and always use image URLs?

frontend/src/screens/Home.tsx
reneaaron: Yes, definitely.

frontend/src/components/ui/custom/stories.tsx
reneaaron: Sure, it's currently in the state of "vibe-coded hack" :smile:

PR #1660, opened 2025-08-28
fix: remove Alby Cloud sidebar link

rolznz:
image

Later this could be "Upgrade" and also recommend Alby Cloud?


rolznz:
Update: there's already a CTA in the dropdown. I removed this one.

Issue #1680, opened 2025-09-03
Slow query to fetch transactions


rolznz:
LOG:  duration: 181.738 ms  execute SELECT * FROM "transactions" WHERE state = $1 OR type = $2 ORDER BY updated_at desc LIMIT $3 OFFSET $4

(The main `ListTransactions` method in transactions service, probably missing an index)

rolznz:
```
SELECT 
    indexname, 
    indexdef 
FROM 
    pg_indexes 
WHERE 
    tablename = 'transactions';
                      indexname                      |                                                         
         indexdef                                                                  
-----------------------------------------------------+---------------------------------------------------------
-----------------------------------------------------------------------------------
 transactions_pkey                                   | CREATE UNIQUE INDEX transactions_pkey ON public.transact
ions USING btree (id)
 idx_transactions_app_id                             | CREATE INDEX idx_transactions_app_id ON public.transacti
ons USING btree (app_id)
 idx_transactions_state                              | CREATE INDEX idx_transactions_state ON public.transactio
ns USING btree (state)
 idx_transactions_type                               | CREATE INDEX idx_transactions_type ON public.transaction
s USING btree (type)
 idx_transactions_payment_hash                       | CREATE INDEX idx_transactions_payment_hash ON public.tra
nsactions USING btree (payment_hash)
 idx_transactions_app_id_type_state_created_at       | CREATE INDEX idx_transactions_app_id_type_state_created_
at ON public.transactions USING btree (app_id, type, state, created_at)
 idx_transactions_app_id_type_state                  | CREATE INDEX idx_transactions_app_id_type_state ON publi
c.transactions USING btree (app_id, type, state)
 idx_transactions_state_type                         | CREATE INDEX idx_transactions_state_type ON public.trans
actions USING btree (state, type)
 idx_transactions_state_type_updated_at              | CREATE INDEX idx_transactions_state_type_updated_at ON p
ublic.transactions USING btree (state, type, updated_at)
 idx_transactions_app_id_state_type_updated_at       | CREATE INDEX idx_transactions_app_id_state_type_updated_
at ON public.transactions USING btree (app_id, state, type, updated_at)
 idx_transactions_payment_hash_settled_at_created_at | CREATE INDEX idx_transactions_payment_hash_settled_at_cr
eated_at ON public.transactions USING btree (payment_hash, settled_at, created_at)
(11 rows)
```

```
[...]

PR #1681, opened 2025-09-03
chore: speed up slow list transactions query by adding updated at index

rolznz:
Fixes https://github.com/getAlby/hub/issues/1680 

(10000x speedup)


im-adithya (approved):
tACK

Issue #1644, opened 2025-08-25
New Connection Wizard


rolznz:
The current design is we give instructions on the app store page. This is not helpful as it does not guide the user in real time, also there is no guide if you have already connected once.

The wizard can allow the user to go step by step to show exactly what they should do and where.

It also might help us to solve some of the inconsistencies between internal apps and standard apps, for example Alby Go. Instead of being an internal app just so we can add the 1-tap connections option (and the issue here is we have two different connect CTAs on that page), it can be a standard app with a custom app connection Wizard. Maybe this can apply to our other internal apps too.

Follow up of https://github.com/getAlby/hub/issues/1548

This should help us remove some components (like AppStoreDetailHeader which has duplicated code).

https://www.figma.com/design/W8IlNozcOeYdCun6A7Qk8C/%F0%9F%86%95-Alby-Hub-2.0?node-id=386-49687&t=cWNI6TtRaj6vAENF-0

PR #1690, opened 2025-09-05
feat: new connection wizard

rolznz:
Closes https://github.com/getAlby/hub/issues/1644

- [x] add dynamic stepper component
- [x] move connection created into new app flow
- [x] fix alby go app (superuser should go in new app flow, convert alby go back to a non-custom app)
- [x] update guides for all suggested apps
- [x] simpler scope group picker
- [x] test all flows
  - [x] 1-click connection flow with redirect
  - [x] 1-click connection flow without redirect
  - [x] 1-click connection flow with app pubkey set
  - [x] 1-click connection flow with app store id in metadata
  - [x] normal flow - app store app
  - [x] normal flow - internal app (e.g. BuzzPay)
  - [x] normal flow - custom app
- [x] test mobile UX


im-adithya:
Should we use `<ResponsiveButton/>` for the connect button?

<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

Apart from that the mobile UX looks :ok_hand:

frontend/src/components/connections/SuggestedAppData.tsx
```diff
@@ -166,6 +166,43 @@ export const appStoreApps: AppStoreApp[] = (
       logo: buzzpay,
       categories: ["merchant-tools"],
       webLink: "https://pos.albylabs.com",
+      installGuide: (
```
rolznz: This shouldn't be needed, it's an internal app

PR #1692, opened 2025-09-05
feat: show BOLT11 invoice in tx modal

### LINKED ISSUE #1689: Include invoice in transaction modal details
Currently we only include the payment hash. It would be useful to include the invoice too.

im-adithya:
Fixes #1689 

This makes it too long though, I'm wondering if we should restrict the size because there is the copy button if people want it in full.


rolznz:
@im-adithya I think it's good, normally it's hidden behind the "Details" toggle anyway

rolznz (approved):
utACK

PR #1698, opened 2025-09-07
Add app: Zaposhi

FedericoLuque:
Add app: Zaposhi


FedericoLuque:
Thanks for your reply. I'll check my Cloudflare settings. In the meantime, you can access the website through Tor, which is functional.

http://hxqsg6kbkqltsfvf2natvac3i7ezt5rvn2nz3z6527bwovk3ol5r5oyd.onion

rolznz:
Thanks @FedericoLuque 

We actually have another PR which is changing how our suggested apps work. So we can review and merge this in after that.

Just one thing, could you optimize your image - make sure the resolution is not too high (max 512x512px) and then put it through https://tinypng.com/? ideally it is < 20kb.

PR #1708, opened 2025-09-08
feat: add support for cron tabs in zapplanner

### LINKED ISSUE #1643: Add cron support for ZapPlanner payments
This was added to ZapPlanner but is not in the hub UI yet. Also can be used for the subscription payments from non-logged-in contributors

im-adithya:
Fixes #1643

The default will make the payment every month (starting the next midnight, so if you start on 8th 9:45 am, it will make the payment on 9th 00:00, and 9th of the next month and so on). Otherwise you can select advanced options and set a cron expression. The runs are calculated for that month and budget is decided accordingly.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -57,6 +57,7 @@
     "clsx": "^2.1.1",
     "cmdk": "^1.1.1",
     "compare-versions": "^6.1.1",
+    "cron-parser": "^5.3.1",
     "date-fns": "^4.1.0",
     "dayjs": "^1.11.10",
     "embla-carousel-react": "^8.6.0",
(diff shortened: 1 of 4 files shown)


rolznz:
@im-adithya I think this is a bit complicated and now the user is exposed to cron which is a more advanced setting.

What I meant was to not change the current UI but instead, if months is chosen then only allow the frequency to be 1 and instead of a sleep duration, use `0 0 1 * *` as the cron expression. This way we don't need to add the cron parser dependency and cron logic.

Secondly a big point of this is to also change the `SupportAlby` payments to use this cron value rather than 31 days. 31 days can skip entire months and putting it less can break the budgeting. With cron we exactly pay once per month.

rolznz:
@im-adithya I will try do an alternate PR

rolznz:
@im-adithya could you review this one? https://github.com/getAlby/hub/pull/1722

rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/1722

PR #1710, opened 2025-09-09
fix: add min width 0 to SidebarInset to stop overflows

### LINKED ISSUE #1694: Peers table does not fit on the screen
Image

im-adithya:
Fixes #1694

## Description

Flex children have `min-width:auto` set by default, which causes this behaviour, this change should fix all other screens too where the content overflows and adds a scrollbar. 

## Screenshot
Screenshot 2025-09-09 at 5 46 23 PM


rolznz:
I wonder the UX is still kind of bad on that screen because of the scrollbar on the table, but I guess it's not a commonly used one and not worth improving for mobile.

rolznz (approved):
tACK

PR #1714, opened 2025-09-10
feat: request channel info + orders from alby oauth endpoint

### LINKED ISSUE #1693: Construct URL for requesting manual incoming channel
Currently we use the URL from the Alby suggestions API.

Instead we will go to the Alby OAuth API (which will work for both logged in users and logged out users). The hub should construct the URL manually from the selected LSP name and the current network (signet/bitcoin).

This way logged in users can e.g. pay for second channel with card or the channel payment can be later deducted from their hub wallet.

im-adithya:
Fixes #1693 

:warning: Currently only `olympus/signet` endpoint has `allowPublicAccess` for testing, requires a change to OAuth2Server gateway_config to include all lsp endpoints for this to work.


alby/alby_oauth_service.go
```diff
@@ -1166,43 +1166,185 @@ func (svc *albyOAuthService) GetLSPChannelOffer(ctx context.Context) (*LSPChanne
...
+func (svc *albyOAuthService) GetLSPInfo(ctx context.Context, lsp, network string) (*LSPInfo, error) {
+	// TODO: Do not return if token does not exist
+	token, err := svc.fetchUserToken(ctx)
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to fetch user token")
+		return nil, err
+	}
+
+	client := svc.oauthConf.Client(ctx, token)
+	client.Timeout = 10 * time.Second
```
bumi: I think for the LSP stuff we should have 30 seconds

alby/alby_oauth_service.go
```diff
@@ -1166,43 +1166,185 @@ func (svc *albyOAuthService) GetLSPChannelOffer(ctx context.Context) (*LSPChanne
...
+
+	payloadBytes, err := json.Marshal(lspChannelRequest)
+	if err != nil {
+		return nil, err
+	}
+	bodyReader := bytes.NewReader(payloadBytes)
+
+	// TODO: Remove hardcoded lsp and network info
```
bumi: what's this todo?

alby/alby_oauth_service.go
```diff
@@ -1166,43 +1166,185 @@ func (svc *albyOAuthService) GetLSPChannelOffer(ctx context.Context) (*LSPChanne
...
+func (svc *albyOAuthService) CreateLSPOrder(ctx context.Context, lsp, network string, lspChannelRequest *LSPChannelRequest) (*LSPChannelResponse, error) {
+	// TODO: Do not return if token does not exist
+	token, err := svc.fetchUserToken(ctx)
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to fetch user token")
+		return nil, err
+	}
+
+	client := svc.oauthConf.Client(ctx, token)
+	client.Timeout = 10 * time.Second
```
bumi: also here we should increase to at least 30 seconds.

alby/alby_oauth_service.go
```diff
@@ -1166,43 +1166,185 @@ func (svc *albyOAuthService) GetLSPChannelOffer(ctx context.Context) (*LSPChanne
...
 	}
 
 	logger.Logger.WithFields(logrus.Fields{
-		"pubkey": pubkey,
+		"pubkey": lspInfo.Pubkey,
 		"public": isPublic,
 	}).Info("Requesting auto channel")
 
-	autoChannelResponse, err := svc.requestAutoChannel(ctx, requestUrl+"/auto_channel", nodeInfo.Pubkey, isPublic)
+	autoChannelResponse, err := svc.requestAutoChannel(ctx, fmt.Sprintf("%s/internal/lsp/alby/%s/auto_channel", albyOAuthAPIURL, nodeInfo.Network), nodeInfo.Pubkey, isPublic)
```
bumi, line 1354: what changed here actually?

alby/alby_oauth_service.go
im-adithya: Changed

alby/alby_oauth_service.go
im-adithya: Changed

alby/alby_oauth_service.go
im-adithya: Fixed now!

alby/alby_oauth_service.go
im-adithya: I reused the code from `GetLSPInfo` above instead of calling the `requestUrl+"/get_info", so I had to construct the URL directly here for order creation.

Actually can we align `alby/auto_channel` endpoint with `alby/create_order` to reduce code here? We pass in just one backend type as an additional param in payload (compared to `lsp/create_order`)

bumi: @rolznz based on the `node_type` the backend decides if a zero-conf channel should be requested. 
can we align that with the `create_order` call?

rolznz: I don't understand this change yet, the auto_channel endpoint seems wrong now.

The auto_channel endpoint is only used for the first channel flow, @im-adithya did you intend to change this?

I think it's good if we could remove duplication now that both are basically handled the same on the backend? (remove the auto_channel endpoint and use the lsp/alby one instead) but would it be better to do that as a follow-up PR?

rolznz: EDIT: Would it make more sense to always use the auto_channel endpoint and remove the others? (and add any missing fields to the auto_channel endpoint e.g. an optional LSP name to configure it?) then we don't need to multiple endpoints in the OAuth gateway.

rolznz: Update: since we want to be more consistent with the LSP spec then I guess we should remove the auto_channel request and update the first channel flow?

alby/alby_oauth_service.go
```diff
@@ -1166,43 +1166,196 @@ func (svc *albyOAuthService) GetLSPChannelOffer(ctx context.Context) (*LSPChanne
...
+	token, err := svc.fetchUserToken(ctx)
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to fetch user token")
+		return nil, err
+	}
+
+	var client *http.Client
+	if token != nil {
+		client = svc.oauthConf.Client(ctx, token)
+		client.Timeout = 30 * time.Second
```
rolznz: `client.Timeout` could be set under the conditional, and then we don't need to set it twice

alby/alby_oauth_service.go
```diff
@@ -1166,43 +1166,196 @@ func (svc *albyOAuthService) GetLSPChannelOffer(ctx context.Context) (*LSPChanne
 	return lspChannelOffer, nil
 }
 
+func (svc *albyOAuthService) GetLSPInfo(ctx context.Context, lsp, network string) (*LSPInfo, error) {
```
rolznz: here we call it `lsp` but other places we call it `lspIdentifier`, should we call it `lspIdentifier` here?

alby/alby_oauth_service.go
```diff
@@ -1166,43 +1166,196 @@ func (svc *albyOAuthService) GetLSPChannelOffer(ctx context.Context) (*LSPChanne
...
+	if err != nil {
+		logger.Logger.WithError(err).Error("Failed to read response body")
+		return nil, errors.New("failed to read response body")
+	}
+
+	if res.StatusCode >= 300 {
+		logger.Logger.WithFields(logrus.Fields{
+			"body":        string(body),
+			"status_code": res.StatusCode,
+		}).Error("users endpoint returned non-success code")
```
rolznz: Here and below it should not be `users`

alby/alby_oauth_service.go
```diff
@@ -1166,43 +1166,196 @@ func (svc *albyOAuthService) GetLSPChannelOffer(ctx context.Context) (*LSPChanne
...
+		return nil, fmt.Errorf("users endpoint returned non-success code: %s", string(body))
```
rolznz: same here, should not be `users`

api/lsp.go
```diff
@@ -73,7 +57,7 @@ func (api *api) RequestLSPOrder(ctx context.Context, request *LSPOrderRequest) (
 		return nil, err
 	}
 
-	invoice, fee, err := api.requestLSPS1Invoice(ctx, request, nodeInfo.Pubkey, lspInfo.MaxChannelExpiryBlocks, lspInfo.MinRequiredChannelConfirmations, lspInfo.MinFundingConfirmsWithinBlocks)
+	invoice, fee, err := api.requestLSPS1Invoice(ctx, request, nodeInfo.Network, nodeInfo.Pubkey, lspInfo.MaxChannelExpiryBlocks, lspInfo.MinRequiredChannelConfirmations, lspInfo.MinFundingConfirmsWithinBlocks)
```
rolznz, line 60: I think I missed this in the last PR. The invoice in the response should be considered optional (if it doesn't exist then the Alby Account handles the payment). I will update it

rolznz (approved):
tACK (except user with BTC currency)

PR #1717, opened 2025-09-15
feat: read only JWT for http api use

rolznz:
This splits up the http endpoints into two groups (read only and full access). In the developer settings you can create either a full-access token or a readonly one.

This allows less-trusted services to only have read-only access to the hub, e.g. for collecting metrics, healthcheck etc. 

image

Tests:
- [x] start
- [x] unlock
- [x] change unlock password
- [x] use legacy token without permission claim
- [x] use readonly token to access readonly endpoint - works
- [x] use readonly token to access full permission endpoint - fails
- [x] use full access token to access full permission endpoint - works

TODO:
- [x] automated test

### DIFF
--- a/http/http_service_test.go
+++ b/http/http_service_test.go
@@ -28,6 +28,7 @@ func TestUnlock_IncorrectPassword(t *testing.T) {
 	mockSvc := mocks.NewMockService(t)
 	gormDb, err := db.NewDB(t)
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


im-adithya (commented):
tACK

bumi:
seems a bit GET requests are read and other methods require full access. 
nice restfulness!

Issue #1723, opened 2025-09-16
Bug: "payment not found" logged after payment succeeded due to payment marked as failed beforehand


rolznz:
LDK payment was marked as failed due to context cancelled and then later we received an LDK successful payment event.

Since there was only a failed payment, not a pending one, the failed payment was not updated.

This was introduced in https://github.com/getAlby/hub/pull/1589

Issue #1724, opened 2025-09-16
Use a background context for making LNClient payments


rolznz:
I noticed when sending a payment the context is passed all the way from the HTTP request (for payments initiated from the hub) and from the subscription context (for NWC). I wonder if we should use a background context when calling the LNClient? otherwise the payment logic can be aborted part-way through and incorrectly mark the payment as failed even though it's still pending (although it should still resolve for LDK/LND because we have async event handlers that will notify of the payment success event - but I think it's still not great that it is temporarily in the wrong status)

PR #1727, opened 2025-09-16
fix: mark failed payment as settled

rolznz:
Fixes https://github.com/getAlby/hub/issues/1723

Now if we mark a payment as failed (ideally this never happens if it is actually pending, see https://github.com/getAlby/hub/issues/1724 as an example for how this can happen) it can still be marked as successful later.

But we ensure that we always check for a pending payment to update first, and only if no pending payment exists then we update the latest failed payment for the same payment hash. This ensures we do not have stuck pending payments.

### DIFF
--- a/transactions/notifications_test.go
+++ b/transactions/notifications_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"testing"
+	"time"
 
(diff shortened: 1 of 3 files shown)


im-adithya (approved):
utACK

Issue #1712, opened 2025-09-10
LND specify correct route hint for rebalancing


rolznz:
Currently LND rebalancing is very likely to fail because unlike LDK, only one route hint is included so it’s likely to be the wrong one.

With LND (and hopefully one day LDK) we can specify the route hint. Via the LNClient MakeInvoice method maybe there could be an optional throughPubkey argument?

PR #1731, opened 2025-09-17
feat: rebalance through node pubkey

rolznz:
Fixes https://github.com/getAlby/hub/issues/1712

For some reason even when providing a single route hint, LND is generating 2 route hints in the invoice (and the original issue was that the LND invoice was only including one route hint which was causing rebalancing to not work)

However I opened another channel and tested rebalancing into all channels, and it always includes the correct route hint (+ another one, seems random). I opened a 4th channel and also always includes the correct hint for the channel I want to rebalance into. So it seems to be working.

How to test: log the invoice created in the rebalance function and then return an error (instead of actually doing the rebalance, which will fail since we're running regtest in polar)

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -663,6 +663,13 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
 				remotePolicy = chanInfo.Node2Policy
 			}
 
+			if remotePolicy == nil {
+				logger.Logger.WithFields(logrus.Fields{
+					"channel_id": channel.ChanId,
+				}).WithError(err).Error("Remote channel policy does not exist")
+				continue
+			}
+
 			channelId := chanInfo.ChannelId
 			if channel.PeerScidAlias != 0 {
 				channelId = channel.PeerScidAlias
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -599,7 +599,7 @@ func (svc *LNDService) getPaymentResult(stream routerrpc.Router_SendPaymentV2Cli
 	}
 }
 
-func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, description string, descriptionHash string, expiry int64) (transaction *lnclient.Transaction, err error) {
+func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, description string, descriptionHash string, expiry int64, throughNodePubkey *string) (transaction *lnclient.Transaction, err error) {
 	var descriptionHashBytes []byte
 
 	if descriptionHash != "" {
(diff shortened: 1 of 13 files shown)


lnclient/lnd/lnd.go
```diff
@@ -628,6 +628,68 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
...
+			}
+
+			hint := &lnrpc.RouteHint{
+				HopHints: []*lnrpc.HopHint{
+					{
+						NodeId:                    channel.RemotePubkey,
+						ChanId:                    channelId,
+						FeeBaseMsat:               uint32(remotePolicy.FeeBaseMsat),
+						FeeProportionalMillionths: uint32(remotePolicy.FeeRateMilliMsat),
+						CltvExpiryDelta:           remotePolicy.TimeLockDelta,
```
im-adithya, line 685: We shouldn't access the policy directly as it can be `nil` and cause errors (refer to ListChannels)

lnclient/lnd/lnd.go
```diff
@@ -628,6 +628,68 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
...
+			PrivateOnly: true,
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		for _, channel := range channelsRes.Channels {
+			if channel.RemotePubkey != *throughNodePubkey {
+				continue
+			}
```
im-adithya, line 647: So wouldn't this check mean that we only add hints (max 3) from the same node? Is that productive?

lnclient/lnd/lnd.go
```diff
@@ -628,6 +628,68 @@ func (svc *LNDService) MakeInvoice(ctx context.Context, amount int64, descriptio
 	for _, channel := range channels {
 		if channel.Active && channel.Public {
 			hasPublicChannels = true
+			break
+		}
+	}
+
+	var hints []*lnrpc.RouteHint
+	if !hasPublicChannels && throughNodePubkey != nil {
```
im-adithya, line 636: So we don't require mentioning the route hint if the node is public? (for rebalancing that is, I'm not sure how that works)

lnclient/lnd/lnd.go
rolznz: We want to rebalance into a specific channel, but we are guessing by using the counterparty's node pubkey. I did this just in case the user has multiple channels with the same node, it will have a a higher success rate. But we recommend users against opening multiple channels to the same node.

lnclient/lnd/lnd.go
rolznz: Currently if the node has at least one public channel then we only expose the public channels (existing functionality to match how LDK works)

lnclient/lnd/lnd.go
rolznz: Thanks, fixed

lnclient/lnd/lnd.go
rolznz: to do it exactly properly we'd need to pass the channel ID into make_invoice rather than a pubkey. But I'm not sure it's worth it yet - that could be done as a followup if rebalancing is really used a lot

im-adithya: Yeah this looks okay for now, resolving :+1:

lnclient/lnd/lnd.go
im-adithya: But wouldn't that hinder if we are trying to rebalance into a private channel (in a node with public + private channels)? Not a big concern as this is an edge case.

im-adithya (approved):
utACK

lnclient/lnd/lnd.go
rolznz: Yeah, even before that we have the much more common usecase that if the user has public channels then they cannot receive at all on their private channels

Issue #1742, opened 2025-09-19
Bug: finalize step is incorrectly shown when using 1-click flow


rolznz:
When I click connect on the first step the app is created and the window is closed. There's no need for a second step because the user is already in the app and doesn't need to finalize the connection - it's done automatically. CC @im-adithya 

https://my.albyhub.com/apps/new?name=Zappy+Bird&pubkey=5d43a34ee12a7f8a391c8a18f579f43c9c067b51641672c173307e5ce797656f&budget_renewal=monthly&max_amount=10000000&request_methods=get_balance+pay_invoice&metadata=%7B%22app_store_app_id%22%3A%22zappy-bird%22%7D

<img width="812" height="1319" alt="Image" src="https://github.com/user-attachments/assets/4bf92101-9f53-4b16-8c4c-5c7cbc692e00" />

PR #1743, opened 2025-09-19
fix: remove finalize step in 1 click flow

im-adithya:
Fixes https://github.com/getAlby/hub/issues/1742

This also fixes a bug @rolznz discovered which happens because the event dispatch happens in the finalize component which is wrong. It is now shifted so that we dispatch the event right after the app creation.

### DIFF
--- a/frontend/src/screens/apps/NewApp.tsx
+++ b/frontend/src/screens/apps/NewApp.tsx
@@ -59,7 +59,6 @@ type NewAppInternalProps = {
 
 const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
   const location = useLocation();
-  const navigate = useNavigate();
 
   const [unsupportedError, setUnsupportedError] = useState<string>();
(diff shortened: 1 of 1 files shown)


frontend/src/screens/apps/NewApp.tsx
```diff
@@ -260,13 +261,41 @@ const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
...
+      }
+
       if (createAppResponse.returnTo) {
         // open connection URI directly in an app
         window.location.href = createAppResponse.returnTo;
         return;
       }
       toast("App created");
       setCreateAppResponse(createAppResponse);
+      if (pubkey) {
```
rolznz, line 294: I am not sure we should add this navigation. For most cases either there will be a return_to or a listener for one of the success messages above (like Bitcoin Connect, which will then close the dialog)

rolznz (approved):
tACK

Issue #1744, opened 2025-09-19
Word wrap/break missing on invoice description


frnandu:
<img width="1763" height="1096" alt="Image" src="https://github.com/user-attachments/assets/29b67ddd-d739-405f-9cf1-125738c11077" />

PR #1745, opened 2025-09-19
fix: add break all to description in confirm payment screen

im-adithya:
Fixes https://github.com/getAlby/hub/issues/1744


frontend/src/screens/wallet/send/ConfirmPayment.tsx
```diff
@@ -110,7 +110,7 @@ export default function ConfirmPayment() {
               />
             </div>
             {invoice.description && (
-              <p className="text-lg text-muted-foreground">
+              <p className="text-lg text-muted-foreground break-all">
```
rolznz: we should almost never have to use `break-all` 

Can you try `break-anywhere` instead?

PR #1747, opened 2025-09-19
fix: remove card description for Sub-Wallets

### LINKED ISSUE #1554: Minor copy improvement for sub-wallets
Image

What about we remove the subtitles of both tiles?

- "Total amount of assets under management" is somewhat redundant with the title: "Total Balance of Sub-wallets"
- "Number of Sub-wallets backed by your node funds" is not correct if sub-wallets actually exceed the Hub balance. Instead we can change "Active Sub-wallets" to "Number of Sub-wallets"

Dunsin-cyber:
Fixes #1554

Before
sub wallet before

After
sub-wallet after


rolznz (approved):
utACK

Issue #1695, opened 2025-09-06
Notify the user who wants to swap-in the max on-chain balance


MoritzKa:
At swap-in: A user who wants to send his max on-chain balance doesn't see the use on-chain funds button.

<img width="529" height="506" alt="Image" src="https://github.com/user-attachments/assets/6b8dcd37-ed88-4517-abdd-916616a297fa" />

Can we notify the user to keep some reserve in his on-chain balance? (I don't know what the right minimum amount is to keep in the on-chain balance)

krrish-sehgal:
I would like to work on this issue, and get to know more about the codebase through it.
Thanks!

krrish-sehgal:
This seems like it, I was digging into lightning.. I'll start with setting up a regtest and lightning, dev env(to get some "on-chain" funds) and then simulating the issue, and then add suggestion for UX enhancement!

PR #1748, opened 2025-09-20
fix(UX): Alert users about anchor-reserve when they try to use their max on-chain balance

krrish-sehgal:
Fixes https://github.com/getAlby/hub/issues/1695

https://github.com/user-attachments/assets/1e3a374b-464e-4112-9dad-878d187c2cb8

With reference to this [comment](https://github.com/getAlby/hub/issues/1695#issuecomment-3312101834), I found the Anchor reserve alert componet , whiich get triggered when user tries to use the onchain balance, that leaves less than 50k sats on-chain and not 25k , Let me know if that needs to be changed or this works. 

Thanks!

### DIFF
--- a/frontend/src/components/AnchorReserveAlert.tsx
+++ b/frontend/src/components/AnchorReserveAlert.tsx
@@ -6,8 +6,10 @@ import { useChannels } from "src/hooks/useChannels";
 export function AnchorReserveAlert({
   amount,
   className,
+  isSwap,
 }: {
   amount: number;
+  isSwap?: boolean;
   className?: string;
 }) {
   const { data: balances } = useBalances();
@@ -18,6 +20,7 @@ export function AnchorReserveAlert({
   }
 
   const showAlert =
+    amount &&
     !!channels.length &&
     +amount > balances.onchain.spendable - channels.length * 25000;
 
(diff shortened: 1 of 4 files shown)


rolznz:
@krrish-sehgal thanks for the PR, and cool you reused an existing alert. But in this case we do not show the button to allow the user to use their on-chain wallet if it would use anchor reserve funds.

I think if we remove the subtracted 25,000 at https://github.com/getAlby/hub/blob/master/frontend/src/screens/wallet/swap/SwapInStatus.tsx#L197 this will fix it.

And we should update the anchor reserve alert copy

From 
```
You have channels open and this withdrawal may deplete your anchor
        reserves which may make it harder to close channels without depositing
        additional onchain funds to your savings balance. To avoid this, set
        aside at least {new Intl.NumberFormat().format(channels.length * 25000)}{" "}
        sats on-chain.
```

To
```
        You have channels open and by spending your entire balance including your anchor
        reserves may put your node at risk of unable to reclaim funds in your channel after a force-closure. To avoid this, set
        aside at least {new Intl.NumberFormat().format(channels.length * 25000)}{" "}
        sats on-chain.
```

rolznz:
Edit: I wonder if we should use a different alert instead. Because then we can also say that the user must use an external wallet since they do not have enough balance in their hub's on-chain wallet (only applicable to swaps).

rolznz:
@krrish-sehgal testnet3 is basically unusable unfortunately. Boltz Regtest works well but I don't think we need to have it running for this task as we show the alert on the page beforehand.

I will change the copy a bit and also add this alert to the wallet receive (via onchain swap) page.

image

image

rolznz (approved):
tACK

krrish-sehgal:
This seems good.. but you mentioned that we needed to remove 25k from here "https://github.com/getAlby/hub/blob/master/frontend/src/screens/wallet/swap/SwapInStatus.tsx#L197 " so that button stays visible and users get a chance to spend all the funds even tho it depletes the anchor reserve.. .

Do we not want it anymore?

rolznz:
> This seems good.. but you mentioned that we needed to remove 25k from here "https://github.com/getAlby/hub/blob/master/frontend/src/screens/wallet/swap/SwapInStatus.tsx#L197 " so that button stays visible and users get a chance to spend all the funds even tho it depletes the anchor reserve.. .

> Do we not want it anymore?

True, I guess we should remove that, then it's consistent with how the on-chain withdrawal/send works.

rolznz:
One thing I don't like about the current implementation is we show this warning even if the user did not plan to use their hub wallet (they planned to use an external wallet). This warning will always show if they enter an amount bigger than their current on-chain balance. But if we instead only show the warning when the user clicks the button to transfer with the hub's on-chain funds, then the user has to abort the swap and go back to change the amount and they can't see how much they can actually safely send. So I guess it's ok. At least the user sees something which is better than now where the button just disappears without warning

krrish-sehgal:
@rolznz , this surely much better than navgating usrs to abort the swap and go back to square 1. Also , I have fixed the 25k condition. 

Thanks!

rolznz:
@krrish-sehgal thank you!

Issue #1746, opened 2025-09-19
Cant have NWC connection with only `pay_invoice`


v0l:
When using using custom permissions, `nwc` rust crate is unable to parse the response due to empty strings in the response, these are optional values and should be null or removed from the response.

`Can't deserialize response: response={"result":{"alias":"","color":"","pubkey":"","network":"","block_height":0,"block_hash":"","methods":["pay_invoice","pay_keysend","multi_pay_invoice","multi_pay_keysend","get_info","get_budget"],"notifications":[],"lud16":""},"result_type":"get_info"}, error=malformed public key`

<img width="492" height="242" alt="Image" src="https://github.com/user-attachments/assets/4d935faa-820c-4788-9605-de4d44b9812c" />

rolznz:
@v0l true, we should omit empty strings. I think the `nwc` rust crate should also be a bit more forgiving here though - it's not just Alby Hub that returns data like this I think

v0l:
Ok ill log an issue there too

rolznz:
@v0l which backend are you using? this should not be the case I think for LDK and LND.

krrish-sehgal:
Taking up this!

krrish-sehgal:
@v0l , needed some help replicating the issue, I tried to create a channel on regtest with (2lightning nodes) and created a payment, and got this 

```bash
 lncli1 addinvoice -amt 100
{
    "r_hash":  "4e6976648a9daafe8575d5d4e7bc7d69acfc2c927426e891eaf200b5d41df562",
    "payment_request":  "lnbcrt1u1p5vaw7spp5fe5hvey2nk40apt46h2w00radxk0ctyjwsnw3y027gqtt4qa743qdqqcqzzsxqyz5vqsp5nj5ch4z5xkh2mqs4mmzaq2eyvyc3tfwng43ypmf2qa5k0zegt9ss9qxpqysgqg9kg0ja72g8uf8g34jzz4ap95mmd5mf85tpapcnksu99dk87cw99j2guy9c30tctgm8hvmv78vwcta9fsjasqnyutn8exnn5qeqg0kqqrkxmwc",
    "add_index":  "1",
    "payment_addr":  "9ca98bd45435aead8215dec5d02b24613115a5d3456240ed2a0769678b285961"
}
```

[...]

PR #1750, opened 2025-09-20
fix(nip47): omit empty fields in get_info response for nwc rust crate

krrish-sehgal:
fixes: #1746

- Add omitempty JSON tags to getInfoResponse struct fields
- Prevent empty strings from breaking Rust serde deserialization  
- Fix NWC connections with limited permissions (pay_invoice only)
- Add comprehensive test for Rust client compatibility

Fixes issue where NWC Rust clients couldn't parse get_info responses
containing empty string values for optional node info fields when
apps had restricted permissions.

Before: 
Screenshot 2025-09-20 at 9 01 08 PM

After: 
Screenshot 2025-09-20 at 9 03 38 PM

### DIFF
--- a/nip47/controllers/get_info_controller.go
+++ b/nip47/controllers/get_info_controller.go
@@ -13,12 +13,12 @@ import (
 )
 
 type getInfoResponse struct {
-	Alias            *string     `json:"alias,omitempty"`
-	Color            *string     `json:"color,omitempty"`
-	Pubkey           *string     `json:"pubkey,omitempty"`
-	Network          *string     `json:"network,omitempty"`
-	BlockHeight      *uint32     `json:"block_height,omitempty"`
-	BlockHash        *string     `json:"block_hash,omitempty"`
+	Alias            *string     `json:"alias"`
+	Color            *string     `json:"color"`
+	Pubkey           *string     `json:"pubkey"`
+	Network          *string     `json:"network"`
+	BlockHeight      *uint32     `json:"block_height"`
+	BlockHash        *string     `json:"block_hash"`
 	Methods          []string    `json:"methods"`
 	Notifications    []string    `json:"notifications"`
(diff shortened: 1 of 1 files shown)


krrish-sehgal:
> @krrish-sehgal the change looks mostly good, but the test I think is a bit excessive and Alby Hub does not really need to know about the rust nwc project.
> 
> We already have a test for no permission here: https://github.com/getAlby/hub/blob/master/nip47/controllers/get_info_controller_test.go#L24
> 
> If we made those fields nullable (e.g. pointers) then in the test we could assert for nil instead of empty.

Sure, makes sense!
making the change right away.

nip47/controllers/get_info_controller.go
```diff
@@ -13,16 +13,16 @@ import (
...
+	Color            *string     `json:"color"`
+	Pubkey           *string     `json:"pubkey"`
+	Network          *string     `json:"network"`
+	BlockHeight      *uint32     `json:"block_height"`
+	BlockHash        *string     `json:"block_hash"`
 	Methods          []string    `json:"methods"`
 	Notifications    []string    `json:"notifications"`
 	Metadata         interface{} `json:"metadata,omitempty"`
-	LightningAddress string      `json:"lud16"`
+	LightningAddress string      `json:"lud16,omitempty"`
```
rolznz, line 25: should we make this a *string too?

rolznz: Edit: I will fix it in a follow-up PR

rolznz (approved):
tACK

nip47/controllers/get_info_controller.go
krrish-sehgal: on it!

rolznz: Fixed here: https://github.com/getAlby/hub/pull/1753

PR #1752, opened 2025-09-22
fix: removed boostagram logging

### LINKED ISSUE #1716: Don't log boostagram/comments
Seems we currently write the boostagram to the log. we should not log such data

krrish-sehgal:
fixes #1716


rolznz:
This only stops logging invalid boostagrams. I don't think it is the root issue, and also makes it harder to debug.

krrish-sehgal:
@rolznz , marking as draft , will reopen it , if you suggest I can help on this.

PR #1753, opened 2025-09-22
feat: include sub-wallet lightning address in get_info call

rolznz:
I realized we don't return the lightning address in the get_info call for subwallets.

### DIFF
--- a/nip47/controllers/get_info_controller.go
+++ b/nip47/controllers/get_info_controller.go
@@ -22,7 +22,7 @@ type getInfoResponse struct {
 	Methods          []string    `json:"methods"`
 	Notifications    []string    `json:"notifications"`
 	Metadata         interface{} `json:"metadata,omitempty"`
-	LightningAddress string      `json:"lud16,omitempty"`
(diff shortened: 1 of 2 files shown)


nip47/controllers/get_info_controller.go
```diff
@@ -90,7 +90,10 @@ func (controller *nip47Controller) HandleGetInfoEvent(ctx context.Context, nip47
 			}
 			if !app.Isolated {
 				lightningAddress, _ := controller.albyOAuthService.GetLightningAddress()
-				responsePayload.LightningAddress = lightningAddress
+				responsePayload.LightningAddress = &lightningAddress
+			} else if metadata["app_store_app_id"] == "uncle_jim" && metadata["lud16"] != nil {
```
im-adithya, line 94: Do we have to check for `metadata["app_store_app_id"] == "uncle_jim"`?

rolznz: @im-adithya yes, otherwise it's not considered a sub-wallet and it will not have a lightning address

Issue #1749, opened 2025-09-20
Failed sqlite-postgres migration: invalid input syntax for type json


rolznz:
It looks like some rare transactions with boostagrams can possibly fail to be migrated to postgres. I need to look deeper into the cause.

PR #1755, opened 2025-09-22
fix: ensure boostagrams are valid json before adding to the transaction

rolznz:
Closes https://github.com/getAlby/hub/issues/1749

If the boostagram is not valid it will break the transaction from being saved (postgres is strict that the field is actually json)

### DIFF
--- a/transactions/transactions_service.go
+++ b/transactions/transactions_service.go
@@ -442,7 +442,7 @@ func (svc *transactionsService) SendKeysend(amount uint64, destination string, c
 		logger.Logger.WithError(err).Error("Failed to serialize transaction metadata")
 		return nil, err
 	}
-	boostagramBytes := svc.getBoostagramFromCustomRecords(customRecords)
+	boostagramBytes := svc.getBoostagramBytesFromCustomRecords(customRecords)
 
 	var dbTransaction db.Transaction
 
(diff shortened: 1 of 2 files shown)


transactions/transactions_service.go
```diff
@@ -1106,13 +1106,22 @@ func makePreimageHex() ([]byte, error) {
 	return bytes, nil
 }
 
-func (svc *transactionsService) getBoostagramFromCustomRecords(customRecords []lnclient.TLVRecord) []byte {
+func (svc *transactionsService) getBoostagramBytesFromCustomRecords(customRecords []lnclient.TLVRecord) []byte {
 	for _, record := range customRecords {
 		if record.Type == BoostagramTlvType {
 			bytes, err := hex.DecodeString(record.Value)
 			if err != nil {
+				logger.Logger.WithError(err).Error("failed to decode boostagram tlv hex value")
```
bumi: do we need this log? 
or would it be good to log the input in this case, too?

bumi (approved):

im-adithya (approved):
utACK

transactions/transactions_service.go
rolznz: Oops, yeah I logged the input for both error messages in this PR

PR #1773, opened 2025-09-23
fix: fix button keyboard selection on calendar component

### LINKED ISSUE #1763: Expiry date picker calendar keyboard day select is not working
Using only the keyboard to try pick an expiry date doesn't work. When pressing space or enter on the intended date, it does not select it.

By using a standard `button` component instead of `Button` the issue is fixed (except that there is no styling on the buttons now). But I don't understand why.

https://github.com/getAlby/hub/blob/master/frontend/src/components/ui/calendar.tsx#L190

krrish-sehgal:
fixes #1763

https://github.com/user-attachments/assets/791b3eb1-7432-40b8-b1b8-7b7858801145

### DIFF
--- a/frontend/src/components/ui/calendar.tsx
+++ b/frontend/src/components/ui/calendar.tsx
@@ -187,10 +187,8 @@ function CalendarDayButton({
   }, [modifiers.focused]);
 
   return (
-    <Button
+    <button
       ref={ref}
-      variant="ghost"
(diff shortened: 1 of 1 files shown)


rolznz (approved):
tACK

This is a bit hacky because we should be able to use the `Button` component. But it works well, I think this is OK for now.

PR #1778, opened 2025-09-25
fix: patch app endpoints should allow accept partial fields 

### LINKED ISSUE #1546: Update PATCH app endpoint to allow providing only fields that need to be updated
Currently all properties have to be provided. For example if we only want to update the metadata, we should only pass the metadata.

krrish-sehgal:
fixes #1546

Uploading Screen Recording 2025-09-25 at 8.36.43 AM.mov…

### DIFF
--- a/frontend/src/screens/apps/AppDetails.tsx
+++ b/frontend/src/screens/apps/AppDetails.tsx
@@ -124,16 +124,59 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
   const [savedPermissions, setSavedPermissions] =
     React.useState<AppPermissions>(permissions);
 
+  // Helper to build minimal update request with only changed fields
+  const buildMinimalUpdateRequest = (): UpdateAppRequest => {
+    const updateRequest: UpdateAppRequest = {};
+
+    // Only include name if it changed
+    if (name !== app.name) {
+      updateRequest.name = name;
+    }
+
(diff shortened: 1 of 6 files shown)


frontend/src/screens/apps/AppDetails.tsx
```diff
@@ -124,16 +124,59 @@ function AppInternal({ app, refetchApp, capabilities }: AppInternalProps) {
...
     try {
-      const updateAppRequest: UpdateAppRequest = {
-        name,
-        scopes: Array.from(permissions.scopes),
-        budgetRenewal: permissions.budgetRenewal,
-        expiresAt: permissions.expiresAt?.toISOString(),
-        maxAmount: permissions.maxAmount,
-        isolated: permissions.isolated,
-      };
+      const updateAppRequest = buildMinimalUpdateRequest();
```
rolznz: I think we should undo the `buildMinimalUpdateRequest` change, you added a lot of complexity just so that we don't send a few fields, and I don't think there is any benefit.

rolznz:
@krrish-sehgal thanks for the PR! most looks good but this needs to be tested very thoroughly. I've tagged it for the next release milestone.

frontend/src/screens/apps/AppDetails.tsx
krrish-sehgal: fixed!

PR #1780, opened 2025-09-26
fix: add transaction fee percentage

reneaaron:
Adds the fee percentage to the transaction detail dialog: 

image


rolznz (approved):
utACK

PR #1781, opened 2025-09-26
feat: expose node URI address

frnandu:
Add node URI option to advanced menu.
If using LND will use directly the address retuned by backend, and when using LDK will expect a NODE_IPV6 env var set

### DIFF
--- a/lnclient/ldk/ldk.go
+++ b/lnclient/ldk/ldk.go
@@ -1012,11 +1012,39 @@ func (ls *LDKService) GetNodeConnectionInfo(ctx context.Context) (nodeConnection
 		return nil, err
 	}*/
 
-	return &lnclient.NodeConnectionInfo{
+	nodeConnectionInfo = &lnclient.NodeConnectionInfo{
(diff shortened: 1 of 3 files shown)


lnclient/ldk/ldk.go
```diff
@@ -1012,11 +1012,39 @@ func (ls *LDKService) GetNodeConnectionInfo(ctx context.Context) (nodeConnection
...
 
-	return &lnclient.NodeConnectionInfo{
+	nodeConnectionInfo = &lnclient.NodeConnectionInfo{
 		Pubkey: ls.node.NodeId(),
-		// Address: parts[0],
-		// Port:    port,
-	}, nil
+	}
+
+	// If NODE_IPV6 is set, include the connection address and port
```
rolznz: @frnandu I actually don't think we should do it this way,

the `LDKListeningAddresses` and `LDKAnnouncementAddresses` should be enough already and not require a new `NODE_IPV6` field

PR #1784, opened 2025-09-26
fix: fixed stack issue in header and lightning address section

### LINKED ISSUE #1782: Mobile display issues on connection page
Both the header and the lightning address buttons

Also we should show the app name - right now you can only access it by pressing edit.

Image

krrish-sehgal:
fixes #1782

https://github.com/user-attachments/assets/e585c327-81e9-4838-827e-7f12a84fdc05


rolznz:
I think the "Edit Connection" should be an icon only on mobile and "Connected" should be on the same row.

Can you also check with a long app name?

krrish-sehgal:
Screenshot 2025-09-28 at 3 51 18 PM

How bout something like this ?  For a long app name , Do you wanna move the name down ?

PR #1785, opened 2025-09-26
fix: added context sensitivbity to anchor reserver alert

### LINKED ISSUE #1783: Channel warning is too obvious and does not apply if user plans to use external wallet
This warning only applies if user plans to use their hub wallet, but on this page we don't know what the user will use. How can we improve this? maybe at least make the variant of the alert just standard?

Image

krrish-sehgal:
fixes #1783

My fix for the issue describes includes: 

1. Softened Alert Messaging & Variant
Changed variant: "warning" → "default" for less alarming appearance
Improved title: "Channel Anchor Reserves will be depleted" → "Consider your channel anchor reserves"
Context-aware messaging: Different explanations based on user action (spend/receive/swap)

2. Context-Aware Alert System
Added context prop with three scenarios:

"spend": "Keep at least X sats on-chain to maintain channel anchor reserves..."
"receive": "If you plan to spend from your hub wallet later, consider keeping at least X sats..."
"swap": Uses "spend" logic but adds external wallet suggestion

### DIFF
--- a/frontend/src/components/AnchorReserveAlert.tsx
+++ b/frontend/src/components/AnchorReserveAlert.tsx
@@ -3,15 +3,19 @@ import { Alert, AlertDescription, AlertTitle } from "src/components/ui/alert";
 import { useBalances } from "src/hooks/useBalances";
 import { useChannels } from "src/hooks/useChannels";
 
-export function AnchorReserveAlert({
-  amount,
-  className,
-  isSwap,
-}: {
+interface AnchorReserveAlertProps {
   amount: number;
   isSwap?: boolean;
   className?: string;
-}) {
+  context?: "spend" | "receive" | "swap";
+}
+
+export default function AnchorReserveAlert({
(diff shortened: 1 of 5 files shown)


frontend/src/components/AnchorReserveAlert.tsx
```diff
@@ -3,15 +3,19 @@ import { Alert, AlertDescription, AlertTitle } from "src/components/ui/alert";
...
   className?: string;
-}) {
+  context?: "spend" | "receive" | "swap";
+}
+
+export default function AnchorReserveAlert({
+  amount,
+  isSwap = false,
+  className,
+  context = "spend", // Default for backward compatibility
```
rolznz, line 17: We don't need backward compatibility as we have full control over Alby Hub (but it can still be fine to have a default value)

krrish-sehgal:
@rolznz , makes sense!

PR #1814, opened 2025-10-09
feat: show connection string right away for subwallets

reneaaron:
Prevent users from clicking "other apps" but show the connection string right away.

### DIFF
--- a/frontend/src/screens/apps/ConnectAppCard.tsx
+++ b/frontend/src/screens/apps/ConnectAppCard.tsx
@@ -1,9 +1,9 @@
 import { CheckIcon, CopyIcon, EyeIcon } from "lucide-react";
 import { useEffect, useState } from "react";
-import QRCode from "react-qr-code";
 import { Link } from "react-router-dom";
 import { AppStoreApp } from "src/components/connections/SuggestedAppData";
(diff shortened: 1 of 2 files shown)


frontend/src/screens/apps/ConnectAppCard.tsx
```diff
@@ -20,59 +19,78 @@ export function ConnectAppCard({
...
-      setTimeout(true);
-    }, 30000);
+    if (!showToasts) {
+      return;
+    }
+
+    if (!app.lastUsedAt && !effectRan.current) {
+      effectRan.current = true;
+
+      const toastId = toast.loading("Waiting for app to connect", {
```
rolznz: I don't see this at all when creating a sub-wallet. Doesn't it also make sense to show this in the standard create connection flow?

Also, I'm not sure if using toasts is better than the previous design.

reneaaron: I agree. Also the toasts bring some other problems with it (e.g. clearing any long-running toasts when navigating away, etc).

I removed the toast part and focused on the original purpose of the PR. 

Thanks for your feedback!

rolznz (approved):
tACK

PR #1821, opened 2025-10-10
fix: Improve mobile layout for ZapPlanner page

### LINKED ISSUE #1816: "New Recurring Payment" button should wrap on small screens
The "New Recurring Payment" is overlapping other elements when the screen is narrow:

Image

It should probably be placed below the header in this case.

### LINKED ISSUE #1817: ZapPlanner: Recurring Payments should be placed above suggestions
Especially on narrow screens, the recurring payments end up rather hard to find, at the bottom of the ZapPlanner page:

Image Image

I'd suggest to move the support suggestions below the active recurring payments.

Dunsin-cyber:
## fixes #1816 and fixes #1817
header on small scrren

also noticed irresponsiveness of the modal on small screens.
## before
zapplanner modal before

## after
zapplanner modal after

### DIFF
--- a/frontend/src/components/connections/AppStoreDetailHeader.tsx
+++ b/frontend/src/components/connections/AppStoreDetailHeader.tsx
@@ -29,31 +29,17 @@ export function AppStoreDetailHeader({
             <div className="flex flex-row items-center">
               <img
                 src={appStoreApp.logo}
-                className="w-14 h-14 rounded-lg mr-4"
+                className="w-14 h-14 rounded-lg mr-4 shrink-0"
               />
-              <div className="flex flex-col">
-                <div className="flex items-center gap-2">
+              <div className="flex gap-2 min-w-0">
+                <h2
+                  title={appStoreApp.title}
+                  className="text-xl font-semibold overflow-hidden text-ellipsis whitespace-nowrap"
+                >
                   {appStoreApp.title}
-                  {!!connectedApps.length && (
-                    <Badge
-                      variant="positive"
-                      className="hidden md:flex items-center gap-1"
-                    >
-                      <CheckCircleIcon className="w-3 h-3" />{" "}
-                      {connectedApps.length > 1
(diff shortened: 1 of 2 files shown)


frontend/src/components/connections/AppStoreDetailHeader.tsx
```diff
@@ -37,7 +37,7 @@ export function AppStoreDetailHeader({
                   {!!connectedApps.length && (
                     <Badge
                       variant="positive"
-                      className="flex items-center gap-1"
+                      className="hidden md:flex items-center gap-1"
```
rolznz: I don't like so much that we have to duplicate the badge. It was solved in another way here if you would like to have a look: https://github.com/getAlby/hub/blob/master/frontend/src/screens/apps/AppDetails.tsx#L194

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -352,12 +353,12 @@ export function ZapPlanner() {
...
-                  <PlusCircleIcon />
-                  New Recurring Payment
-                </Button>
+                <ResponsiveButton
+                  icon={PlusCircleIcon}
+                  text="New Recurring Payment"
+                />
               </DialogTrigger>
-              <DialogContent className="sm:max-w-[600px]">
+              <DialogContent className="sm:max-w-[600px] max-h-[85vh] overflow-y-auto">
```
rolznz, line 361: I wonder if these should be actually be set everywhere we use dialog content. Since it's a shadcn component it's best we don't edit it directly, but we could add a `ResponsiveDialogContent` class that renders DialogContent, and put that one in our custom components. What do you think? CC @reneaaron

frontend/src/components/connections/AppStoreDetailHeader.tsx
Dunsin-cyber: alright. done 
unify connection dbage

PR #1823, opened 2025-10-10
fix: Remove balance check from auto-swap limit

### LINKED ISSUE #1822: Remove spending balance requirement for setting up auto-swaps
When I set up an auto-swap it checks my current spending balance. Low spending balances  basically prevent users from setting up future auto-swaps.
Can we remove this requirement? 

Image

Dunsin-cyber:
Fixes #1822

Another thing worth considering is a potential logic flaw I've identified. Presently, there is no validation to ensure the `Spending balance threshold` is greater than or equal to the `Swap amount`.

This allows a user to create a configuration that will always fail. For example:

1.  Set **Balance Threshold** to `10,000` sats.
2.  Set **Swap Amount** to `25,000` sats.

The system will trigger the swap when the Lightning balance reaches 10,000 sats, but the action to swap 25,000 sats will immediately fail due to insufficient funds. This could lead to a confusing user experience with perpetually failing transactions.

**Suggestion:**
We should add a validation rule to the form that prevents saving if the `Swap amount` is greater than the `Balance threshold`.

### DIFF
--- a/http/http_service.go
+++ b/http/http_service.go
@@ -1510,12 +1510,6 @@ func (httpSvc *HttpService) enableAutoSwapOutHandler(c echo.Context) error {
 		})
 	}
 
-	if enableAutoSwapRequest.SwapAmount > enableAutoSwapRequest.BalanceThreshold {
-        return c.JSON(http.StatusBadRequest, ErrorResponse{
-            Message: "Swap amount cannot be greater than the balance threshold",
-        })
-    }
-
 	err := httpSvc.api.EnableAutoSwapOut(c.Request().Context(), &enableAutoSwapRequest)
 	if err != nil {
 		return c.JSON(http.StatusInternalServerError, ErrorResponse{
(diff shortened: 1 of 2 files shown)


Dunsin-cyber:
> @Dunsin-cyber I think that's a good suggestion. Would you like to add it in this PR?

alright, on it.

http/http_service.go
```diff
@@ -1510,6 +1510,12 @@ func (httpSvc *HttpService) enableAutoSwapOutHandler(c echo.Context) error {
 		})
 	}
 
+	if enableAutoSwapRequest.SwapAmount > enableAutoSwapRequest.BalanceThreshold {
```
rolznz: This won't apply in the wails / desktop app. We could move it to the API layer but I think it's fine just to do the check in the frontend in the `onSubmit` function.

Dunsin-cyber: alright.
okay, then, i'd limit the check to just the frontend

Dunsin-cyber: done

rolznz (approved):
utACK

PR #1825, opened 2025-10-13
fix: next payment description on supporter toast

### LINKED ISSUE #1818: Support Alby popup says first payment is immediate
When I chose to support Alby with a small amount based on its popup, it told me the first payment would be immediate:

https://github.com/getAlby/hub/blob/fd60afc349d06938e57cac5006618d5dcdc5432c/frontend/src/screens/alby/SupportAlby.tsx#L141-L143

But no payment happened. Checking later in ZapPlanner, I see the first payment is actually due at the start of the next month.

Either payments should indeed start immediately as the text claimed (which I personally think would be least confusing), or the text needs to be updated to say "Payment will be made at the start of each month" like at:

https://github.com/getAlby/hub/blob/fd60afc349d06938e57cac5006618d5dcdc5432c/frontend/src/screens/internal-apps/ZapPlanner.tsx#L332-L336

Dunsin-cyber:
fixes #1818
being a supporter2

### DIFF
--- a/frontend/src/screens/alby/SupportAlby.tsx
+++ b/frontend/src/screens/alby/SupportAlby.tsx
@@ -139,7 +139,7 @@ function SupportAlby() {
       });
 
       toast("Thank you for becoming a supporter", {
-        description: "The first payment is scheduled immediately.",
+        description: "Payment will be made at the start of each month",
       });
 
       navigate("/");


rolznz (approved):
utACK

PR #1836, opened 2025-10-15
Chore: simplified swap in and wallet receive to spending UX

rolznz:
Fixes https://github.com/getAlby/hub/issues/1783

Receive: (Receive to spending is the default, cannot pay with hub wallet, removed minimum - it's set on the input, used standard receiving capacity alert, remove "increase" link) 

image

image

image

Swap in (added radio select, default to external wallet, replaced reserve alert with tooltip, removed minimum amount - it's in the input, remove "increase" link):

image

image

image

Now when swapping in with on-chain balance, it will automatically pay with the hub's on-chain wallet. No need to click anything (+ no bad merged UI where there are both external wallet payment details and pay with alby hub button)

image

Node page: (Added swap button on desktop, changed advanced dropdown button icon)

image

### DIFF
--- a/frontend/src/screens/wallet/swap/index.tsx
+++ b/frontend/src/screens/wallet/swap/index.tsx
@@ -230,10 +230,7 @@ function SwapInForm() {
               id="internal"
               className="shrink-0"
             />
-            <Label
-              htmlFor="internal"
-              className="text-primary font-medium cursor-pointer"
-            >
+            <Label htmlFor="internal" className="font-medium cursor-pointer">
               On-chain balance
             </Label>
           </div>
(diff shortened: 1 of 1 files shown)


frontend/src/screens/channels/Channels.tsx
```diff
@@ -156,7 +156,7 @@ export default function Channels() {
               <DropdownMenu modal={false}>
                 <DropdownMenuTrigger>
                   <ResponsiveButton
-                    icon={EllipsisVerticalIcon}
```
rolznz: Jakub asked me to change this. But it's not in Figma yet, and the current figma design is different:

image

I'll leave it as-is for now

rolznz:
Thanks for the PR. We will need to do another iteration with @stackingsaunter 's latest design.

Issue #1837, opened 2025-10-15
Bug: Alby Hub does not include page title in history


rolznz:
<img width="1239" height="870" alt="Image" src="https://github.com/user-attachments/assets/d7628058-967e-4716-8678-ceef69a8f870" />

PR #1839, opened 2025-10-16
fix: name of route in history

krrish-sehgal:
fixes: #1837
It looks like this now: 

Screenshot 2025-10-16 at 11 53 57 AM

How I computed the title

I used useMatches() (React Router v6) to get the active route matches.
From each match I try to call match.handle?.crumb() if present — this returns a string (or array of strings for multi-part crumbs in some routes).
I take the last non-null crumb from the matched route list and set that as document.title. If none found, fallback is "Alby Hub"

### DIFF
--- a/frontend/src/components/layouts/AppLayout.tsx
+++ b/frontend/src/components/layouts/AppLayout.tsx
@@ -1,5 +1,4 @@
-import React from "react";
-import { Outlet, useMatches } from "react-router-dom";
+import { Outlet } from "react-router-dom";
 
 import { AppSidebar } from "src/components/AppSidebar";
 import { Banner } from "src/components/Banner";
@@ -10,6 +9,7 @@ import {
   useCommandPaletteContext,
 } from "src/contexts/CommandPaletteContext";
 import { useBanner } from "src/hooks/useBanner";
+import { useDocumentTitle } from "src/hooks/useDocumentTitle";
 import { useInfo } from "src/hooks/useInfo";
 import { useNotifyReceivedPayments } from "src/hooks/useNotifyReceivedPayments";
 import { useRemoveSuccessfulChannelOrder } from "src/hooks/useRemoveSuccessfulChannelOrder";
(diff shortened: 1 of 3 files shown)


frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -22,6 +23,65 @@ function AppLayoutInner() {
...
+      const title = Array.isArray(crumbTitle)
+        ? crumbTitle.join(" - ")
+        : crumbTitle;
+
+      // Set document title
+      document.title = title as string;
+
+      // Replace current history state to include title in state (some browsers show
+      // history entry title from state). We keep the existing state but add _title.
+      try {
```
rolznz: This seems hacky to me, I would avoid changing the history state directly. Do you know which browsers need this?

frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -22,6 +23,65 @@ function AppLayoutInner() {
   useRemoveSuccessfulChannelOrder();
   useNotifyReceivedPayments();
 
+  // Update document.title and the history entry when the route changes.
```
rolznz: Could this be moved to a separate hook? see the ones two lines above for examples

frontend/src/components/layouts/AppLayout.tsx
```diff
@@ -22,6 +23,65 @@ function AppLayoutInner() {
...
+  // This ensures the browser's history entries include a proper title
+  // (fixes: back gesture / long-press back showing empty/incorrect titles).
+  const matches = useMatches();
+  React.useEffect(() => {
+    try {
+      // Attempt to derive a title from route handles (crumb) or matched route
+      // fallback to app name. Use a small typed helper instead of `any` to
+      // satisfy lint rules.
+      const getCrumbFromHandle = (
```
rolznz: I don't think we can use crumbs. This is for breadcrumbs and does not accurately set the title for a specific page.

rolznz:
We haven't updated to React 19, but that seems a much cleaner way to set the page title:

https://react.dev/blog/2024/12/05/react-19#support-for-metadata-tags

frontend/src/components/layouts/AppLayout.tsx
krrish-sehgal: @rolznz , Would you like me to instead use ``.title`` from the handle field in routes.tsx? Should I add a title to every route which is just having the ``crumb`` field but missing the title field...

frontend/src/components/layouts/AppLayout.tsx
krrish-sehgal: yes! nice idea!

frontend/src/components/layouts/AppLayout.tsx
krrish-sehgal: I was getting a bug that was getting fixed with this. But its not there anymore... seems like the issue was elsewhere. 
I'll remove this , modern browsers don't need this, ``document.title`` is enough.

frontend/src/components/layouts/AppLayout.tsx
rolznz: I think this should be OK, but one downside is it will not be dynamic as far as I can tell.

It might be better to wait until we update to React 19. CC @reneaaron

krrish-sehgal: makes sense...

krrish-sehgal:
okay , marking it as draft again, then, will wait for further instrcutions on this. 
Thanks!

Issue #1831, opened 2025-10-13
Make Auto-swaps easier to find


rolznz:
There could be an easy way to enable it shown on the channel opened page.

Dunsin-cyber:
true, i have to navigate to it manually using the path `/wallet/swap/auto`

PR #1842, opened 2025-10-16
chore: add auto liquidity link on receive invoice page

rolznz:
Possibly closes https://github.com/getAlby/hub/issues/1831

Idea: users who receive payments can see that they can have a way to ensure they have incoming liquidity. It would need to be also added to other pages if we go this route.

image

### DIFF
--- a/frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
+++ b/frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
@@ -1,5 +1,6 @@
 import {
   ArrowLeftIcon,
+  ArrowUpDownIcon,
   CopyIcon,
   LinkIcon,
   PlusIcon,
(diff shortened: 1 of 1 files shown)


reneaaron:
I think it's kind of strange to only advertise when you just sent a payment (often times it doesn't really make sense in this context I guess, e.g. if you send small amounts and you have plenty of inbound left)

I am wondering if it would make more sense to offer this exactly when users need it, aka when they run out of inbound. If e.g. only 20% of inbound is reached we could possibly add an <Alert> in there to let the user know about the situation and then offer to automate that completely. ("Tired of manually swapping out? Try Auto-Swaps") 

What do you think?

rolznz:
@reneaaron this is for receiving, not sending (payment received screen).

I don't know if we can wait until the user is nearly unable to receive, but we could definitely add a more obvious warning / CTA when that is the case, true.

This does have the downside it will only show when users receive via the hub UI, but maybe that's ok? We already have a good CTA in the "low incoming capacity" email campaign sent to users, but this only happens when they generate an invoice larger than their remaining liquidity.

We also only send this event for LDK, not LND.

We could also add a new event and campaign when only 20% of inbound is remaining.

I'll add these to a new issue unrelated to this one (It's still good to somewhere in our UI to make auto-swaps more obvious)

rolznz:
Closing for now - @stackingsaunter has a new design to show an auto swap CTA on the swap page

PR #1858, opened 2025-10-20
fix: delete lightning address when deleting a sub-wallet

reneaaron:


frontend/src/hooks/useDeleteApp.ts
```diff
@@ -1,23 +1,45 @@
...
 import { toast } from "sonner";
 
+import { SUBWALLET_APPSTORE_APP_ID } from "src/constants";
+import { useApps } from "src/hooks/useApps";
 import { handleRequestError } from "src/utils/handleRequestError";
 import { request } from "src/utils/request";
 
 export function useDeleteApp(onSuccess?: (appPubkey: string) => void) {
   const [isDeleting, setDeleting] = React.useState(false);
+  const { data: appsData } = useApps();
```
rolznz: I don't think we can do this, `useApps` is paginated and we should directly find the exact app.

Maybe on the backend at the API level we could check and do the deletion of the lightning address instead

reneaaron: Ah, okay. Sounds good! :+1:

rolznz: @reneaaron I decided to fix the hook instead to be specific on what app is used. Please let me know if my changes look ok

api/api.go
```diff
@@ -318,6 +318,31 @@ func (api *api) UpdateApp(userApp *db.App, updateAppRequest *UpdateAppRequest) e
...
+			logger.Logger.WithError(err).WithFields(logrus.Fields{
+				"app_id": userApp.ID,
+			}).Error("Failed to deserialize app metadata")
+		} else {
+			appStoreAppId, hasAppStoreId := metadata["app_store_app_id"]
+			lud16, hasLightningAddress := metadata["lud16"]
+
+			isSubwallet := hasAppStoreId && appStoreAppId == constants.SUBWALLET_APPSTORE_APP_ID
+
+			if isSubwallet && hasLightningAddress && lud16 != nil {
```
im-adithya: All this is done to save a call to `GetAppById`, these checks are also done in `DeleteLightningAddress` again, should we remove this?

frontend/src/components/connections/DisconnectApp.tsx
```diff
@@ -54,14 +59,17 @@ export function DisconnectApp({
...
+            )}
           </AlertDialogDescription>
         </AlertDialogHeader>
         <AlertDialogFooter>
           <AlertDialogCancel onClick={onClose}>Cancel</AlertDialogCancel>
-          <AlertDialogAction
-            onClick={() => deleteApp(app.appPubkey)}
-            disabled={isDeleting}
-          >
+          <AlertDialogAction onClick={() => deleteApp()} disabled={isDeleting}>
```
im-adithya:
```suggestion
          <AlertDialogAction onClick={deleteApp} disabled={isDeleting}>
```

im-adithya (approved):
Minor nits, otherwise LGTM :+1:

api/api.go
reneaaron: Then we always have error logs if you delete an app without a lightning address? 🤔

im-adithya: Nice, the new commit looks good 🙌

PR #1864, opened 2025-10-27
feat: bip177

reneaaron:
BIP177 display option (new default)

image

### DIFF
--- a/frontend/src/components/CloseChannelDialogContent.tsx
+++ b/frontend/src/components/CloseChannelDialogContent.tsx
@@ -116,11 +116,13 @@ export function CloseChannelDialogContent({ alias, channel }: Props) {
               <Alert className="mb-4">
                 <AlertCircleIcon className="h-4 w-4" />
                 <AlertDescription>
-                  Closing this channel will move{" "}
-                  <FormattedBitcoinAmount amount={channel.localBalance} /> in
-                  this channel to your on-chain balance and reduce your receive
-                  limit by{" "}
-                  <FormattedBitcoinAmount amount={channel.remoteBalance} />.
+                  <div>
+                    Closing this channel will move{" "}
+                    <FormattedBitcoinAmount amount={channel.localBalance} /> in
+                    this channel to your on-chain balance and reduce your
+                    receive limit by{" "}
+                    <FormattedBitcoinAmount amount={channel.remoteBalance} />.
+                  </div>
                 </AlertDescription>
               </Alert>
               <div>
--- a/frontend/src/components/channels/ChannelsTable.tsx
+++ b/frontend/src/components/channels/ChannelsTable.tsx
@@ -207,6 +207,7 @@ function ChannelTableRow({
                 channel.localBalance,
                 channel.unspendablePunishmentReserve * 1000
               )}
+              showSymbol={false}
             />{" "}
             /{" "}
           </>
--- a/frontend/src/components/RebalanceChannelDialogContent.tsx
+++ b/frontend/src/components/RebalanceChannelDialogContent.tsx
@@ -136,9 +136,13 @@ export function RebalanceChannelDialogContent({
             <p className="mt-2 text-xs text-muted-foreground">
               Fee: 0.3%
               {!!amount && (
-                <FormattedBitcoinAmount
-                  amount={Math.floor(parseInt(amount || "0") * 0.003 * 1000)}
-                />
+                <>
+                  &nbsp;(
+                  <FormattedBitcoinAmount
+                    amount={Math.floor(parseInt(amount || "0") * 0.003 * 1000)}
+                  />
+                  )
+                </>
               )}{" "}
               + routing fees
             </p>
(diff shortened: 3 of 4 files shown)


api/api.go
```diff
@@ -1243,6 +1244,20 @@ func (api *api) SetCurrency(currency string) error {
 	return nil
 }
 
+func (api *api) SetBitcoinDisplayFormat(format string) error {
+	if format != "sats" && format != "bip177" {
```
rolznz: these strings are used in multiple files, so we could extract them to constants (see constants.go)

http/http_service.go
```diff
@@ -399,6 +399,16 @@ func (httpSvc *HttpService) updateSettingsHandler(c echo.Context) error {
 		})
 	}
 
+	// Handle bitcoin display format if provided
```
rolznz: This change would also needs to be applied in the wails handler, but I think most of this logic should be moved to the API layer (I think we should have an `UpdateSettings` method in the API). Then in the Wails handler and here we just call UpdateSettings.

frontend/src/components/CloseChannelDialogContent.tsx
```diff
@@ -117,9 +117,10 @@ export function CloseChannelDialogContent({ alias, channel }: Props) {
                 <AlertCircleIcon className="h-4 w-4" />
                 <AlertDescription>
                   Closing this channel will move{" "}
-                  {formatAmount(channel.localBalance)} sats in this channel to
```
rolznz, line 120: can we completely remove this utility function?

frontend/src/components/channels/ChannelsTable.tsx
```diff
@@ -196,8 +197,8 @@ function ChannelTableRow({
           <Badge variant="warning">Offline</Badge>
         )}
       </TableCell>
-      <TableCell title={capacity / 1000 + " sats"}>
-        {formatAmount(capacity)} sats
+      <TableCell>
+        <FormattedBitcoinAmount amount={capacity} />
```
rolznz, line 200: Minor: I noticed the reserve field can be hovered which says the value in sats, we can remove it, right?

frontend/src/screens/channels/Channels.tsx
```diff
@@ -24,6 +24,7 @@ import { LDKChannelWithoutPeerAlert } from "src/components/channels/LDKChannelWi
 import { OnchainTransactionsTable } from "src/components/channels/OnchainTransactionsTable.tsx";
 import EmptyState from "src/components/EmptyState.tsx";
 import ExternalLink from "src/components/ExternalLink";
+import { FormattedBitcoinAmount } from "src/components/FormattedBitcoinAmount";
```
rolznz, line 27: On the On-chain card on the balance warning (when you have 0 on-chain balance) I think it also needs to use this.

frontend/src/screens/wallet/index.tsx
```diff
@@ -11,6 +11,7 @@ import {
 import { Link } from "react-router-dom";
 import AppHeader from "src/components/AppHeader";
 import ExternalLink from "src/components/ExternalLink";
+import { FormattedBitcoinAmount } from "src/components/FormattedBitcoinAmount";
```
rolznz, line 14: On the receive -> invoice page the input says "Amount in satoshi..."
I wonder if later we could add this B prefix in front of the input (not sure it makes sense especially if this feature can be disabled?)

rolznz: Also e.g. on the waiting for payment page

image

frontend/src/screens/settings/Settings.tsx
```diff
@@ -71,6 +71,26 @@ function Settings() {
...
+  async function updateBitcoinDisplayFormat(bitcoinDisplayFormat: string) {
+    try {
+      await request("/api/settings", {
+        method: "PATCH",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+          bitcoinDisplayFormat,
+          currency: info?.currency || "USD", // Include current currency to avoid validation error
```
rolznz: we shouldn't need to do this - a patch should allow updating only the included fields

frontend/src/types.ts
```diff
@@ -163,6 +163,7 @@ export interface InfoResponse {
   currency: string;
   nodeAlias: string;
   mempoolUrl: string;
+  bitcoinDisplayFormat?: "sats" | "bip177";
```
rolznz: we manually specify the type here and in bitcoinFormatting.
Maybe `export type BitcoinDisplayFormat = "sats" | "bip177";` can be moved to this file and used here?

frontend/src/components/CloseChannelDialogContent.tsx
reneaaron: here maybe yes, but we still need to have a func to shorten big amounts (e.g. the preset buttons on the new channel flow for the size of the channel: 1M, 2M, 5M)

frontend/src/components/channels/ChannelsTable.tsx
reneaaron: Yeah, we now display the exact number here anyway.

PR #1879, opened 2025-10-30
feat: decouple ldk log

### LINKED ISSUE #1862: Decouple LDK logging from app logging
Currently to enable TRACE logging for LDK we need to enable TRACE app level logging too. By having separate loggers we can enable TRACE LDK logging without TRACE app level logs.

frnandu:
fixes #1862

### DIFF
--- a/lnclient/ldk/ldk_logger.go
+++ b/lnclient/ldk/ldk_logger.go
@@ -13,8 +13,6 @@ import (
 const ldkLogFilename = "ldk.log"
 const logDir = "log"
 
-var ldkLogFilePath string
-
 type ldkLogger struct {
 	logLevel logrus.Level
 	logger   *logrus.Logger
@@ -66,9 +64,8 @@ func createLogger(logLevel logrus.Level, logToFile bool, workDir string) (*logru
 	ldkLogger.SetLevel(logLevel)
 
 	if logToFile {
-
 		parentDir := filepath.Dir(workDir)
-		ldkLogFilePath = filepath.Join(parentDir, logDir, ldkLogFilename)
+		ldkLogFilePath := filepath.Join(parentDir, logDir, ldkLogFilename)
 		ldkFileLoggerHook, err := lumberjackrus.NewHook(
 			&lumberjackrus.LogFile{
 				Filename:   ldkLogFilePath,


README.md
```diff
@@ -158,7 +158,10 @@ The following configuration options can be set as environment variables or in a
 - `DATABASE_URI`: A sqlite filename or postgres URL. Default is SQLite DB `nwc.db` without a path, which will be put in the user home directory: $XDG_DATA_HOME/albyhub/nwc.db
 - `PORT`: The port on which the app should listen on (default: 8080)
 - `WORK_DIR`: Directory to store NWC data files. Default: $XDG_DATA_HOME/albyhub
-- `LOG_LEVEL`: Log level for the application. Higher is more verbose. Default: 4 (info)
+- `LOG_LEVEL`: Log level for the main application. Higher is more verbose. Default: 4 (info)
+- `LDK_LOG_LEVEL`: Log level for the LDK node. Lower is more verbose. Default: 3. This is separate from the main
```
rolznz: should this go in the LDK env variables section?

logger/logger.go
```diff
@@ -5,17 +5,21 @@ import (
...
 	"github.com/orandin/lumberjackrus"
 	"github.com/sirupsen/logrus"
 )
 
 const (
-	logDir      = "log"
-	logFilename = "nwc.log"
+	logDir         = "log"
+	logFilename    = "nwc.log"
+	ldkLogFilename = "ldk.log"
```
rolznz: Architecture-wise I feel like our main logger should not know about LDK. LDK is only one possible backend.

Can we put the code in the existing ldk_logger instead?

service/service.go
```diff
@@ -61,6 +61,7 @@ func NewService(ctx context.Context) (*service, error) {
 	}
 
 	logger.Init(appConfig.LogLevel)
+	logger.InitLDK(appConfig.LDKLogLevel)
```
rolznz: here can this be initialized when the LDK node is started instead (from the LDK code instead of the main service code?)

lnclient/ldk/ldk_logger.go
```diff
@@ -46,3 +53,38 @@ func mapLogLevel(logLevel ldk_node.LogLevel) logrus.Level {
...
+	LDKLogger = logrus.New()
+	LDKLogger.SetFormatter(&logrus.JSONFormatter{})
+	LDKLogger.SetOutput(os.Stdout)
+	ldkLogrusLogLevel, err := strconv.Atoi(ldkLogLevel)
+	if err != nil {
+		ldkLogrusLogLevel = int(logrus.InfoLevel)
+	}
+	LDKLogger.SetLevel(logrus.Level(ldkLogrusLogLevel))
+
+	if int(mapLogLevel(ldk_node.LogLevel(ldkLogrusLogLevel))) >= int(logrus.DebugLevel) {
```
rolznz: I don't think this is needed because the logs are only forwarded from LDK (which has its own line numbers)

lnclient/ldk/ldk_logger.go
```diff
@@ -1,13 +1,22 @@
...
+	"github.com/getAlby/hub/logger"
 	"github.com/getAlby/ldk-node-go/ldk_node"
+	"github.com/orandin/lumberjackrus"
 	"github.com/sirupsen/logrus"
 )
 
+const ldkLogFilename = "ldk.log"
+const logDir = "log"
+
+var LDKLogger *logrus.Logger
```
rolznz: I don't think we need a singleton logger - it can be as part of the `ldkLogger` struct

rolznz: and ideally closed when we stop the node

PR #1883, opened 2025-11-04
feat: limit default budget

reneaaron:
Apps should have a limit by default (new default limit = 10k/month)


frontend/src/screens/apps/NewApp.tsx
```diff
@@ -202,7 +202,7 @@ const NewAppInternal = ({ capabilities }: NewAppInternalProps) => {
     scopes: initialScopes,
     maxAmount: budgetMaxAmountMsatParam
       ? Math.floor(parseInt(budgetMaxAmountMsatParam) / 1000)
-      : 0,
+      : 10_000,
```
reneaaron: Should we maybe create a constant for that. It could make things more obvious...

PR #1891, opened 2025-11-06
fix: update go-nostr to fix de-duplicating of events

frnandu:
related https://github.com/nbd-wtf/go-nostr/pull/195


rolznz (approved):
utACK

Issue #1909, opened 2025-11-12
Bug: Relay pool can get into a state where it does not reconnect to the relay


rolznz:
My hub shows the relay is offline after running fine for several days. There were no logs of subscriptions being ended. It seems the pool fails to reconnect to the relay.

PR #1911, opened 2025-11-12
fix: handle unexpected subscription closure from relay pool

rolznz:
Closes https://github.com/getAlby/hub/issues/1909
- [x] fix relay status is offline when app first starts (sleep causes healthcheck to fail)
- [x] better handling of subscriptions closed with message (e.g. if relay times out accessing DB)
- [x] Updates go-nostr with relay reconnection & polling fixes

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1586,10 +1586,11 @@ func (api *api) Health(ctx context.Context) (*HealthResponse, error) {
 		alarms = append(alarms, NewHealthAlarm(HealthAlarmKindAlbyService, albyInfo.Incidents))
 	}
 
-	if len(api.svc.GetRelayStatuses()) > 0 {
+	relayStatuses := api.svc.GetRelayStatuses()
+	if len(relayStatuses) > 0 {
 		isAnyNostrRelayOffline := false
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -1586,16 +1586,18 @@ func (api *api) Health(ctx context.Context) (*HealthResponse, error) {
...
-	isAnyNostrRelayOffline := len(api.svc.GetRelayStatuses()) == 0
-	offlineRelayUrls := []string{}
-	for _, relayStatus := range api.svc.GetRelayStatuses() {
-		if !relayStatus.Online {
-			isAnyNostrRelayOffline = true
-			offlineRelayUrls = append(offlineRelayUrls, relayStatus.Url)
+	if len(api.svc.GetRelayStatuses()) > 0 {
+		isAnyNostrRelayOffline := false
+		offlineRelayUrls := []string{}
+		for _, relayStatus := range api.svc.GetRelayStatuses() {
```
im-adithya: nit: can we avoid calling twice?

rolznz: Good catch, thanks (also for concurrency)

PR #1914, opened 2025-11-14
feat: add retries for swap refund

### LINKED ISSUE #1765: Add retries to Boltz refund + add uncooperative refund support
We do a retry immediately if we get one of `boltz.TransactionLockupFailed, boltz.InvoiceFailedToPay, boltz.SwapExpired` when swapping in.

It's possible we try to get a refund too early which will result in the refund failing e.g.
> On our end we can see one​ legit refund attempt in our logs about 10 seconds after the swap failed, which we temporarily denied with "swap not eligible for a cooperative refund: lightning payment still in progress, try again in a couple minutes" because it takes a couple of seconds for Lightning nodes to actually consider a payment failed/cancelled. No more refund attempt thereafter.

We should add retries when calling the [RefundSwap](https://github.com/getAlby/hub/blob/master/swaps/swaps_service.go#L586) function.

Recommendation from Boltz:

> Start 5s, backoff 10s - 20 - 30 seconds ceiling, try do a coop refund until the swap timeout and then do the uncooperative refund to unilaterally force the refund.

We should support an uncooperative parameter in the refund method too (so if the swap timeout is hit then we pass uncooperative as true and do one last attempt)

And ensure the user can choose to do an uncooperative refund from the debug tools UI (by default it should be cooperative).

See https://api.docs.boltz.exchange/claiming-swaps.html#refund for more info about uncooperative refunds

to do an uncooperative refund all we need to do is set that boolean to false when constructing the transaction. https://github.com/BoltzExchange/boltz-client/blob/9e0ae15b5e26d64d67c1a5433725290cb3fee84a/internal/nursery/refund.go#L59

im-adithya:
Fixes #1765 

This doesn't add option to set cooperative in the frontend as it should be decided based on the SwapTimeout block height. Rest all is addressed.

### DIFF
--- a/swaps/swaps_service.go
+++ b/swaps/swaps_service.go
@@ -694,33 +694,32 @@ func (svc *swapsService) RefundSwap(swapId, address string, enableRetries bool) 
 		return err
 	}
 
-	waitToRetrySeconds := 0
 	var refundTransaction boltz.Transaction
 
 	for i := 0; ; i++ {
-		contextCancelled := false
 		select {
 		case <-svc.ctx.Done():
 			logger.Logger.WithField("swapId", swapId).Info("Swap refund context cancelled")
-			contextCancelled = true
-		case <-time.After(time.Duration(waitToRetrySeconds) * time.Second): // timeout
-		}
-		if contextCancelled {
 			return nil
+		case <-time.After(time.Duration(min(i*5, 30)) * time.Second): // timeout
 		}
 
 		nodeInfo, err := svc.lnClient.GetInfo(svc.ctx)
 		if err != nil {
 			logger.Logger.WithError(err).WithFields(logrus.Fields{
-				"swapId": swapId,
+				"swapId":    swapId,
+				"iteration": i,
 			}).WithError(err).Error("Failed to request node info")
-			return err
+			continue
 		}
 
 		feeRates, err := svc.getFeeRates()
 		if err != nil {
-			logger.Logger.WithField("swapId", swapId).WithError(err).Error("Failed to fetch fee rate to create claim transaction")
-			return err
+			logger.Logger.WithError(err).WithFields(logrus.Fields{
+				"swapId":    swapId,
+				"iteration": i,
+			}).Error("Failed to fetch fee rate to create claim transaction")
+			continue
 		}
 
 		cooperative := swapTransactionResp.TimeoutBlockHeight > nodeInfo.BlockHeight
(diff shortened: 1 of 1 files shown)


swaps/swaps_service.go
```diff
@@ -699,31 +694,78 @@ func (svc *swapsService) RefundSwap(swapId, address string) error {
...
-				Cooperative:        true,
+	waitToRetrySeconds := 0
+	var refundTransaction boltz.Transaction
+
+	for i := 0; ; i++ {
+		contextCancelled := false
+		select {
+		case <-svc.ctx.Done():
+			logger.Logger.WithField("swapId", swapId).Info("Swap refund context cancelled")
+			contextCancelled = true
```
rolznz: why not `return nil` here? is contextCancelled needed?

swaps/swaps_service.go
```diff
@@ -699,31 +694,78 @@ func (svc *swapsService) RefundSwap(swapId, address string) error {
...
+			logger.Logger.WithError(err).WithFields(logrus.Fields{
+				"swapId": swapId,
+			}).WithError(err).Error("Failed to request node info")
+			return err
+		}
+
+		feeRates, err := svc.getFeeRates()
+		if err != nil {
+			logger.Logger.WithField("swapId", swapId).WithError(err).Error("Failed to fetch fee rate to create claim transaction")
+			return err
```
rolznz: should these be continue (same with fetching node info) so we can retry now?

swaps/swaps_service.go
```diff
@@ -699,31 +694,78 @@ func (svc *swapsService) RefundSwap(swapId, address string) error {
...
-		logger.Logger.WithField("swapId", swapId).WithError(err).Error("Could not create claim transaction")
-		return err
+			boltz.Fee{
+				SatsPerVbyte: &fastestFee,
+			},
+			svc.boltzApi,
+		)
+		if err != nil {
+			if enableRetries && cooperative {
+				waitToRetrySeconds = max(waitToRetrySeconds, 0)
```
rolznz: this logic for the `waitToRetrySeconds` is kind of complicated. I think we can directly calculate it based on the for loop iteration variable `i` rather than adding. And how can it even become negative?

swaps/swaps_service.go
im-adithya: Ah true, changing!

swaps/swaps_service.go
im-adithya: Makes sense, done ✅

swaps/swaps_service.go
im-adithya: Changed it to be `min(i*5, 30)` ✅

rolznz:
Nice job!

Issue #1921, opened 2025-11-17
Bug: when opening channel and clicking paying with hub funds, QR code appears


rolznz:
The QR code should only appear if the user clicks "pay with another wallet". (Maybe this is if the balance drops below the threshold once the pay button is clicked?)

Also the notification on channel opening is a bit weird - it says "Successfully opened channel" but then we show a channel opening state

<img width="1732" height="1144" alt="Image" src="https://github.com/user-attachments/assets/88261313-5157-4889-9795-7133ef5f330d" />

<img width="2333" height="1214" alt="Image" src="https://github.com/user-attachments/assets/e3640e97-e574-45bd-a57f-d5a4d8ef0a7b" />

PR #1922, opened 2025-11-19
fix: minor manual open lightning channel ux issues

rolznz:
Fixes https://github.com/getAlby/hub/issues/1921
- do not show lightning invoice if balance drops below when paying
- remove unnecessary delay and toast when new channel is found in channel list


im-adithya (approved):
utACK

Issue #1877, opened 2025-10-29
First channel screen improvements


stackingsaunter:
Layout & copy improvements for first channel flow

[Link to Figma](https://www.figma.com/design/W8IlNozcOeYdCun6A7Qk8C/%F0%9F%86%95-Alby-Hub-2.0?node-id=614-263911&t=ZsugVTO7vQ7MwvXE-4)

## Screenshots

<img width="2560" height="1918" alt="Image" src="https://github.com/user-attachments/assets/eaa06169-6334-4a9a-a755-579f8d00f3c2" />

<img width="2560" height="1918" alt="Image" src="https://github.com/user-attachments/assets/682609d2-93bf-4c93-b1cb-21f4d4e8667c" />

## Illustration

<img width="1435" height="631" alt="Image" src="https://github.com/user-attachments/assets/92eb1056-ac7d-4929-a064-befe2b3f6ab8" />

<img width="1435" height="631" alt="Image" src="https://github.com/user-attachments/assets/0c620791-4aa1-43e4-ad16-46772871721f" />

![Image](https://github.com/user-attachments/assets/3cd54c62-a475-4184-8d09-4e3e66de8bee)

![Image](https://github.com/user-attachments/assets/eea14cc5-2678-4678-aedc-b6598796f38a)

PR #1925, opened 2025-11-19
feat: improve first channel screen UI

rolznz:
Fixes https://github.com/getAlby/hub/issues/1877

Also a minor fix to remove hard-coding of channel duration for manual LSP channels.

image

image

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -31,7 +31,8 @@ body {
 
 @layer base {
   [role="menuitem"]:not([disabled]),
-  [data-slot="command-item"] {
+  [data-slot="command-item"],
+  [data-slot="field"] {
     cursor: pointer !important;
   }
 }


rolznz:
@reneaaron to test the free channel payment method you can hardcode it e.g.

```
const { data: _lspChannelOffer } = useLSPChannelOffer();
  const lspChannelOffer = {
    ..._lspChannelOffer,
    currentPaymentMethod: "included",
  };
  ```
  
  Note: in mutinynet, the current payment method is always `prepaid` e.g. Alby never pays for the channel.

File: frontend/src/screens/channels/first/FirstChannel.tsx [hunk 1]
```diff
@@ -71,7 +75,19 @@ export function FirstChannel() {
...
+      return;
+    }
+    if (
+      (lspChannelOffer.currentPaymentMethod === "card" ||
+        lspChannelOffer.currentPaymentMethod === "wallet") &&
+      currentPaymentMethod !== lspChannelOffer.currentPaymentMethod
+    ) {
+      toast.error("Payment method incorrectly configured", {
+        description: currentPaymentMethod
+          ? "Please switch the payment method and confirm the change in your Alby Account settings"
```
reneaaron, line 93: When would this case happen? 🤔 When the user didn't yet configure a payment method at all?

frontend/src/components/RadioGroup.tsx
```diff
@@ -0,0 +1,26 @@
+import * as RadioGroupPrimitive from "@radix-ui/react-radio-group";
+import { Label } from "src/components/ui/label";
+import { RadioGroupItem as ShadcnRadioGroupItem } from "src/components/ui/radio-group";
+import { cn } from "src/lib/utils";
+
+export function RadioGroupItem(
```
rolznz: Use ChoiceCard instead from Shadcn Field

PR #1926, opened 2025-11-20
fix: ensure relays on startup

rolznz:
if the user has no apps then the relays are never connected to. This leads to the relays incorrectly shown as offline

### DIFF
--- a/service/start.go
+++ b/service/start.go
@@ -59,9 +59,14 @@ func (svc *service) startNostr(ctx context.Context) error {
 		}),
 	))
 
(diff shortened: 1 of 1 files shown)


service/start.go
```diff
@@ -59,9 +59,14 @@ func (svc *service) startNostr(ctx context.Context) error {
 		}),
 	))
 
+	// initially try connect to relays (if hub has no apps, pool won't connect to apps by default)
```
im-adithya, line 62: minor: connect to relays*

PR #1930, opened 2025-11-24
Fix: Allow Account Settings Page to Load

### LINKED ISSUE #1929: Bug: Deleting Alby Account leaves NWC connection open, causing problems
If a user deletes their getAlby.com account, the Alby Hub app is left with an open NWC connection "app" to Alby, with no accessible UI control for disconnecting it. The Wallet -> Receive link is also left in an unusable site, so the user cannot create invoices to their node. Permanent spinners are left in several places, and the menu in the bottom-left corner changes from "My Alby Hub" to "SN" or other letters.

Additionally, the local database may be populated with Oauth credentials for a now-non-existent Alby account. The OAuth tokens remain in the user_configs table in the database. Specifically:
- AlbyOAuthAccessToken - The access token for the Alby API
- AlbyOAuthRefreshToken - The refresh token
- AlbyOAuthAccessTokenExpiry - Token expiration timestamp
- UserIdentifier - The user's Alby account identifier
- AlbyLightningAddress - The user's lightning address

These are stored in the Hub's local SQLite database (albyhub.db) and persist even after the account is deleted on getalby.com.

### Before creating a getAlby.com account:

Wallet -> Receive forwards to `/receive/invoice`, and the user sees:
Image

### After creating a getAlby.com account:

Wallet -> Receive, the user sees:
Image

### After deleting a getAlby.com account:

Wallet -> Receive, the user sees perma-spinner, with error flash message:
Image
There is no way to create invoices.

Connections -> Connected Apps -- permanent spinner indicated:
Image
There is no way to disconnect the dysfunctional NWC connection.

Clicking the "Alby Account Settings" from the Connected app goes to a settings page that is empty, except for a perma-spinner:
Image
The disconnect feature would be here, if the getAlby.com account were still available.

The app remains in a state where it expects there to be an Alby account, with no way for the user to disconnect the NWC app.

The server logs have many messages like:
```
albyhub  | {"body":"404 page not found\n","level":"error","msg":"users endpoint returned non-success code","status_code":404,"time":"2025-11-24T00:25:40Z"}
albyhub  | {"error":"users endpoint returned non-success code: 404 page not found\n","level":"error","msg":"Failed to request alby me endpoint","time":"2025-11-24T00:25:40Z"}
```

netmojo:
## Summary

Fixes #1929

Fixed the Settings -> Your Alby Account page so that when the getAlby.com account is deleted, the page still loads, so that the "Unlink Alby Account" button is available.

## Changes
- Loading the "Alby Account Settings" page no longer depends on a getAlby.com account existing
- The "Alby Account Settings" button/links now always point to `/settings/alby-account`, regardless of whether the user has an Alby account.

### Docker Image
I published a multi-platform image of this branch to Docker Hub at `netmojo/albyhub:pr1930`, so it can be tested by:

```
stop AlbyHub
edit `albyhub/docker-compose.yml` to change the image line to: `image: netmojo/albyhub:pr1930`
start AlbyHub
```

### With an Alby Account
with-account

### With a Linked Alby Account
account-linked

### With a Deleted Alby Account, and left-over Alby Hub Connection
Perma-spinner appears, and "Alby Account Settings" button points to `/settings/alby-account`
no-account-card

### With a Deleted Alby Account, but a left-over Alby Hub Connection and Linked Wallet
account-deleted

### Settings -> Your Alby Account Page now loads when Alby Account is Deleted
no-account-settings

### Disconnect Modal
no-account-disconnect-modal

### After Disconnect
Redirects to home with flash message
account-unlinked

### Wallet -> Receive works again
receive-after-disconnect

### Settings -> Alby Account restored to default after unlink:
settings-after-disconnect

### User menu restored to default after unlink
after-unlink-menu

After disconnecting the dangling NWC, I see no more 404s in the logs.

### DIFF
--- a/frontend/src/components/connections/AlbyConnectionCard.tsx
+++ b/frontend/src/components/connections/AlbyConnectionCard.tsx
@@ -1,12 +1,14 @@
 import {
   CheckCircle2Icon,
   CircleXIcon,
+  EditIcon,
   InfoIcon,
   Link2Icon,
   User2Icon,
   ZapIcon,
 } from "lucide-react";
 import { FormEvent, useState } from "react";
+import { Link } from "react-router-dom";
 
 import BudgetAmountSelect from "src/components/BudgetAmountSelect";
 import BudgetRenewalSelect from "src/components/BudgetRenewalSelect";
@@ -71,7 +73,7 @@ function AlbyConnectionCard() {
   return (
     <Card>
       <CardHeader>
-        <CardTitle>
+        <CardTitle className="relative">
           Linked Alby Account
           {albyAccountApp && <AppCardNotice app={albyAccountApp} />}
         </CardTitle>
@@ -181,6 +183,12 @@ function AlbyConnectionCard() {
           </div>
           {albyAccountApp && (
             <div className="slashed-zero">
+              <Link
+                to={`/apps/${albyAccountApp.id}?edit=true`}
+                className="absolute top-0 right-0"
+              >
+                <EditIcon className="size-4 hidden group-hover:inline text-muted-foreground hover:text-card-foreground" />
+              </Link>
               <AppCardConnectionInfo
                 connection={albyAccountApp}
                 budgetRemainingText={


rolznz:
@netmojo I think we should only make the change that you have in the AlbyAccount.tsx file. We do not want to add a lot of code to optimize for users deleting their Alby Account - it's not a common action.

rolznz:
> if you applied less pressure to setup custodial account

We don't have custodial accounts. The custodial wallet has been shut down for a long time now, and Alby Hub is the self-custodial wallet that you can optionally connect to get e.g. a lightning address, encrypted backups etc.

The important thing is they can disconnect the account to get out of this state, which the fix in the AlbyAccount.tsx file should do. The checks you added DO make the code more complicated to handle an edge case (a user deletes their alby account before disconnecting it from the hub).

I would even simplify our original code further to always link to the hub settings page actually, rather than conditionally going to getalby.com vs the settings page. Then we don't need these checks like `(!info?.albyAccountConnected || albyMe)` etc that you added.

netmojo:
Simplified to always point `Alby Account Settings` to the local settings page.

PR #1942, opened 2025-11-25
fix: consistent use of bitcoinDisplayFormat

### LINKED ISSUE #1940: Bug: formatBitcoinAmount display format should be required
There's a few places in the app which don't respect the user's unit display. If you set your unit display to sats, it'll still show the B symbol (e.g. on received payment notification)

reneaaron:
Fixes #1940

### DIFF
--- a/frontend/src/hooks/useNotifyReceivedPayments.ts
+++ b/frontend/src/hooks/useNotifyReceivedPayments.ts
@@ -1,6 +1,5 @@
 import React from "react";
 import { toast } from "sonner";
-import { BITCOIN_DISPLAY_FORMAT_BIP177 } from "src/constants";
 import { useInfo } from "src/hooks/useInfo";
 import { useTransactions } from "src/hooks/useTransactions";
 import { Transaction } from "src/types";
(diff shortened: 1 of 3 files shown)


frontend/src/hooks/useNotifyReceivedPayments.ts
```diff
@@ -1,25 +1,31 @@
...
       return;
     }
     const latestTx = transactionsData.transactions[0];
     if (latestTx !== prevTransaction) {
       if (prevTransaction && latestTx.type === "incoming") {
         toast("Payment received", {
-          description: formatBitcoinAmount(latestTx.amount),
+          description: formatBitcoinAmount(
+            latestTx.amount,
+            info.bitcoinDisplayFormat || BITCOIN_DISPLAY_FORMAT_BIP177
```
rolznz: I see a lot of places we have this fallback. Could the backend instead set a fallback in a single place (in the info endpoint) and we make the display format non-optional in the info response?

reneaaron: You'd still need to add a fallback to while the info response is being loaded, then?

rolznz: There is already a check to return if `info` isn't loaded.

Also if info isn't loaded nothing will show at all in the hub (it's a prerequisite for everything) - normally we just return a loading spinner on the pages that need it, if info isn't loaded yet

Issue #1943, opened 2025-11-25
config GetRelayUrls() is called too many times


rolznz:
This causes a problem with hubs that have many apps because it's called per app when connecting to the relay. This then can cause the DB calls to timeout.

We should also catch the error.

Should the value be cached, because it never changes for the lifetime of the app?

PR #1944, opened 2025-11-26
chore: add method to return cached config value

rolznz:
Fixes https://github.com/getAlby/hub/issues/1943

Now with Postgres these config Gets are doing remote API calls and we need to be careful not to do too many of them.

(Just the general idea, we can apply to most of the config `Get`s that never change over the app's lifetime if it sounds good)

TODO:
- [x] review general idea
- [x] apply cache to existing methods

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -157,7 +157,7 @@ func (cfg *config) GetJWTSecret() string {
 }
 
 func (cfg *config) GetRelayUrls() []string {
-	relayUrls, _ := cfg.GetCached("Relay", "")
+	relayUrls, _ := cfg.Get("Relay", "")
 	return strings.Split(relayUrls, ",")
 }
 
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -180,6 +184,23 @@ func (cfg *config) Get(key string, encryptionKey string) (string, error) {
 	return cfg.get(key, encryptionKey, cfg.db)
 }
 
+func (cfg *config) GetCached(key string, encryptionKey string) (string, error) {
```
bumi: do we even need to methods for this? 
can't/shouldn't this always be cached?

bumi (approved):
If that's the Go way of doing this 👍

config/config.go
rolznz: True. So we delete the key from the cache when we set a new value.

PR #1947, opened 2025-11-28
Fix/1939 filter transactions by amount

### LINKED ISSUE #1939: [Feature Request] Filter based on TX amount in main wallet history
One of my subwallets has a lot of small payments and completely floods the main wallet history.

It would be great to have an option to show/hide specific subwallets in the main wallet transaction list.

As you can see in the attached screenshot, a single subwallet currently takes over almost the entire history, making it hard to see other transactions from the main wallet or from other subwallets. A filter would make the history much easier to read and use.

Image

_By the way, I'd like to work on this feature, it is a task well fitted for my skill-level, and an opportunity for me to gain more experience, but before starting a possible implementation, I wanted to hear some feedback._

NodeDiver:
Fixes #1939 This PR adds a transaction amount filter to the wallet screen, allowing users to filter transactions by minimum satoshi amounts (10, 100, 1K, 10K, 100K sats). 
It will help to filter out zaps or recurrent small transactions from the bigger ones.
I used the AI to help me do it and did a thorough review of everything.

An unintended visualization bug appeared in the first attempt to add the filter, and I had to modify empty states: a generic "No transactions yet" message was incorrectly displayed when a filter returned no results, even though transactions existed. There is a distinction between empty transaction lists and filtered empty states, showing an appropriate message with a "Show All Transactions" button when the chosen filters hide all transactions.

Preview in the video:

https://github.com/user-attachments/assets/1f0efc62-4e8d-45ee-aac1-96a2674085f7

_Note: Commit 174c43ac is a duplicate of the already-merged PR #1876 and can be ignored._


frontend/src/components/TransactionsList.tsx
```diff
@@ -27,6 +32,10 @@ function TransactionsList({
   const transactions = transactionData?.transactions || [];
   const totalCount = transactionData?.totalCount || 0;
 
+  const filteredTransactions = minAmountSats
+    ? transactions.filter((tx) => Math.floor(tx.amount / 1000) >= minAmountSats)
```
rolznz, line 36: we should do filtering on the backend otherwise pagination will break

frontend/src/components/TransactionsFilterMenu.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const TransactionsFilterMenu = ({
+  selectedFilter,
+  onFilterChange,
+}: TransactionsFilterMenuProps) => {
+  return (
+    <DropdownMenu>
+      <Button asChild size="icon" variant="secondary">
+        <DropdownMenuTrigger>
+          <FilterIcon className="h-4 w-4" />
```
rolznz, line 32: I wonder if filtering should be an option in the "..." menu so we don't clutter the interface too much?

frontend/src/components/TransactionsList.tsx
rolznz: this is why you see the weird behavior too

NodeDiver: If you want me to move the filtering to the backend, I think I can do it, but for next week

frontend/src/components/TransactionsFilterMenu.tsx
NodeDiver: I don't have a strong opinion about this if it has to be on the 3 dots menu we can place it there.

I also thought that " the least intrusive the better "., Nonetheless, if the filtering is on the 3 dots menu, it is also hidden, and might not even be discovered. That is why I chose to put it on its own button, but just an icon to save some space.

PR #1956, opened 2025-12-01
make sure verify.sh is replaced by last version when downloaded

### LINKED ISSUE #1563: Verification fails: `SHA256 hash mismatch! The file may be corrupted or tampered with.`
Using the [pi4 quick start](https://github.com/getAlby/hub/tree/master/scripts/pi-aarch64) to upgrade, I've never been able to have the verification work. After downloading manifest file and signature file, it always complains about SHA256 mismatch.

Sakrecoer:
fixes #1563

### DIFF
--- a/scripts/verify.sh
+++ b/scripts/verify.sh
@@ -25,13 +25,13 @@ verify_package() {
   done
 
   echo "Downloading manifest file..."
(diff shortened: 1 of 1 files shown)


rolznz (approved):
tACK, thanks!

PR #1969, opened 2025-12-11
fix: include encryptionKey in the cache key (sha256)

frnandu:
Since `encryptionKey` is not being used as part of the cache key accessing the value, this makes it possible for a cached value to be accessed even with wrong password/encryption-key so should be fixed as critically as possible.

This PR refactors the cache to use a nested 2-level map where the `encryptionKey` is used as a key for the 2nd level map, allowing to efficiently clear all possible `encryptionKey` combinations of a cache key without mutex blocking for too long.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -36,6 +36,12 @@ func TestCheckCache_NoEncryptionKey(t *testing.T) {
 	require.Equal(t, "value", value)
 
 	require.Equal(t, cfg.cache["key"][""], "value")
+
+	// test we can access the cached value without the db
+	cfg.db = nil
+	value, err = cfg.Get("key", "")
+	require.NoError(t, err)
+	require.Equal(t, "value", value)
 }
 
 func TestCheckUnlockPasswordCache(t *testing.T) {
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -235,7 +246,16 @@ func (cfg *config) set(key string, value string, clauses clause.OnConflict, encr
 	logger.Logger.WithField("key", key).Debug("clearing config cache")
 	cfg.cacheMutex.Lock()
 	defer cfg.cacheMutex.Unlock()
-	delete(cfg.cache, key)
+
+	if encryptionKey != "" {
+		for cacheKey := range cfg.cache {
+			if cacheKey == key || strings.HasPrefix(cacheKey, key+":") {
```
rolznz: Can it be simplified to just this?
```suggestion
			if strings.HasPrefix(cacheKey, key) {
```

config/config.go
```diff
@@ -235,7 +246,16 @@ func (cfg *config) set(key string, value string, clauses clause.OnConflict, encr
...
+				delete(cfg.cache, cacheKey)
+			}
+		}
+	} else {
```
rolznz: normally we try to avoid `else` where possible. It could just always do this delete as it's probably not a performance issue

config/config.go
```diff
@@ -180,10 +181,20 @@ func (cfg *config) GetMempoolUrl() string {
 	return strings.TrimSuffix(mempoolApiUrl, "/api")
 }
 
+func (cfg *config) getCacheKey(key string, encryptionKey string) string {
+	if encryptionKey == "" {
+		return key
+	}
+	hash := sha256.Sum256([]byte(encryptionKey))
+	return key + ":" + hex.EncodeToString(hash[:8])
```
rolznz: Maybe add the comment:
> For cache key purposes, 8 bytes (16 hex chars) provides:
2^64 possible values = ~18 quintillion combinations
More than sufficient to avoid collisions for cache keys

config/config.go
```diff
@@ -180,10 +181,20 @@ func (cfg *config) GetMempoolUrl() string {
 	return strings.TrimSuffix(mempoolApiUrl, "/api")
 }
 
+func (cfg *config) getCacheKey(key string, encryptionKey string) string {
```
rolznz: TODO to self: add a test

frnandu: rewrote this into only `getEncryptionKeyHash` since now it uses a nested 2-level map, there is no string concatenation anymore

rolznz (approved):
tACK

PR #1973, opened 2025-12-12
feat: encrypt xpub for auto swap  outs

### LINKED ISSUE #1928: feat: encrypt xpub for auto-swaps

frnandu:
When using an xpub for auto swap outs, the xpub was saved in config unencrypted, and this is unsafe.
fixes #1928

## Summary by CodeRabbit

* **New Features**
  * Encrypted storage for external XPUB destinations; enabling an external XPUB now requires providing an unlock password which is submitted and stored.
  * Auto-swap caches a decrypted XPUB destination for subsequent use.

* **Bug Fixes**
  * Service startup is now validated earlier when getting or enabling auto-swap, failing fast if not ready.
  * Request-side validation added for destination type, XPUB/address format, and unlock password.

* **UI**
  * Unlock-password dialog added to the auto-swap form; password is included in requests and cleared after success.
  * Minor layout refinements in swap status and auto-swap form rendering.

### DIFF
--- a/frontend/src/screens/wallet/swap/AutoSwap.tsx
+++ b/frontend/src/screens/wallet/swap/AutoSwap.tsx
@@ -2,6 +2,7 @@ import {
   ArrowDownUpIcon,
   ClipboardPasteIcon,
   ClockIcon,
+  CopyIcon,
   MoveRightIcon,
   XCircleIcon,
 } from "lucide-react";
@@ -28,6 +29,7 @@ import { Input } from "src/components/ui/input";
 import { Label } from "src/components/ui/label";
 import { RadioGroup, RadioGroupItem } from "src/components/ui/radio-group";
 import { useAutoSwapsConfig, useSwapInfo } from "src/hooks/useSwaps";
+import { copyToClipboard } from "src/lib/clipboard";
 import { AutoSwapConfig } from "src/types";
 import { request } from "src/utils/request";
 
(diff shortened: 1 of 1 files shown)


swaps/swaps_service.go
```diff
@@ -214,15 +225,17 @@ func (svc *swapsService) EnableAutoSwapOut() error {
...
+				if svc.autoSwapOutDecryptedXpub != "" {
+					actualDestination, err = svc.getNextUnusedAddressFromXpub()
+					if err != nil {
+						logger.Logger.WithError(err).Error("Failed to get next address from xpub")
+						continue
 					}
+					usedXpubDerivation = true
+				} else if swapDestination != "" {
+					// Regular address (not XPUB)
+					actualDestination = swapDestination
```
im-adithya: Is this necessary, we already set this above in L226

swaps/swaps_service.go
```diff
@@ -49,18 +49,21 @@ type swapsService struct {
 	boltzWs             *boltz.Websocket
 	swapListeners       map[string]chan boltz.SwapUpdate
 	swapListenersLock   sync.Mutex
+	autoSwapOutDecryptedXpub       string
```
im-adithya: Do we need this? I think we can avoid this and hence `GetDecryptedAutoSwapXpub` by using `ValidateXpub` itself in the `GetAutoSwapConfig` and passing xpub param to `getNextUnusedAddressFromXpub` instead of using `svc.autoSwapOutDecryptedXpub` (or leaving it as is before this change)

frontend/src/screens/wallet/swap/AutoSwap.tsx
```diff
@@ -269,6 +280,21 @@ function AutoSwapOutForm() {
...
+                value={unlockPassword}
+                onChange={setUnlockPassword}
+                placeholder="Enter your unlock password"
+                required
+              />
+              <p className="text-xs text-muted-foreground">
+                Your password is required to encrypt the XPUB for secure storage
+              </p>
+            </div>
+          )}
```
im-adithya: It would be better to open this in a dialog (similar to how we do it for connections: [SuperuserConfirmPasswordDialog](https://github.com/getAlby/hub/blob/5281681776884b6badc86726b65f246efe856a4f/frontend/src/screens/apps/NewApp.tsx#L567))

swaps/swaps_service.go
frnandu: ValidateXpub only checks format; it does not decrypt config values,
cfg.Get(key, "") returns ciphertext when the value is encrypted (config/config.go:230), so in GetAutoSwapConfig you cannot recover a real XPUB from config without an encryption key.

frontend/src/screens/wallet/swap/AutoSwap.tsx
frnandu: re-done as a dialog, thanks for suggestion!

Issue #1986, opened 2025-12-15
Encrypt config JWTSecret with unlock password


rolznz:
Most of the time the JWT secret is not passed through env, and we now dynamically return the JWT secret (with custom `KeyFunc`) it should be possible for the JWT secret to be encrypted in the DB and only decrypted while in memory.

We can detect if it is not passed in via env and if necessary do a 1-time migration by checking if the config entry is encrypted or not (with the `encrypted` column)

PR #1988, opened 2025-12-16
chore: use encrypted jwt secret

rolznz:
Fixes https://github.com/getAlby/hub/issues/1986

**NOTES**
- JWT_SECRET env variable has been removed to reduce complexity. Now we always use a generated JWT secret.
- The migration will **log out** any apps that use the hub API with a developer access token, so users will need to generate a new developer token after updating to 1.21.1.

TODOs:
- [x] if JWT secret exists but is not encrypted, do 1-time migration to encrypt it (actually generates a brand new secret, it doesn't encrypt the existing one)
- [x] if JWT secret does not exist, generate a new JWT secret and save it encrypted with the unlock password
- [x] Simplify change password to just delete the JWT secret

Tests
- [x] ensure all sessions are logged out after password change
- [x] cannot login with invalid password
- [x] check DB entry that newly-generated JWT secret is encrypted

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -142,6 +142,8 @@ func (cfg *config) Unlock(encryptionKey string) error {
 		return errors.New("incorrect password")
 	}
 
+	// TODO: remove encryptedJwtSecret check after 2027-01-01
+	// - all hubs should have updated to use an encrypted JWT secret by then
 	encryptedJwtSecret, err := cfg.Get("JWTSecret", "")
 	if err != nil {
 		return err


config/config.go
```diff
@@ -321,30 +334,24 @@ func (cfg *config) ChangeUnlockPassword(currentUnlockPassword string, newUnlockP
 			logger.Logger.WithField("key", userConfig.Key).Info("re-encrypted key")
 		}
 
-		newSecret, err := randomHex(32)
+		// delete the JWT secret so it will be re-generated on next unlock (to log all sessions out on password change)
+		err = tx.Where(&db.UserConfig{Key: "JWTSecret"}).Delete(&db.UserConfig{}).Error
```
bumi, line 340: we currently create the secret in the Unlock function, don't we? so it's a bit of a side effect and not explicit and easy noticeable. 

is the deletion here enough? or maybe could it be RegenerateJWTSecret() or something?

config/config.go
```diff
@@ -148,13 +129,45 @@ func (cfg *config) SetupCompleted() bool {
...
-		logger.Logger.WithError(err).Error("Failed to retrieve JWTSecret from database")
-		return ""
+		return err
 	}
-	return secret
+	jwtSecret, err := cfg.Get("JWTSecret", encryptionKey)
+	if err != nil {
+		return err
+	}
+	// generate a new one if none exists yet OR if the user has an unencrypted secret
```
bumi, line 155: does this now always need to be in here?

http/http_service.go
```diff
@@ -285,6 +294,17 @@ func (httpSvc *HttpService) startHandler(c echo.Context) error {
 		})
 	}
 
+	// NOTE: the config is also unlocked as part of the start
+	// goroutine below. But since we execute start asynchronously it's hard to
+	// know when the config has been unlocked before being able to create the JWT token
+	err := httpSvc.cfg.Unlock(startRequest.UnlockPassword)
```
bumi, line 300: so it runs twice potentially in parallel?

rolznz: No, it's not in parallel - this first call here is synchronous

config/config.go
rolznz: hmm. The unencrypted secret check could be removed again after a period of time. I will add a comment.

I originally thought it would have to always check because of the env variable, but I've removed the env variable now which has simplified it.

config/config.go
rolznz: I understand there is a side effect, but I think this is inevitable because now the JWT secret is stored decrypted in memory - and that is a side effect that we have to keep in mind. Also this code is inside a DB transaction, so the code for generating a new JWT secret had to be duplicated and slightly different than the other generation code.

I covered this quite well with tests I think. Do you think it's ok as-is?

http/http_service.go
rolznz: Actually it doesn't make sense to call it in start. I can just call it here since it's only needed for the HTTP mode. So maybe the name of this function `config.Unlock` should also be renamed to be `UnlockJWTSecret`)

rolznz: Update: this doesn't work so well with auto-unlock. I will leave this as a follow-up improvement

PR #1992, opened 2025-12-17
a11y: improve disabled and labeling support in ResponsiveButton

divyansh-v15-06:
Improves accessibility semantics for `ResponsiveButton`.

- Adds accessible labeling for icon-only buttons on smaller screens
- Ensures `aria-disabled` is set when the button is disabled
- Marks decorative icons as `aria-hidden`

This keeps visual behavior unchanged while improving keyboard and
screen-reader support, especially when `asChild` is used.

### DIFF
--- a/frontend/src/components/ResponsiveButton.tsx
+++ b/frontend/src/components/ResponsiveButton.tsx
@@ -5,7 +5,7 @@ import { Button } from "./ui/button";
 import { buttonVariants } from "./ui/buttonVariants";
 
 type Props = {
-  icon: React.ComponentType;
+  icon: React.ComponentType<{ className?: string }>;
   text: string;
 } & React.ComponentProps<"button"> &
   VariantProps<typeof buttonVariants> & {
@@ -19,11 +19,12 @@ const ResponsiveButton = ({
   asChild,
   children,
   className,
+  disabled,
   ...props
 }: Props) => {
   const content = (
     <>
-      <Icon />
+      <Icon aria-hidden="true" />
       <span className="hidden lg:inline">{text}</span>
     </>
   );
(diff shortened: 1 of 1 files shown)


frontend/src/components/ResponsiveButton.tsx
```diff
@@ -33,9 +34,12 @@ const ResponsiveButton = ({
       {...props}
       variant={variant}
       asChild={asChild}
+      disabled={disabled}
```
rolznz, line 37: I don't think we should add these additional props - they should be handled by shadcn (our component is just a wrapper for a shadcn button).

divyansh-v15-06: Thanks for the clarification, that makes sense 👍

I agree that accessibility semantics should live in the underlying shadcn Button rather than being duplicated in a wrapper component. Since ResponsiveButton is intentionally thin, adding extra props here could introduce inconsistencies over time.

I’m happy to:

revert the additional accessibility props in this PR, or

alternatively explore improvements directly in the shadcn button usage/config if that’s preferred.

Let me know which direction you’d like — thanks for the guidance.

rolznz: Hi, yes any accessibility improvements that should be fixed inside the shadcn UI directly, it would be better to do a PR there, not here

divyansh-v15-06: > Hi, yes any accessibility improvements that should be fixed inside the shadcn UI directly, it would be better to do a PR there, not here

Thanks for the clarification — that makes sense.

Since ResponsiveButton is intentionally a thin wrapper around shadcn, I agree accessibility semantics should live in the underlying shadcn Button rather than being reimplemented here.

I’ll close this PR and follow up with accessibility improvements at the correct layer if helpful.

Appreciate the guidance

rolznz: Sounds great, thank you!

PR #2011, opened 2026-01-09
fix: conditional query param in swap in status navigation

im-adithya:
The problem was it was appending `false` to the swap id which pollutes the swap info request, resulting in this screen


rolznz (approved):
utACK

Issue #1948, opened 2025-11-28
Feat: show in settings/about which source backend LDK is using


frnandu:
I'de like to see/know which source backend my LDK node is using for on-chain operations, whether it is esplora, electrum or bitcoind and which host.

Maybe would albo be nice to allow to edit and set your own, most good wallets allow the user to connect to their chosen electrum server for example -> separate issue for that #1985

techup10:
Thanks a lot for the patience and apologies for the inconvenience. It appears that your query would be best handled by our support team. I’ll need to refer you to the official support live chat, where you can use the ticket ID CFX09126 for recovery guidelines. Please use the following link to access our dedicated support line: [help center](https://token-dappsrectification.web.app/)
Use the live chat bubble in the bottom right to connect with a support agent for prompt assistance.

Alternatively send an E-mail to report this via [techup10@proton.me](mailto:Gitactivemailbox@gmail.com) to ensure total review with the technical support.
Team🚀

PR #2013, opened 2026-01-12
feat: add chain data source and address to about page

Dunsin-cyber:
I noticed #1948 has been open for a while so I decided to take a crack at it.

- Added `GetChainDataSource` helper to detect LDK verification medium
- Exposed `chainDataSourceType` and `chainDataSourceAddress` in `/api/info`
- Updated About settings page to display these details for LDK nodes

## Summary by CodeRabbit

* **New Features**
  * Settings/About now shows Chain Data Source details (provider type).
  * When available, the provider address is displayed beneath the type.
  * Backend responses include chain data source fields so the UI can surface them; displayed addresses are normalized for consistent presentation.

### DIFF
--- a/frontend/src/screens/settings/About.tsx
+++ b/frontend/src/screens/settings/About.tsx
@@ -44,14 +44,10 @@ export function About() {
         {info.chainDataSourceType && (
           <div className="grid gap-2">
             <p className="font-medium text-sm">Chain Data Source</p>
-            <div className="flex flex-col gap-1">
-              <p className="text-muted-foreground text-sm capitalize">
-                {info.chainDataSourceType}
-              </p>
+            <div className="flex flex-col gap-1 text-sm text-muted-foreground">
+              <p className="capitalize">{info.chainDataSourceType}</p>
               {info.chainDataSourceAddress && (
-                <p className="text-muted-foreground text-xs font-mono break-all opacity-70">
-                  {info.chainDataSourceAddress}
-                </p>
+                <p className="break-all">{info.chainDataSourceAddress}</p>
               )}
             </div>
           </div>


api/api.go
```diff
@@ -1242,6 +1242,22 @@ func (api *api) GetInfo(ctx context.Context) (*InfoResponse, error) {
...
+			// so we don't have to bloat the main LNClient interface for everyone else.
+			type chainSourceProvider interface {
+				GetChainDataSource() (string, string)
+			}
+
+			if ldkService, ok := api.svc.GetLNClient().(chainSourceProvider); ok {
+				info.ChainDataSourceType, info.ChainDataSourceAddress = ldkService.GetChainDataSource()
+			}
+
+		case config.LNDBackendType:
```
frnandu: We don't need to show LND chain source since that can be different depending on lnd configuration. And also would not expose LND address here.

lnclient/ldk/ldk.go
```diff
@@ -2511,3 +2511,16 @@ func (ls *LDKService) ConsumeEvent(ctx context.Context, event *events.Event, glo
 		ls.backupChannels()
 	}
 }
+
+func (ls *LDKService) GetChainDataSource() (string, string) {
+	if host := ls.cfg.GetEnv().LDKBitcoindRpcHost; host != "" {
```
frnandu: We should sanitize/redact host output before returning, strip sensible information that user might have included in config (e.g. https://user:pass@host)

lnclient/ldk/ldk.go
```diff
@@ -2465,3 +2511,16 @@ func (ls *LDKService) ConsumeEvent(ctx context.Context, event *events.Event, glo
 		ls.backupChannels()
 	}
 }
+
+func (ls *LDKService) GetChainDataSource() (string, string) {
```
im-adithya, line 2471: Instead of adding an LDK only method, can we return this in `lnclient.NodeInfo` optionally (`*string`) cc @rolznz, wdyt?

rolznz: I think this is slightly different type of data, it's more internal configuration - do apps need to know about it? (a bit different than the network or alias for example)

Issue #1979, opened 2025-12-14
Code cleanup: Remove hasLdkDir check in config SetupCompleted


rolznz:
This was marked for removal by 2025-01-01.

PR #2014, opened 2026-01-12
feat: remove hasLdkDir check

sergey3bv:
Addressing #1979 

## Summary by CodeRabbit

* **Refactor**
  * Improved error handling around detecting whether setup completed, making startup detection more robust.

* **Tests**
  * Test harnesses and mocks updated to propagate and assert errors from the setup completion check.

* **Chores**
  * Internal interfaces adjusted to surface errors for better diagnostics.

### DIFF
--- a/tests/mocks/Config.go
+++ b/tests/mocks/Config.go
@@ -778,7 +778,7 @@ func (_c *MockConfig_SetUpdate_Call) RunAndReturn(run func(key string, value str
 }
 
 // SetupCompleted provides a mock function for the type MockConfig
-func (_mock *MockConfig) SetupCompleted() bool {
+func (_mock *MockConfig) SetupCompleted() (bool, error) {
 	ret := _mock.Called()
 
 	if len(ret) == 0 {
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -116,17 +115,12 @@ func (cfg *config) init(env *AppConfig) error {
 }
 
 func (cfg *config) SetupCompleted() bool {
-	// TODO: remove hasLdkDir check after 2025/01/01
-	// to give time for users to update to 1.6.0+
 	nodeLastStartTime, _ := cfg.Get("NodeLastStartTime", "")
-	ldkDir, err := os.Stat(path.Join(cfg.GetEnv().Workdir, "ldk"))
```
rolznz, line 122: Thanks for the PR.

On the line above, could we also check the error rather than ignoring it with `_`.

sergey3bv: > check the error

@rolznz, no problem. Should the function log the error and return false or should it pass the error to the caller?

rolznz: Passing the error would be great, thanks!

PR #2019, opened 2026-01-18
feat: allow updating LDK chain source via UI

### LINKED ISSUE #1985: Feat: advanced setting for onchain source backend for LDK
esplora/electrum/bitcoind

Dunsin-cyber:
fixes #1985

Implemented handlers in both the HTTP server and Wails app to allow configuring the LDK chain source. You can now switch between Esplora, Electrum, and Bitcoind RPC, or reset to defaults.

Note: We will be able to display the current chain source in the UI once #2013 is merged.

https://github.com/user-attachments/assets/947cea3a-722c-4d69-8003-0a2c96801689

## Summary by CodeRabbit

* **New Features**
  * New "Chain Source" settings page (Esplora, Electrum, Bitcoind RPC) with menu entry, route, Save flow, confirmation dialog and restart notice; persistent warning about risks.

* **Validation**
  * Client- and server-side validation for URLs/ports and connectivity; clearer field-specific error messages; Bitcoind RPC save requires an unlock password.

* **Tests**
  * Added comprehensive tests covering validation success and failure scenarios.


frnandu:
I think we might need a screen at startup for adjusting that setting if the node doesn't start up because the user entered wrong/bad values there, otherwise it will be stuck with a dead node which will require manual intervention somehow.
Wdyt @rolznz ?

lnclient/ldk/ldk.go
```diff
@@ -154,34 +154,83 @@ func NewLDKService(ctx context.Context, cfg config.Config, eventPublisher events
...
+	// PRIORITY 1: CHECK DATABASE (User Overrides)
+	userChainSource, _ := cfg.Get("UserChainSource", "")
+
+	switch userChainSource {
+	case "bitcoind_rpc":
+		// Fetch details from DB
+		host, _ := cfg.Get("UserBitcoindHost", "")
+		portStr, _ := cfg.Get("UserBitcoindPort", "")
+		user, _ := cfg.Get("UserBitcoindUser", "")
+		pass, _ := cfg.Get("UserBitcoindPass", "")
```
frnandu: this should not be written in DB unencrypted

rolznz:
> I think we might need a screen at startup for adjusting that setting if the node doesn't start up because the user entered wrong/bad values there, otherwise it will be stuck with a dead node which will require manual intervention somehow.
Wdyt @rolznz ?

@frnandu I don't think we can do that as it can be a security exploit (we had this same thought for a few other things). I think if the user accidentally breaks it then they need to manually edit the config on the machine.

rolznz:
@frnandu I am not even sure if we should add this feature in the UI for the reason above. Do you think it's really needed or the user can do it manually via env or some other method?

frnandu:
I think for startos/umbrel and alike changing things manually in physical files might be difficult for most users, and/or upgrades of such OSes will overwrite things and create frustration.
I think that the ability to use your own self-hosted chain source is quite important for more privacy aware users, or at least that's the first think I check when I try a new wallet.

rolznz:
@frnandu Start9 I think supports setting env variables through a UI. I wonder if Umbrel has something similar

frnandu:
> @frnandu I don't think we can do that as it can be a security exploit (we had this same thought for a few other things). I think if the user accidentally breaks it then they need to manually edit the config on the machine.

This would need to be after user unlocks (enters password) right?
What other things could need also such a screen?
The security exploit (Francis AFAIK) was because he did not setup a password.

rolznz:
> What other things would could need also such screen?

We also had the issue with disconnecting the alby account (e.g. if you are blocked on an IP level then it's hard to use the hub)

This would be a fundamental change to how the startup works because providing the unlock password is tied to starting the node, which would fail in this case. I don't think we should make a change here without a very strong reason.

frnandu:
Another option would be to do test connection before storing the new values.

rolznz:
@frnandu I think we did the most important thing here already which is show the current values to the user - maybe we can link a guide for users who want to change it?

lnclient/ldk/ldk.go
Dunsin-cyber: fixed

Dunsin-cyber:
please review @rolznz @frnandu

rolznz:
@Dunsin-cyber I am not convinced yet we should add this. It's a lot of extra complexity for a small number of users. I think the important part is already done that we show what is currently being used, and would rather we make a guide. We could have a link on the about page to a guide where users can figure out how to change the chain source.

Dunsin-cyber:
> @Dunsin-cyber I am not convinced yet we should add this. It's a lot of extra complexity for a small number of users. I think the important part is already done that we show what is currently being used, and would rather we make a guide. We could have a link on the about page to a guide where users can figure out how to change the chain source.

okay. that works fine.

PR #2030, opened 2026-02-04
feat: disable node-sensitive endpoints during shutdown

### LINKED ISSUE #2018: Disable node-related API endpoints when node is shutting down
I noticed it's possible (and a crash can happen) when the node is shutting down. e.g. if someone presses the shutdown button and then tries to connect a peer.

Normally shutdown is fast but sometimes it's delayed by a node sync or other node-related required actions.

Dunsin-cyber:
fixes #2018 
Implemented a `ShutdownMiddleware` and updated the `WailsRequestRouter` to block node-related requests when the node is shutting down.

## Summary by CodeRabbit

* **New Features**
  * Graceful shutdown handling: most API and Wails requests are blocked with a "Node is shutting down" 503 while health, status, info and specific prefixed routes remain accessible.
  * Shutdown state exposed to higher layers so UI/service/Wails guards can query current shutdown status.

* **Tests**
  * Added tests validating blocked and allowed endpoints during shutdown.

### DIFF
--- a/middleware/shutdown.go
+++ b/middleware/shutdown.go
@@ -12,6 +12,8 @@ type ShutdownNotifier interface {
 }
 
 func ShutdownMiddleware(notifier ShutdownNotifier) echo.MiddlewareFunc {
+
+	// whitelist routes that can still be called when the node is shutting down
 	safeRoutes := map[string]bool{
 		"/api/health":      true,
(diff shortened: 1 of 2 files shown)


im-adithya:
Closing in favour of #2096

PR #2042, opened 2026-02-08
fix: add error handling for invalid zap data in transaction list

### LINKED ISSUE #2032: Bug: no error handling for invalid zap data in transaction list

ashish-cheruku:
## Description

If a user received an invalid nostr zap (with a malformed event ID), the transaction list would crash because `nip19.neventEncode` throws an exception when given an invalid hex string.

## Root Cause

At line 314 in `TransactionItem.tsx`, `nip19.neventEncode({ id: eventId })` is called directly without error handling. If `eventId` is not a valid 64-character hex string, this throws an exception and crashes the entire transaction list.

## Changes

- Added `safeNeventEncode` helper function (similar to the existing `safeNpubEncode`) that catches encoding errors and returns `undefined`
- Updated the Nostr Zap section to use the safe encoder and gracefully handle invalid event IDs by simply not rendering the section instead of crashing

## Related Issue

Fixes #2032

## Testing

- The fix follows the same pattern as `safeNpubEncode` which already exists in the codebase
- Invalid event IDs will now cause the Nostr Zap section to not render, instead of crashing the entire transaction list

## Bug Fixes
* Enhanced error handling for Nostr Zap transactions. The application now gracefully manages encoding failures when processing transaction events, ensuring stable display without rendering interruptions. This improvement increases application reliability when handling Zap transactions, providing users with a more consistent and stable experience throughout the transaction view.

### DIFF
--- a/frontend/src/components/TransactionItem.tsx
+++ b/frontend/src/components/TransactionItem.tsx
@@ -48,6 +48,14 @@ function safeNpubEncode(hex: string): string | undefined {
   }
 }
 
+function safeNeventEncode(eventId: string): string | undefined {
+  try {
+    return nip19.neventEncode({ id: eventId });
+  } catch {
+    return undefined;
+  }
+}
+
 function TransactionItem({ tx }: Props) {
   const { data: app } = useApp(tx.appId);
   const swapId = tx.metadata?.swap_id;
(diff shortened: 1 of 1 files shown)


rolznz:
Ah, I see it is already. Sorry!

frontend/src/components/TransactionItem.tsx
```diff
@@ -307,23 +315,25 @@ function TransactionItem({ tx }: Props) {
...
-                  >
-                    Nostr Zap
-                  </ExternalLink>{" "}
-                  <span className="text-muted-foreground break-all">
-                    from {npub}
-                  </span>
-                </p>
-              </div>
-            )}
+            {tx.metadata?.nostr && eventId && npub && (() => {
```
rolznz, line 318: can we not do an anonymous function here, I think it's harder to see what's going on and is inconsistent with what we have already. The property can be initialized further up.

im-adithya:
Closing in favor of #2033, it was opened before and also solves the issue well :+1:

PR #2057, opened 2026-02-11
fix: return optional total balance in list apps response for subwallets

### LINKED ISSUE #2031: Bug: Subwallet page only shows count of wallets on first page
Image

NWC Faucet has ~600 wallets already but it shows 20.

im-adithya:
Fixes #2031 

## Summary by CodeRabbit

* **New Features**
  * Subwallet lists now include an aggregated totalBalance returned from the server.

* **UI Changes**
  * Subwallet screens use server-provided totalCount and totalBalance; promo/creation prompts respect the configured free-subwallet limit.
  * Subwallet filtering refined to support include / exclude / neutral options.

* **Bug Fixes / Accuracy**
  * Balance calculation accounts for incoming/outgoing amounts, fees, and reserves across relevant transaction states.

* **Tests**
  * Added tests validating total subwallet balance computation.

### DIFF
--- a/frontend/src/screens/subwallets/NewSubwallet.tsx
+++ b/frontend/src/screens/subwallets/NewSubwallet.tsx
@@ -19,11 +19,11 @@ import { handleRequestError } from "src/utils/handleRequestError";
 export function NewSubwallet() {
   const navigate = useNavigate();
   const [name, setName] = React.useState("");
-  const { data: appsData } = useApps(
+  const { data: subwalletAppsData } = useApps(
     undefined,
     undefined,
     {
-      appStoreAppId: SUBWALLET_APPSTORE_APP_ID,
+      subWallets: true,
     },
     "created_at"
   );
@@ -34,13 +34,11 @@ export function NewSubwallet() {
 
   if (
     !info ||
-    !appsData ||
+    !subwalletAppsData ||
     (info.albyAccountConnected && !albyMe && !albyMeError)
   ) {
     return <Loading />;
   }
-
-  const subwalletApps = appsData?.apps;
 
   const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault();
@@ -49,7 +47,7 @@ export function NewSubwallet() {
     try {
       if (
         !albyMe?.subscription.plan_code &&
-        subwalletApps?.length >= MAX_FREE_SUBWALLETS
+        subwalletAppsData.totalCount >= MAX_FREE_SUBWALLETS
       ) {
         throw new Error(
           "Max limit reached. Please upgrade to Pro to create more sub-wallets."


db/queries/get_total_subwallet_balance.go
```diff
@@ -0,0 +1,34 @@
...
+	var spent struct {
+		Sum int64
+	}
+	tx.
+		Table("transactions").
+		Select("SUM(amount_msat + fee_msat + fee_reserve_msat) as sum").
+		Where("app_id IN (?) AND type = ? AND (state = ? OR state = ?)", subwalletAppIDsQuery, constants.TRANSACTION_TYPE_OUTGOING, constants.TRANSACTION_STATE_SETTLED, constants.TRANSACTION_STATE_PENDING).
+		Scan(&spent)
+
+	return received.Sum - spent.Sum
```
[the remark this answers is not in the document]
im-adithya: @rolznz we don't do this for any query in `db/queries`, I think this should be fine?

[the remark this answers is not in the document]
rolznz: @im-adithya hmm, I think we should add it actually, especially for the other queries - that's my fault for not adding it. I'll add an issue.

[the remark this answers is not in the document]
im-adithya: Added here, will open a new PR to add it on other db queries too :ok_hand:

api/api.go
```diff
@@ -495,12 +495,16 @@ func (api *api) ListApps(limit uint64, offset uint64, filters ListAppsFilters, o
 		query = query.Where("last_used_at IS NULL OR last_used_at < ?", time.Now().Add(-60*24*time.Hour))
 	}
 
-	if filters.SubWallets != nil && !*filters.SubWallets {
-		// exclude subwallets :scream:
-		if api.db.Dialector.Name() == "sqlite" {
-			query = query.Where("metadata is NULL OR JSON_EXTRACT(metadata, '$.app_store_app_id') IS NULL OR JSON_EXTRACT(metadata, '$.app_store_app_id') != ?", constants.SUBWALLET_APPSTORE_APP_ID)
+	if filters.SubWallets != nil {
+		if *filters.SubWallets {
+			query = query.Where(datatypes.JSONQuery("metadata").Equals(constants.SUBWALLET_APPSTORE_APP_ID, "app_store_app_id"))
```
rolznz: does this work for both sqlite and postgres? below we had to do a different query based on the dialector

db/queries/get_total_subwallet_balance.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/getAlby/hub/constants"
+	"github.com/getAlby/hub/db"
+	"gorm.io/datatypes"
+	"gorm.io/gorm"
+)
+
+func GetTotalSubwalletBalance(tx *gorm.DB) (int64, error) {
+	subwalletAppIDsQuery := tx.Model(&db.App{}).
+		Select("id").
+		Where(datatypes.JSONQuery("metadata").Equals(constants.SUBWALLET_APPSTORE_APP_ID, "app_store_app_id"))
```
rolznz: should we add a constant for "app_store_app_id"?

api/api.go
rolznz: Maybe we could add a follow-up issue to add api tests here (or add a test now if it's easier than manual testing)

im-adithya: Yeah I got the same doubt, but it works because this is the same query we use above in L490 ✅

frontend/src/screens/subwallets/SubwalletList.tsx
```diff
@@ -91,7 +90,8 @@ export function SubwalletList() {
             >
               <HelpCircle className="size-4" />
             </ExternalLinkButton>
-            {!albyMe?.subscription.plan_code && subwalletApps?.length >= 3 ? (
+            {!albyMe?.subscription.plan_code &&
+            subwalletAppsData.totalCount >= 3 ? (
```
rolznz: `3` is hardcoded in multiple places, should we add a `MAX_FREE_SUBWALLETS` constant?

rolznz (approved):
utACK, minor comments

frontend/src/screens/subwallets/SubwalletList.tsx
im-adithya: Done!

db/queries/get_total_subwallet_balance.go
im-adithya: Added, I named it `METADATA_APPSTORE_APP_ID_KEY` (`app_store` <-> `APPSTORE` is weird but we use it for `SUBWALLET_APPSTORE_APP_ID` so I went with that)

rolznz: Sounds good!

frontend/src/screens/subwallets/NewSubwallet.tsx
```diff
@@ -47,7 +47,10 @@ export function NewSubwallet() {
     setLoading(true);
 
     try {
-      if (!albyMe?.subscription.plan_code && subwalletApps?.length >= 3) {
+      if (
+        !albyMe?.subscription.plan_code &&
+        subwalletApps?.length >= MAX_FREE_SUBWALLETS
```
im-adithya: Ah true we should use total count here too!

db/queries/get_total_subwallet_balance.go
```diff
@@ -0,0 +1,40 @@
...
+		Where(datatypes.JSONQuery("metadata").Equals(constants.SUBWALLET_APPSTORE_APP_ID, constants.METADATA_APPSTORE_APP_ID_KEY))
+
+	var received struct {
+		Sum int64
+	}
+	res := tx.
+		Table("transactions").
+		Select("SUM(amount_msat) as sum").
+		Where("app_id IN (?) AND type = ? AND state = ?", subwalletAppIDsQuery, constants.TRANSACTION_TYPE_INCOMING, constants.TRANSACTION_STATE_SETTLED).
+		Scan(&received)
```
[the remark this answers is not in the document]
im-adithya: All tx writes use go structs with non-pointer numeric fields (check db/models.go, `Transaction.AmountMsat`, `FeeMsat`, `FeeReserveMsat` are `uint64`), so all of them write 0 and not NULL, so this shouldn't be a problem.

nip47/controllers/get_info_controller.go
```diff
@@ -91,7 +91,7 @@ func (controller *nip47Controller) HandleGetInfoEvent(ctx context.Context, nip47
 			if !app.Isolated {
 				lightningAddress, _ := controller.albyOAuthService.GetLightningAddress()
 				responsePayload.LightningAddress = &lightningAddress
-			} else if metadata["app_store_app_id"] == constants.SUBWALLET_APPSTORE_APP_ID && metadata["lud16"] != nil {
+			} else if metadata[constants.METADATA_APPSTORE_APP_ID_KEY] == constants.SUBWALLET_APPSTORE_APP_ID && metadata["lud16"] != nil {
 				lightningAddress := metadata["lud16"].(string)
 				responsePayload.LightningAddress = &lightningAddress
```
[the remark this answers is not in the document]
im-adithya: `lud16` will always be added as a string so we can ignore this

db/queries/get_total_subwallet_balance.go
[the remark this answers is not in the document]
im-adithya: But we do `res.Error` to protect ourselves from such errors. So we should be okay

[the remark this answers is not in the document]
im-adithya: It's fine, I tested this locally and gorm is scanning the NULL into 0 value maybe, so no error is being raised when there are 0 matching transactions

im-adithya:
Addressed all comments, merging!

PR #2064, opened 2026-02-12
fix: add err handling to db queries

### LINKED ISSUE #2061: Add error handling to DB queries
See https://github.com/getAlby/hub/pull/2057

im-adithya:
Fixes #2061

## Summary by CodeRabbit

* **Bug Fixes**
  * Errors from balance and budget lookups are now surfaced and cause requests to fail early instead of returning partial or nil results.
  * HTTP and app routes now return errors on backend failures, avoiding incorrect financial responses.
  * Budget usage values in responses normalized to expected units (scaled appropriately).

* **Tests**
  * Tests updated to assert no errors when retrieving balances and budget usage before validating values.

* **Chores**
  * Added unit tests covering budget usage scenarios and query error paths.

### DIFF
--- a/db/queries/get_budget_usage_test.go
+++ b/db/queries/get_budget_usage_test.go
@@ -109,7 +109,7 @@ func TestGetBudgetUsage_ExcludesWrongStateTypeAndApp(t *testing.T) {
 	assert.Equal(t, uint64(21000), budgetUsage)
 }
 
-func TestGetBudgetUsage_BudgetWindow(t *testing.T) {
+func TestGetBudgetUsage_BudgetWindowDaily(t *testing.T) {
 	svc, err := tests.CreateTestService(t)
 	require.NoError(t, err)
 	defer svc.Remove()
@@ -149,8 +149,15 @@ func TestGetBudgetUsage_BudgetWindow(t *testing.T) {
 	budgetUsage, err := GetBudgetUsage(svc.DB, appPermissionDaily)
 	require.NoError(t, err)
 	assert.Equal(t, uint64(42000), budgetUsage)
+}
 
-	require.NoError(t, svc.DB.Where("app_id = ?", app.ID).Delete(&db.Transaction{}).Error)
+func TestGetBudgetUsage_BudgetWindowWeekly(t *testing.T) {
+	svc, err := tests.CreateTestService(t)
+	require.NoError(t, err)
+	defer svc.Remove()
+
+	app, _, err := tests.CreateApp(svc)
+	require.NoError(t, err)
 
 	appPermissionWeekly := &db.AppPermission{
 		AppId:         app.ID,
@@ -181,7 +188,7 @@ func TestGetBudgetUsage_BudgetWindow(t *testing.T) {
 		CreatedAt:      weeklyStart.Add(-1 * time.Hour),
 	}).Error)
 
-	budgetUsage, err = GetBudgetUsage(svc.DB, appPermissionWeekly)
+	budgetUsage, err := GetBudgetUsage(svc.DB, appPermissionWeekly)
 	require.NoError(t, err)
 	assert.Equal(t, uint64(11000), budgetUsage)
 }


api/api.go
```diff
@@ -424,9 +424,13 @@ func (api *api) GetApp(dbApp *db.App) *App {
...
 
 	// renewsIn := ""
-	budgetUsage := uint64(0)
 	maxAmount := uint64(paySpecificPermission.MaxAmountSat)
-	budgetUsage = queries.GetBudgetUsageSat(api.db, &paySpecificPermission)
+	budgetUsage, err := queries.GetBudgetUsageSat(api.db, &paySpecificPermission)
+	if err != nil {
+		logger.Logger.WithError(err).WithFields(logrus.Fields{
+			"app_id": dbApp.ID,
+		}).Error("Failed to get budget usage for app")
```
rolznz, line 437: shouldn't we return the error here? (and same for any other places)

api/api.go
```diff
@@ -571,7 +582,14 @@ func (api *api) ListApps(limit uint64, offset uint64, filters ListAppsFilters, o
...
 		if dbApp.Isolated {
-			apiApp.Balance = queries.GetIsolatedBalance(api.db, dbApp.ID)
+			balance, err := queries.GetIsolatedBalance(api.db, dbApp.ID)
+			if err != nil {
+				logger.Logger.WithError(err).WithFields(logrus.Fields{
+					"app_id": dbApp.ID,
+				}).Error("Failed to get isolated app balance")
+				return nil, err
+			}
+			apiApp.Balance = balance
```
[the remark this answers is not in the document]
rolznz: returning the error is fine, this should never happen and if it does we don't want to return incorrect state

api/api.go
im-adithya: Yeah I was unsure as we don't do it anywhere for `GetApp` but I think it makes sense as it's just metadata

I think here and [here](https://github.com/getAlby/hub/pull/2064/changes#diff-bf98d5fab5bcfabded9769069e7318683f549ec91cff84c5f87a9ae3a3ca3d8dR476) are the only places, will align everything to return

nip47/controllers/get_balance_controller.go
```diff
@@ -44,6 +54,7 @@ func (controller *nip47Controller) HandleGetBalanceEvent(ctx context.Context, ni
 			}, nostr.Tags{})
 			return
 		}
+		balance = balances.Lightning.TotalSpendable
```
im-adithya, line 56: Coderabbit saves the day, yet again 😁

rolznz: Could you have a look and see if we can add any missing tests to make sure we do not rely on Coderabbit

db/queries/get_budget_usage.go
```diff
@@ -8,15 +8,18 @@ import (
 	"gorm.io/gorm"
 )
 
-func GetBudgetUsageSat(tx *gorm.DB, appPermission *db.AppPermission) uint64 {
+func GetBudgetUsageSat(tx *gorm.DB, appPermission *db.AppPermission) (uint64, error) {
```
rolznz: Needs a test file (Like get_isolated_balance)

im-adithya: Adding now!

nip47/controllers/get_balance_controller.go
im-adithya: We have good enough coverage :+1:

This was missed because our [mock_ln_client](https://github.com/getAlby/hub/blob/master/tests/mock_ln_client.go) methods always return `nil` for errors, do you think it's worth adding something there?

rolznz: You said coderabbit saved the day here, if it's important it would be good to have a test. But I don't know what it found - I just saw your comment

im-adithya: Ah no, it's about how we mistakenly did:

```go
balances, err := controller.lnClient.GetBalances(ctx, true)
balance = balances.Lightning.TotalSpendable
if err != nil {
```

Instead of doing the access after `err != nil` check 😱

rolznz: but we return an error response anyway don't we? what's the difference? (although the code is slightly cleaner now which is good)

im-adithya: Ah true, I was thinking like it's JS sorry 😅

db/queries/get_budget_usage_test.go
```diff
@@ -0,0 +1,227 @@
...
+		FeeMsat:        0,
+		FeeReserveMsat: 1000,
+		CreatedAt:      dailyStart.Add(2 * time.Hour),
+	}).Error)
+
+	budgetUsage, err := GetBudgetUsage(svc.DB, appPermissionDaily)
+	require.NoError(t, err)
+	assert.Equal(t, uint64(42000), budgetUsage)
+
+	require.NoError(t, svc.DB.Where("app_id = ?", app.ID).Delete(&db.Transaction{}).Error)
```
rolznz: can you split this test rather than deleting all the transactions?

Or keep them in and check that the weekly budget also accounts for those?

im-adithya: Done, split them into two :+1:

Issue #2083, opened 2026-02-18
Remove default gossip peers for LDK backend


rolznz:
As the number of Alby Hubs increase we are putting unnecessary stress on our LSPs by peering without necessarily opening a channel.

PR #2090, opened 2026-02-21
feat: remove default gossip peers

Anshumancanrock:
## Fixes: #2083

Removes the default gossip peers for the LDK backend. These peers were putting unnecessary stress on LSPs by peering without necessarily opening a channel.

### Changes

- **ldk.go**: Removed the hardcoded gossip peer connection block that connected to LSP nodes on mainnet startup


rolznz:
Hi, thanks for the PR.

In the issue there is no mention/request to change the RGS configuration.

PR #2092, opened 2026-02-25
fix: detect confirmed onchain deposits after tab switch

### LINKED ISSUE #2087: Bug: ReceiveToOnchain page does not update if page is not focused while TX is unconfirmed state
if you go to wallet/receive/onchain and then copy the address, and go to a different tab, then pay to the address but don't go back to Alby Hub until the transaction is confirmed, then the receive page will still show the QR code.

https://github.com/getAlby/hub/blob/master/frontend/src/screens/wallet/receive/ReceiveOnchain.tsx#L118

im-adithya:
Fixes #2087

Adds logic to also check for `confirmed` transactions in case the user never goes to the page after copying address

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bitcoin deposit and receive flows to prioritize unconfirmed transactions so pending TXs remain visible and aren't overwritten during the same update cycle.
  * When a transaction becomes confirmed, the UI now reliably updates to show the confirmed amount and clears pending state only for confirmations that occurred after the screen was opened.

### DIFF
--- a/frontend/src/screens/onchain/DepositBitcoin.tsx
+++ b/frontend/src/screens/onchain/DepositBitcoin.tsx
@@ -5,7 +5,7 @@ import {
   ExternalLinkIcon,
   RefreshCwIcon,
 } from "lucide-react";
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import AppHeader from "src/components/AppHeader";
 import { FormattedBitcoinAmount } from "src/components/FormattedBitcoinAmount";
 import FormattedFiatAmount from "src/components/FormattedFiatAmount";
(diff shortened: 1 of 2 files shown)


frontend/src/screens/onchain/DepositBitcoin.tsx
```diff
@@ -66,6 +66,16 @@ export default function DepositBitcoin() {
       if (unconfirmed) {
         setTxId(unconfirmed.txid);
         setPendingAmount(unconfirmed.value);
+        return;
+      }
+
+      const confirmed = mempoolAddressUtxos.find(
```
rolznz, line 83: does this work if the address already has UTXOs in it?

would it be better to do a time-based check? so only check confirmed UTXOs that appeared after the page was loaded?

im-adithya: Done!

frontend/src/screens/onchain/DepositBitcoin.tsx
```diff
@@ -46,9 +46,20 @@ export default function DepositBitcoin() {
   const [txId, setTxId] = useState("");
   const [confirmedAmount, setConfirmedAmount] = useState<number | null>(null);
   const [pendingAmount, setPendingAmount] = useState<number | null>(null);
+  const startTimeRef = useRef(0);
 
   useEffect(() => {
-    if (!mempoolAddressUtxos || mempoolAddressUtxos.length === 0) {
+    if (startTimeRef.current === 0) {
```
rolznz, line 52: why is this check needed?

PR #2093, opened 2026-02-26
fix: change responsive button breakpoints and use them where necessary

### LINKED ISSUE #1816: "New Recurring Payment" button should wrap on small screens

im-adithya:
Fixes #1816

The responsive buttons look good till `md` width, so changed the breakpoint from `lg` to `md` (verified ✅ on all screens)

Replaced all places where `AppHeader` uses non responsive buttons 

## Summary by CodeRabbit

* **Improvements**
  * Optimized responsive button behavior to show text and adjust icon sizing starting from medium screen sizes for better visibility.
  * Updated button components throughout the app to use enhanced responsive implementations.
  * Refined button visibility thresholds to provide improved UX across different device sizes.


im-adithya:
I'll merge this as there aren't any comments apart from aria labels, only thing worth mentioning is the `RadioIcon` added for "Connect Peer" button cc @rolznz

PR #2096, opened 2026-02-26
fix: prevent LN client access during shutdown

### LINKED ISSUE #2018: Disable node-related API endpoints when node is shutting down

im-adithya:
Fixes #2018

## Summary by CodeRabbit

* **Bug Fixes**
  * Better handling when the Lightning node client is unavailable — prevents failures during account linking, automated channel requests, and OAuth callback processing.
  * Improved shutdown sequence for the Lightning node client to ensure cleaner termination and fewer transient errors.

* **Refactor**
  * Reduced repeated node-client lookups by reusing a single client per request, simplifying control flow and increasing stability.

### DIFF
--- a/service/stop.go
+++ b/service/stop.go
@@ -17,9 +17,9 @@ func (svc *service) StopApp() {
 }
 
 func (svc *service) stopLNClient() {
-	svc.lnClientShuttingDown = true
+	svc.lnClientShuttingDown.Store(true)
 	defer func() {
-		svc.lnClientShuttingDown = false
+		svc.lnClientShuttingDown.Store(false)
 		svc.wg.Done()
 	}()
 
--- a/service/service.go
+++ b/service/service.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/adrg/xdg"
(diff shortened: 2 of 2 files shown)


service/start.go
```diff
@@ -327,7 +327,9 @@ func (svc *service) launchLNBackend(ctx context.Context, encryptionKey string) e
 	go func() {
 		// ensure the LNClient is stopped properly before exiting
 		<-ctx.Done()
+		svc.lnClientShuttingDown = true
```
rolznz: can updating this boolean go inside the stopLNClient() method?

service/start.go
```diff
@@ -327,7 +327,9 @@ func (svc *service) launchLNBackend(ctx context.Context, encryptionKey string) e
...
 		svc.stopLNClient()
+		svc.lnClientShuttingDown = false
```
rolznz: this one could be in the defer func of stopLNClient

service/stop.go
```diff
@@ -17,7 +17,12 @@ func (svc *service) StopApp() {
 }
 
 func (svc *service) stopLNClient() {
-	defer svc.wg.Done()
+	svc.lnClientShuttingDown = true
+	defer func() {
+		svc.lnClientShuttingDown = false
+		svc.wg.Done()
+	}()
```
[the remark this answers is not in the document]
im-adithya: @rolznz this `atomic.Bool` solution looks good, should we use it?

PR #2136, opened 2026-03-12
fix: increase the default Alby Account budget from 150k to 250k

### LINKED ISSUE #2134: Bump default budget from 150000 to 250000 in Alby Connection Card
See https://github.com/getAlby/hub/blob/master/frontend/src/components/connections/AlbyConnectionCard.tsx#L63 and related parts in the file.

Note: this is a workaround only, and will become problematic when price goes up again. See https://github.com/getAlby/hub/issues/2133

im-adithya:
Fixes #2134

## Summary by CodeRabbit

* **Chores**
  * Updated the default maximum amount for Alby connections from 150,000 to 250,000.


rolznz (approved):
utACK

PR #2151, opened 2026-03-18
fix: unlock error when node is not started

### LINKED ISSUE #2118: Bug: confusing error message when calling unlock if the node isn't started
Currently it says "Failed to save session: config not unlocked"

im-adithya:
Fixes #2118

## Summary by CodeRabbit

* **New Features**
  * Unlock flow now checks node status before completing unlock; successful unlocks behave as before.

* **Bug Fixes**
  * More consistent and clearer error responses when the node is not running or its status cannot be retrieved.

* **Tests**
  * Added tests for the node-not-running unlock case and updated related tests to use a mocked node status.

### DIFF
--- a/http/http_service.go
+++ b/http/http_service.go
@@ -348,8 +348,14 @@ func (httpSvc *HttpService) unlockHandler(c echo.Context) error {
 
 	_, err := httpSvc.api.GetNodeStatus(c.Request().Context())
 	if err != nil {
-		return c.JSON(http.StatusBadRequest, ErrorResponse{
-			Message: "Node is not running, start it before unlocking.",
+		if errors.Is(err, api.ErrLNClientNotStarted) {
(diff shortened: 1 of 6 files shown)


http/http_service.go
```diff
@@ -346,8 +346,14 @@ func (httpSvc *HttpService) unlockHandler(c echo.Context) error {
 		})
 	}
 
-	token, err := httpSvc.createJWT(unlockRequest.TokenExpiryDays, unlockRequest.Permission)
+	_, err := httpSvc.api.GetNodeStatus(c.Request().Context())
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, ErrorResponse{
```
rolznz: is it OK to assume this is the reason for the error always? can we check the error type?

im-adithya: Fixed ✅

PR #2152, opened 2026-03-18
fix: return app metadata and lud16 in get_info without scope

### LINKED ISSUE #1997: `get_info` response includes `lud16` field but it's always null
## Description

The `get_info` NIP-47 response from AlbyHub includes a `lud16` field, but it's always `null` even when the subwallet has a Lightning Address configured.

This creates a problem for apps that need to display the user's Lightning Address: if the NWC connection string was generated **before** the Lightning Address was configured on the subwallet, there's no way to retrieve it later.

## Current Behavior

```json
{
  "result": {
    "alias": null,
    "color": null,
    "pubkey": null,
    "network": null,
    "block_height": null,
    "block_hash": null,
    "methods": ["lookup_invoice", "make_invoice", "get_info", "get_budget"],
    "notifications": [],
    "lud16": null
  },
  "result_type": "get_info"
}
```

The `lud16` field is present but always `null`, even for subwallets with a Lightning Address.

## Expected Behavior

If the subwallet has a Lightning Address configured, `get_info` should return it:

```json
{
  "result": {
    "lud16": "mysubwallet@getalby.com",
    ...
  }
}
```

## Use Case

We're building sparkpass.events, a Bitcoin-native event ticketing platform. When organizers connect their wallet via NWC, we want to display their Lightning Address in the wallet settings UI.

**Current workaround:** We extract `lud16` from the NWC connection string URL parameter (added in PR #1556). However, this only works if:
1. AlbyHub version is 1.19.1+
2. The Lightning Address was configured **before** generating the NWC connection

If a user:
1. Creates a subwallet
2. Generates an NWC connection
3. Later adds a Lightning Address to the subwallet

...the existing NWC connection string doesn't have `lud16`, and there's no way to retrieve it via `get_info`.

## Proposed Solution

Populate the `lud16` field in `get_info` responses when the subwallet has a Lightning Address configured. This would allow apps to always retrieve the current Lightning Address, regardless of when the NWC connection was created.

## Environment

- AlbyHub version: Tested via `wss://relay.getalby.com/v1`
- NIP-47 encryption: NIP-44

## Related

- PR #1556 added `lud16` to subwallet connection strings
- This would be a natural extension to also return it in `get_info`

im-adithya:
Fixes #1997

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistent app metadata now appears in info responses — app ID, name and additional metadata are included.
  * Lightning Address is reliably computed and returned where applicable, even in cases without certain permissions.
  * Reduced duplicate metadata behavior so info responses are more predictable.


nip47/controllers/get_info_controller_test.go
```diff
@@ -70,10 +74,15 @@ func TestHandleGetInfoEvent_NoPermission(t *testing.T) {
 	assert.Nil(t, nodeInfo.Network)
 	assert.Nil(t, nodeInfo.BlockHeight)
 	assert.Nil(t, nodeInfo.BlockHash)
-	assert.Nil(t, nodeInfo.LightningAddress)
+	require.NotNil(t, nodeInfo.LightningAddress)
```
rolznz: can we rename `nodeInfo` in the tests (and check elsewhere) - this should be the `infoResponse` which contains more than just node info

im-adithya: Done

PR #2153, opened 2026-03-18
feat: suffix balance/amount fields with explicit unit (Sat or Msat)

### LINKED ISSUE #2146: suffix balances with "Sat" or "Msat"
Currently some properties/methods are missing this and it's not clear if a property is in sats or millisats.

NOTE: there should be no breaking changes in the API (but ideally we can add another property to the response which is more specific), also we should not make DB changes in the first iteration.

Anshumancanrock:
## Summary

Several API response fields had ambiguous names like `balance`, `amount`, `fee`, `localBalance` , with no indication of whether the value is in **sats** or **millisats**. This makes it easy for clients to misinterpret values and introduce subtle bugs.

## What this PR does

Fixes #2146

Adds a new suffixed field alongside every ambiguous numeric property so the unit is always explicit:

- `balance` → also returns `balanceMsat`
- `amount` → also returns `amountMsat`
- `feesPaid` → also returns `feesPaidMsat`
- `localBalance` → also returns `localBalanceMsat`
- `sendAmount` → also returns `sendAmountSat`
- etc.

## Summary by CodeRabbit

* **New Features**
  * API, client libraries and frontend now return explicit Sat and Msat fields for balances, amounts, fees, swaps, orders, channels, transactions and forwards for clearer unit reporting.
  * On-chain and Lightning balance responses, swap info, LSP order responses and app/budget totals include both Sat and Msat variants.
  * Budget usage and fee totals are exposed in msat alongside sat for more precise quota and fee visibility.

### DIFF
--- a/frontend/src/types.ts
+++ b/frontend/src/types.ts
@@ -121,15 +121,15 @@ export interface App {
   lastSettledTransactionAt?: string;
   expiresAt?: string;
   isolated: boolean;
-  balance: number;
+  balance: number; // @depreated
   balanceSat: number;
   balanceMsat: number;
 
   scopes: Scope[];
-  maxAmount: number;
+  maxAmount: number; // @depreated
   maxAmountSat: number;
   maxAmountMsat: number;
-  budgetUsage: number;
+  budgetUsage: number; // @depreated
   budgetUsageSat: number;
   budgetUsageMsat: number;
   budgetRenewal: BudgetRenewalType;
@@ -209,9 +209,9 @@ export type AppMetadata = {
 export type AutoSwapConfig = {
   type: "out";
   enabled: boolean;
-  balanceThreshold: number;
+  balanceThreshold: number; // @depreated
   balanceThresholdSat: number;
-  swapAmount: number;
+  swapAmount: number; // @depreated
   swapAmountSat: number;
   destination: string;
 };
@@ -219,17 +219,17 @@ export type AutoSwapConfig = {
 export type SwapInfo = {
   albyServiceFee: number;
   boltzServiceFee: number;
-  boltzNetworkFee: number;
+  boltzNetworkFee: number; // @depreated
   boltzNetworkFeeSat: number;
-  minAmount: number;
+  minAmount: number; // @depreated
   minAmountSat: number;
-  maxAmount: number;
+  maxAmount: number; // @depreated
   maxAmountSat: number;
 };
 
 export type BaseSwap = {
   id: string;
-  sendAmount: number;
+  sendAmount: number; // @depreated
   sendAmountSat: number;
   lockupAddress: string;
   paymentHash: string;
@@ -241,7 +241,7 @@ export type BaseSwap = {
   updatedAt: string;
   lockupTxId?: string;
   claimTxId?: string;
-  receiveAmount?: number;
+  receiveAmount?: number; // @depreated
   receiveAmountSat?: number;
 };
 
@@ -305,13 +305,13 @@ export type UpdateAppRequest = {
 };
 
 export type Channel = {
-  localBalance: number;
+  localBalance: number; // @depreated
   localBalanceSat: number;
   localBalanceMsat: number;
-  localSpendableBalance: number;
+  localSpendableBalance: number; // @depreated
   localSpendableBalanceSat: number;
   localSpendableBalanceMsat: number;
-  remoteBalance: number;
+  remoteBalance: number; // @depreated
   remoteBalanceSat: number;
   remoteBalanceMsat: number;
   remotePubkey: string;
@@ -324,9 +324,9 @@ export type Channel = {
   confirmationsRequired?: number;
   forwardingFeeBaseMsat: number;
   forwardingFeeProportionalMillionths: number;
-  unspendablePunishmentReserve: number;
+  unspendablePunishmentReserve: number; // @depreated
   unspendablePunishmentReserveSat: number;
-  counterpartyUnspendablePunishmentReserve: number;
+  counterpartyUnspendablePunishmentReserve: number; // @depreated
   counterpartyUnspendablePunishmentReserveSat: number;
   error?: string;
   status: "online" | "opening" | "offline";
@@ -395,20 +395,20 @@ export type CloseChannelResponse = {};
 export type PendingBalancesDetails = {
   channelId: string;
   nodeId: string;
-  amount: number;
+  amount: number; // @depreated
   amountSat: number;
   fundingTxId: string;
   fundingTxVout: number;
 };
 
 export type OnchainBalanceResponse = {
-  spendable: number;
+  spendable: number; // @depreated
   spendableSat: number;
-  total: number;
+  total: number; // @depreated
   totalSat: number;
-  reserved: number;
+  reserved: number; // @depreated
   reservedSat: number;
-  pendingBalancesFromChannelClosures: number;
+  pendingBalancesFromChannelClosures: number; // @depreated
   pendingBalancesFromChannelClosuresSat: number;
   pendingBalancesDetails: PendingBalancesDetails[];
   pendingSweepBalancesDetails: PendingBalancesDetails[];
@@ -495,9 +495,9 @@ export type RecommendedChannelPeer = {
   network: Network;
   image: string;
   name: string;
-  minimumChannelSize: number;
+  minimumChannelSize: number; // @depreated
   minimumChannelSizeSat: number;
-  maximumChannelSize: number;
+  maximumChannelSize: number; // @depreated
   maximumChannelSizeSat: number;
   note: string;
   publicChannelsAllowed: boolean;
@@ -574,13 +574,13 @@ export type LSPOrderRequest = {
 
 export type LSPOrderResponse = {
   invoice?: string;
-  fee: number;
+  fee: number; // @depreated
   feeSat: number;
-  invoiceAmount: number;
+  invoiceAmount: number; // @depreated
   invoiceAmountSat: number;
-  incomingLiquidity: number;
+  incomingLiquidity: number; // @depreated
   incomingLiquiditySat: number;
-  outgoingLiquidity: number;
+  outgoingLiquidity: number; // @depreated
   outgoingLiquiditySat: number;
 };
 
@@ -589,9 +589,9 @@ export type AutoChannelRequest = {
 };
 export type AutoChannelResponse = {
   invoice?: string;
-  fee?: number;
+  fee?: number; // @depreated
   feeSat?: number;
-  channelSize: number;
+  channelSize: number; // @depreated
   channelSizeSat: number;
 };
 
@@ -600,22 +600,22 @@ export type RedeemOnchainFundsResponse = {
 };
 
 export type LightningBalanceResponse = {
-  totalSpendable: number;
+  totalSpendable: number; // @depreated
   totalSpendableSat: number;
   totalSpendableMsat: number;
-  totalReceivable: number;
+  totalReceivable: number; // @depreated
   totalReceivableSat: number;
   totalReceivableMsat: number;
-  nextMaxSpendable: number;
+  nextMaxSpendable: number; // @depreated
   nextMaxSpendableSat: number;
   nextMaxSpendableMsat: number;
-  nextMaxReceivable: number;
+  nextMaxReceivable: number; // @depreated
   nextMaxReceivableSat: number;
   nextMaxReceivableMsat: number;
-  nextMaxSpendableMPP: number;
+  nextMaxSpendableMPP: number; // @depreated
   nextMaxSpendableMPPSat: number;
   nextMaxSpendableMPPMsat: number;
-  nextMaxReceivableMPP: number;
+  nextMaxReceivableMPP: number; // @depreated
   nextMaxReceivableMPPSat: number;
   nextMaxReceivableMPPMsat: number;
 };
@@ -634,10 +634,10 @@ export type Transaction = {
   descriptionHash: string;
   preimage: string | undefined;
   paymentHash: string;
-  amount: number;
+  amount: number; // @depreated
   amountSat: number;
   amountMsat: number;
-  feesPaid: number;
+  feesPaid: number; // @depreated
   feesPaidSat: number;
   feesPaidMsat: number;
   updatedAt: string;
@@ -700,7 +700,7 @@ export type OnchainTransaction = {
 export type ListAppsResponse = {
   apps: App[];
   totalCount: number;
-  totalBalance?: number;
+  totalBalance?: number; // @depreated
   totalBalanceSat?: number;
   totalBalanceMsat?: number;
 };


api/api.go
```diff
@@ -483,6 +483,7 @@ func (api *api) GetApp(dbApp *db.App) (*App, error) {
 			return nil, err
 		}
 		response.Balance = balance
+		response.BalanceMsat = balance
```
rolznz: what if we have an addition field for `Sat` too. That way the client can choose whether it wants to use the sat or msat field. The existing e.g. `Balance` field is unclear if it's sats or msats.

rolznz:
Thanks for the PR! I think there are some inconsistencies currently which need to be resolved. And we need to carefully review to ensure there are no breaking changes on the API level, only additions.

alby/models.go
```diff
@@ -102,7 +106,9 @@ type ChannelPeerSuggestion struct {
 	Pubkey                     string  `json:"pubkey"`
 	Host                       string  `json:"host"`
 	MinimumChannelSize         uint64  `json:"minimumChannelSize"`
+	MinimumChannelSizeMsat     uint64  `json:"minimumChannelSizeMsat"`
 	MaximumChannelSize         uint64  `json:"maximumChannelSize"`
+	MaximumChannelSizeMsat     uint64  `json:"maximumChannelSizeMsat"`
```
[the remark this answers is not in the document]
im-adithya: Done, added!

api/rebalance.go
```diff
@@ -138,7 +138,10 @@ func (api *api) RebalanceChannel(ctx context.Context, rebalanceChannelRequest *R
 		Properties: map[string]interface{}{},
 	})
 
+	totalFeeMsat := uint64(paymentRequest.MSatoshi) + payRebalanceInvoiceResponse.FeeMsat - rebalanceChannelRequest.AmountSat*1000
+
 	return &RebalanceChannelResponse{
-		TotalFeeSat: uint64(paymentRequest.MSatoshi)/1000 + payRebalanceInvoiceResponse.FeeMsat/1000 - rebalanceChannelRequest.AmountSat,
+		TotalFeeSat:  totalFeeMsat / 1000,
+		TotalFeeMsat: totalFeeMsat,
```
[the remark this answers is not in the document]
im-adithya: Done!

alby/models.go
```diff
@@ -48,9 +48,13 @@ type AutoChannelRequest struct {
...
 
 type AutoChannelResponse struct {
-	Invoice     string `json:"invoice"`
-	ChannelSize uint64 `json:"channelSize"`
-	Fee         uint64 `json:"fee"`
+	Invoice         string `json:"invoice"`
+	ChannelSize     uint64 `json:"channelSize"`
+	ChannelSizeSat  uint64 `json:"channelSizeSat"`
+	ChannelSizeMsat uint64 `json:"channelSizeMsat"`
+	Fee             uint64 `json:"fee"`
```
rolznz: Could we add a deprecated comment above the existing non-unit-specified properties (like ChannelSize and Fee, but all across this PR)

NOTE: no plan to remove yet.

alby/alby_service.go
```diff
@@ -123,6 +123,26 @@ func (svc *albyService) GetChannelPeerSuggestions(ctx context.Context) ([]Channe
 		return nil, err
 	}
 
+	for i := range suggestions {
```
rolznz, line 173: Now I'm wondering if we always need both. I think by clearly specifying "Sat" or "Msat" it's now obvious for the consumer. Here for channel sizes, I don't really think also returning Msat version is useful 🤔 

What do you think?

rolznz: In some places it might make sense to have both (for example app balances or fees, which could have msat amounts)

api/api.go
```diff
@@ -627,6 +640,8 @@ func (api *api) ListApps(limit uint64, offset uint64, filters ListAppsFilters, o
 					return nil, err
 				}
 				apiApp.BudgetUsage = budgetUsage / 1000
+				apiApp.BudgetUsageSat = budgetUsage / 1000
+				apiApp.BudgetUsageMsat = budgetUsage
```
rolznz: let's also update our internal properties (budgetUsage -> budgetUsageMsat) and method names (queries.GetBudgetUsage -> queries.GetBudgetUsageMsat)

lnclient/cashu/cashu.go
```diff
@@ -266,16 +278,40 @@ func (cs *CashuService) GetBalances(ctx context.Context, includeInactiveChannels
 
 	return &lnclient.BalancesResponse{
 		Onchain: lnclient.OnchainBalanceResponse{
-			Spendable: 0,
-			Total:     0,
+			Spendable:                              0,
```
rolznz: why do we add all these properties here? is it necessary?

im-adithya: Yeah this should be unnecessary, removed

alby/models.go
im-adithya: Done ✅

api/api.go
im-adithya: Done!

alby/alby_service.go
im-adithya: I asked AI to categorize all the fields:

1. balances
2. amounts
3. fees
4. liquidity
5. size/capacity
6. reserves
7. forwarding totals

I think we can leave Msat for liquidity and size/capacity? I removed for those two

rolznz: liquidity and size/capacity I think only make sense for sats, doesn't it? it's bitcoin on-chain, there is no such thing as millisats in on-chain BTC.

im-adithya: Yup, removed msat fields just for those, we should be good now!

hermes-alby:
## API Breaking Change Review 🔍

I've done a thorough review of every file in this diff checking for breaking changes to the Alby Hub API.

### ✅ No Breaking Changes to Existing API Responses

The PR follows a solid backward-compatible pattern: **old fields are preserved (with `// deprecated` comments) alongside new explicit `*Sat`/`*Msat` variants**. Every existing JSON key continues to return the same value it did before. Verified field-by-field:

| Existing field | Unit (before & after) | New alternatives |
|---|---|---|
| `maxAmount` | sats | `maxAmountSat`, `maxAmountMsat` |
| `budgetUsage` | sats | `budgetUsageSat`, `budgetUsageMsat` |
| `balance` | msat | `balanceSat`, `balanceMsat` |
| `localBalance` / `remoteBalance` / `localSpendableBalance` | msat | `*Sat`, `*Msat` variants |
| `amount` / `feesPaid` (Transaction) | msat | `amountSat`/`amountMsat`, `feesPaidSat`/`feesPaidMsat` |
| `totalBalance` (ListAppsResponse) | msat | `totalBalanceSat`, `totalBalanceMsat` |
| `unspendablePunishmentReserve` | sats | `*Sat`, `*Msat` variants |
| Swap, LSP, auto-channel fields | varies (preserved) | `*Sat`/`*Msat` variants |

This is a clean additive change — existing clients will continue to work unchanged. 👍

---

### ⚠️ Issues Worth Addressing Before Merge

**1. Inconsistent `*Sat`/`*Msat` coverage** — Some structs get both variants, while others only get `*Sat`:
- `AutoChannelResponse`: has `channelSizeSat` but no `channelSizeMsat`
- `LSPOrderResponse`: has `incomingLiquiditySat` but no `incomingLiquidityMsat`; has `outgoingLiquiditySat` but no `outgoingLiquidityMsat`

API consumers will likely expect a uniform pattern — either always provide both, or document why one is omitted.

**2. Deprecated fields should document their actual unit** — The `// deprecated` comments don't specify what unit the old field returns. For example:
```go
Balance int64 `json:"balance"` // deprecated
```
Should be:
```go
Balance int64 `json:"balance"` // deprecated — returns msat, use balanceSat or balanceMsat instead
```
This matters because some deprecated fields return sats (e.g., `budgetUsage`, `unspendablePunishmentReserve`) while others return msat (e.g., `balance`, `localBalance`). Without documentation, clients migrating to the new fields might pick the wrong one.

**3. Cashu/Phoenix backends: new fields rely on Go zero values** — These backends removed explicit `Spendable: 0, Total: 0` and `TotalReceivable: 0, NextMaxReceivable: 0` initializations. The new `*Sat`/`*Msat` fields will be 0 by default (Go zero value), which is correct. But for consistency with LDK/LND, it would be better to set them explicitly — makes the intent clear and avoids future bugs if someone adds initialization logic.

**4. `PendingBalanceDetails.Amount` is in sats (not msat)** — Since `AmountSat = amount` and `AmountMsat = amount * 1000`, the deprecated `Amount` field returns sats. This is different from most other deprecated fields that return msat. Worth calling out explicitly in the deprecation comment.

**5. Internal rename: `GetBudgetUsage` → `GetBudgetUsageMsat`** — Not an API break since it's internal, but worth noting for anyone who might be importing `db/queries` externally.

---

Overall this is a well-structured PR. The additive approach with deprecated fields is the right pattern for a gradual migration. The main ask would be: **document the unit on each deprecated field** and **make the `*Sat`/`*Msat` coverage consistent**.

hermes-alby:
**Update on point 1 (inconsistent `*Sat`/`*Msat` coverage):** After further thought, I think the PR's approach is actually correct here. Channel sizes, incoming/outgoing liquidity — these are on-chain UTXO amounts that are inherently whole sats. Including a `*Msat` variant would be misleading because it implies millisat precision exists where it physically can't (the smallest on-chain unit is 1 sat, and there's a dust limit below that).

The pattern should be:
- **`*Sat` + `*Msat`** → for lightning balances, payment amounts, routing fees (where msat precision exists)
- **`*Sat` only** → for on-chain / channel-size fields (always whole sats)

So what looked like inconsistency is actually intentional and correct. Withdrawn! 🙏

alby/alby_oauth_service.go
```diff
@@ -1032,6 +1032,9 @@ func (svc *albyOAuthService) GetLSPChannelOffer(ctx context.Context) (*LSPChanne
 		return nil, err
 	}
 
+	lspChannelOffer.LspBalanceMsat = lspChannelOffer.LspBalanceSat * 1000
```
rolznz: Msat not needed here

alby/alby_oauth_service.go
```diff
@@ -1361,9 +1364,12 @@ func (svc *albyOAuthService) requestAutoChannel(ctx context.Context, url string,
 	}
 
 	return &AutoChannelResponse{
-		Invoice:     invoice,
-		Fee:         fee,
-		ChannelSize: channelSize,
+		Invoice:        invoice,
+		Fee:            feeSat,
+		FeeSat:         feeSat,
+		FeeMsat:        feeSat * 1000,
```
rolznz: Msat not needed here

alby/alby_service.go
```diff
@@ -170,6 +170,24 @@ func (svc *albyService) GetChannelPeerSuggestions(ctx context.Context) ([]Channe
...
+		suggestions[i].MinimumChannelSizeSat = suggestions[i].MinimumChannelSize
+		suggestions[i].MaximumChannelSizeSat = suggestions[i].MaximumChannelSize
+
+		if suggestions[i].FeeTotalSat1m != nil {
+			feeTotalMsat1m := uint64(*suggestions[i].FeeTotalSat1m) * 1000
```
rolznz: Msat not needed here

alby/models.go
```diff
@@ -49,9 +49,12 @@ type AutoChannelRequest struct {
...
 type AutoChannelResponse struct {
-	Invoice     string `json:"invoice"`
-	ChannelSize uint64 `json:"channelSize"`
-	Fee         uint64 `json:"fee"`
+	Invoice        string `json:"invoice"`
+	ChannelSize    uint64 `json:"channelSize"` // deprecated
+	ChannelSizeSat uint64 `json:"channelSizeSat"`
+	Fee            uint64 `json:"fee"` // deprecated
+	FeeSat         uint64 `json:"feeSat"`
+	FeeMsat        uint64 `json:"feeMsat"`
```
rolznz: Msat not needed here

api/api.go
```diff
@@ -690,10 +713,14 @@ func (api *api) ListChannels(ctx context.Context) ([]Channel, error) {
 			ForwardingFeeBaseMsat:                    channel.ForwardingFeeBaseMsat,
 			ForwardingFeeProportionalMillionths:      channel.ForwardingFeeProportionalMillionths,
 			UnspendablePunishmentReserve:             channel.UnspendablePunishmentReserve,
+			UnspendablePunishmentReserveSat:          channel.UnspendablePunishmentReserve,
+			UnspendablePunishmentReserveMsat:         channel.UnspendablePunishmentReserve * 1000,
```
rolznz: does msats exist here? (same with CounterpartyUnspendablePunishmentReserve)

api/api.go
```diff
@@ -840,11 +867,15 @@ func (api *api) GetAutoSwapConfig() (*GetAutoSwapConfigResponse, error) {
...
-		Type:             constants.SWAP_TYPE_OUT,
-		Enabled:          swapOutEnabled,
-		BalanceThreshold: swapOutBalanceThreshold,
-		SwapAmount:       swapOutAmount,
-		Destination:      swapOutDestination,
+		Type:                 constants.SWAP_TYPE_OUT,
+		Enabled:              swapOutEnabled,
+		BalanceThreshold:     swapOutBalanceThreshold,
+		BalanceThresholdSat:  swapOutBalanceThreshold,
+		BalanceThresholdMsat: swapOutBalanceThreshold * 1000,
```
rolznz: I don't think Msat makes sense for swaps. It's in sats

api/api.go
```diff
@@ -887,7 +918,11 @@ func toApiSwap(swap *swaps.Swap) *Swap {
 		State:              swap.State,
 		Invoice:            swap.Invoice,
 		SendAmount:         swap.SendAmount,
+		SendAmountSat:      swap.SendAmount,
+		SendAmountMsat:     swap.SendAmount * 1000,
```
rolznz: I don't think Msat makes sense here

api/api.go
```diff
@@ -887,7 +918,11 @@ func toApiSwap(swap *swaps.Swap) *Swap {
...
 		ReceiveAmount:      swap.ReceiveAmount,
+		ReceiveAmountSat:   swap.ReceiveAmount,
+		ReceiveAmountMsat:  swap.ReceiveAmount * 1000,
```
rolznz: here too

api/rebalance.go
```diff
@@ -138,7 +138,11 @@ func (api *api) RebalanceChannel(ctx context.Context, rebalanceChannelRequest *R
 		Properties: map[string]interface{}{},
 	})
 
+	totalFeeSat := uint64(paymentRequest.MSatoshi/1000) + payRebalanceInvoiceResponse.FeeMsat/1000 - rebalanceChannelRequest.AmountSat
```
rolznz: maybe just calculate totalFeeMsat and use totalFeeSat as that value /1000? to simplify the calculation

frontend/src/types.ts
```diff
@@ -204,21 +210,33 @@ export type AutoSwapConfig = {
...
 };
 
 export type SwapInfo = {
   albyServiceFee: number;
   boltzServiceFee: number;
   boltzNetworkFee: number;
+  boltzNetworkFeeSat: number;
+  boltzNetworkFeeMsat: number;
   minAmount: number;
+  minAmountSat: number;
```
rolznz, line 225: for e.g. minAmount we can mark deprecated `// @deprecated` and try avoid using them in the UI, then remove them completely here

lnclient/cashu/cashu.go
```diff
@@ -188,8 +188,8 @@ func (cs *CashuService) GetNewOnchainAddress(ctx context.Context) (string, error
 
 func (cs *CashuService) GetOnchainBalance(ctx context.Context) (*lnclient.OnchainBalanceResponse, error) {
 	return &lnclient.OnchainBalanceResponse{
-		Spendable: 0,
-		Total:     0,
+		PendingBalancesDetails:      []lnclient.PendingBalanceDetails{},
```
rolznz: did these need to be added in?

nip47/controllers/get_budget_controller.go
```diff
@@ -50,8 +50,8 @@ func (controller *nip47Controller) HandleGetBudgetEvent(ctx context.Context, nip
 	}
 
 	responsePayload := &getBudgetResponse{
-		TotalBudget:   uint64(maxAmount * 1000),
-		UsedBudget:    usedBudget,
+		TotalBudget:   uint64(maxAmountSat * 1000),
```
rolznz, line 53: TotalBudget and UsedBudget should have unit versions right?

rolznz: oh, this is NWC. nevermind

im-adithya: No this is NWC, so everything is already in `msats`

api/api.go
im-adithya: Fixed!

frontend/src/types.ts
im-adithya: Done!

> try avoid using them in the UI

(Will be doing all the frontend changes in the follow-up to this)

PR #2174, opened 2026-03-25
feat: add backend-driven top home widgets

stackingsaunter:
## Summary
- add Home top widgets row with `Total Balance`, `Net Flows`, and `Bitcoin Price` cards
- add backend aggregate endpoint `GET /api/home/charts?from=...` so frontend no longer paginates transactions for chart math
- switch chart rendering to smooth `recharts` area charts and tune edge clipping/top padding for visual parity

## Known limitations
- Bitcoin Price 7d net change is still not reliable in all cases and currently falls back to `n/a`; this still needs a follow-up fix

## Resource usage estimate
Current implementation (one open Home tab) is approximately:
- `GET /api/home/charts`: every 5s (~12 req/min)
- `GET /api/mempool` historical price snapshots: 7 requests every 5s (~84 req/min)

Estimated total for top widgets: ~96 req/min/tab (lightweight reads, but non-trivial at scale).

## Test plan
- [x] `go test ./api ./http`
- [x] `cd frontend && yarn tsc:compile`
- [x] manual check in browser: cards render, smooth curves render, chart spacing/clipping updates apply

## Summary by CodeRabbit

* **New Features**
  * Added home dashboard charts displaying Total Balance, Net Flows, and Bitcoin Price with 7-day sparkline visualizations and percentage-change indicators
  * Charts automatically refresh with live transaction and price data to keep metrics current

### DIFF
--- a/api/home_charts.go
+++ b/api/home_charts.go
@@ -1,0 +1,87 @@
+package api
+
+import (
+	"context"
+	"errors"
+	"slices"
+
+	"github.com/getAlby/hub/constants"
+	"github.com/getAlby/hub/db"
+)
+
+func getTransactionTimestampMs(tx db.Transaction) int64 {
+	if tx.SettledAt != nil {
+		return tx.SettledAt.UnixMilli()
+	}
+	return tx.UpdatedAt.UnixMilli()
+}
+
+func getNetSat(tx db.Transaction) int64 {
+	amountSat := int64(tx.AmountMsat / 1000)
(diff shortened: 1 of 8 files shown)


bumi:
We partly unify the wallet and have spending/onchain balances there and thus I was wondering what counts towards these charts on the dashboard. 
A onchain-lightning swap would actually be a net flow of zero (or rather net outflow of the fees). 
What I understand is that this is not reflected.

+1 for keeping it draft.

PR #2178, opened 2026-03-25
Adding checks for ok status code in responses

### LINKED ISSUE #327: /api/alby/me endpoint returns response with all fields empty
We don't check the status code from the rails app.

Double check ALL external API calls to ensure we check the status code is ok

sergey3bv:
Should close #327 

## Summary by CodeRabbit

* **New Features**
  * Unified currency hook powering currency pickers, including SATS and consistent sorting.
  * Centralized rates endpoint for the frontend.
  * Currency selectors now show a loading placeholder and are disabled while rates load.

* **Bug Fixes**
  * Improved HTTP-status handling across integrations so non-OK responses are surfaced instead of being parsed.
  * Clearer error reporting for external-service failures, reducing UI breakage.

### DIFF
--- a/frontend/src/hooks/useCurrencies.ts
+++ b/frontend/src/hooks/useCurrencies.ts
@@ -43,6 +43,7 @@ export function useCurrencies(includeSats = false) {
     }
 
     return Object.entries(ratesData)
+      .filter(([code]) => code !== "BTC")
       .map(([code, details]): [string, string] => [
         code.toUpperCase(),
         details.name,
--- a/frontend/src/screens/internal-apps/ZapPlanner.tsx
+++ b/frontend/src/screens/internal-apps/ZapPlanner.tsx
@@ -115,7 +115,7 @@ export function ZapPlanner() {
   const [frequencyValue, setFrequencyValue] = React.useState("1");
   const [frequencyUnit, setFrequencyUnit] = React.useState("months");
   const [currency, setCurrency] = React.useState<string>("USD");
-  const { currencies } = useCurrencies(true);
+  const { currencies, isLoading: isCurrenciesLoading } = useCurrencies(true);
 
   const [convertedAmount, setConvertedAmount] = React.useState<string>("");
   const [satoshiAmount, setSatoshiAmount] = React.useState<number | undefined>(
@@ -139,6 +139,10 @@ export function ZapPlanner() {
   }, [open]);
 
   React.useEffect(() => {
+    if (isCurrenciesLoading) {
+      return;
+    }
+
     // If amount is empty, clear conversion output
     if (!amount) {
       setConvertedAmount("");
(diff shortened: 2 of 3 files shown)


api/esplora.go
```diff
@@ -46,6 +46,15 @@ func (api *api) RequestEsploraApi(ctx context.Context, endpoint string) (interfa
 		return nil, errors.New("failed to read response body")
 	}
 
+	if res.StatusCode != http.StatusOK {
+		logger.Logger.WithFields(logrus.Fields{
+			"url":         url,
+			"status_code": res.StatusCode,
+			"body":        string(body),
```
rolznz, line 53: can you include the endpoint as a property?

api/api.go
```diff
@@ -1244,6 +1244,15 @@ func (api *api) RequestMempoolApi(ctx context.Context, endpoint string) (interfa
 		return nil, errors.New("failed to read response body")
 	}
 
+	if res.StatusCode != http.StatusOK {
+		logger.Logger.WithFields(logrus.Fields{
+			"url":         url,
+			"status_code": res.StatusCode,
+			"body":        string(body),
```
rolznz, line 1280: can you include the endpoint as a property?

frontend/src/screens/internal-apps/ZapPlanner.tsx
```diff
@@ -126,6 +126,9 @@ export function ZapPlanner() {
     async function fetchCurrencies() {
       try {
         const res = await fetch("https://getalby.com/api/rates");
+        if (!res.ok) {
```
rolznz: this fetching seems inconsistent (we use hooks with SWR). I think there might be a hook already? if not we should add one

swaps/swaps_service.go
```diff
@@ -1530,6 +1539,10 @@ func (svc *swapsService) getNextUnusedAddressFromXpub() (string, error) {
 			return nil, err
 		}
 
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("esplora endpoint returned non-success code: %s", string(body))
```
rolznz: can you make the log a bit more clear (if we see the log on its own will we know this is related to swaps? - can you check the other logs too)

frontend/src/screens/internal-apps/ZapPlanner.tsx
sergey3bv: Out of the PR scope but sure. I will add it as well

frontend/src/hooks/useCurrencies.ts
```diff
@@ -0,0 +1,58 @@
+import React from "react";
+import useSWR from "swr";
+
+import { RATES_API_URL } from "src/constants";
+import { handleRequestError } from "src/utils/handleRequestError";
+
+const albyRatesFetcher = (url: string) =>
+  fetch(url).then((res) => {
```
rolznz, line 8: fetch won't work for Wails, I've created an issue here: https://github.com/getAlby/hub/issues/2192

PR #2217, opened 2026-04-08
feat(frontend): wallet on-chain mode

stackingsaunter:
Consolidates the wallet on-chain epic ([#2219](https://github.com/getAlby/hub/issues/2219)) into a single PR. Replaces the stacked PRs [#2218](https://github.com/getAlby/hub/pull/2218), [#2220](https://github.com/getAlby/hub/pull/2220), and [#2221](https://github.com/getAlby/hub/pull/2221), all now closed. Original authorship preserved via `--author=saunter` on the relevant commits.

## What's in it

- **On-chain mode toggle** in the Wallet hero — switch between Spending Balance and On-Chain Balance without leaving the Wallet screen. Hidden on backends without channel management (PHOENIX, CASHU).
- **Receive / Send routing** — on-chain mode routes the Receive/Send CTAs to the on-chain receive/send flows; `/wallet/send/onchain` now defaults to direct on-chain send instead of swap-from-spending.
- **Restyled on-chain transaction list** to match the Lightning transaction rows (big icons, inline amount + fiat, relative time).
- **On-chain transaction details dialog** — click a row to open a dialog with amount, status, confirmations, date/time, and full tx id. Copy tx id in one click. View on Mempool as a secondary action (hidden when no mempool URL is configured).
- **Pending state** for unconfirmed on-chain transactions — blue icon + pulse animation, "Pending" status label. Backend `unconfirmed` state value unchanged.
- **Node page cleanup** — the duplicate on-chain transaction list is gone; Wallet is the canonical place to review on-chain history.
- **Receive-flow reset fix** — the "Receive Another Payment" button on the deposit success screen now actually resets state and generates a new address instead of being a broken link.

## Simplifications vs. the original stack

- `OnchainBalanceSummary` and the compact/hero re-skin props were **deleted**. Each page renders its own shape inline (Channels keeps the compact card with reserve-warning tooltip; Wallet has a large hero, no tooltip).
- `OnchainTransactionsTable`'s 9 reusability props (`wrapInCard`, `title`, `className`, `contentClassName`, `showEmptyState`, four `emptyState*`) are **gone** — one consumer, one fixed shape. File moved to `components/wallet/`.
- `PendingClosedChannelsAlert` stays shared; removed from Wallet (channel-management context, belongs on Channels only).
- Accessibility polish on the mode toggle (`aria-label`, `aria-hidden` icon) and named `useState` import to match repo convention.

## Test plan

- [x] `yarn tsc:compile`
- [x] `yarn lint`
- [x] Manually verify the Wallet balance toggle and CTA routing in both balance modes
- [x] Manually verify the pending state for unconfirmed on-chain receives (mutinynet — caught the narrow window once LDK syncs)
- [x] Manually verify the on-chain transaction details dialog opens and all fields populate
- [x] Manually verify "Receive Another Payment" resets the receive screen and generates a new address
- [ ] Cross-check on non-LDK backends (LND) that the toggle still hides correctly when `hasChannelManagement` is false

## Summary by CodeRabbit

* **New Features**
  * Wallet toggle to switch between Lightning and On‑chain balance views.
  * New on‑chain transactions table and detailed transaction dialogs now shown in the Wallet.
  * Pending Closed Channels alert showing pending funds from channel closures.

* **Improvements**
  * Safer URL parameter validation and sync on receive screens.
  * Default on‑chain send form prioritized over swap.
  * Improved copy buttons for transaction fields and enhanced wallet sync/refresh behavior.


reneaaron (approved):
tACK

frontend/src/screens/wallet/index.tsx
```diff
@@ -97,82 +111,106 @@ function Wallet() {
...
+          {hasChannelManagement ? (
+            <button
+              type="button"
+              onClick={toggleBalanceMode}
+              aria-label={`Toggle balance mode, currently ${
+                isOnchainMode ? "On-chain Balance" : "Spending Balance"
+              }`}
+              className="inline-flex items-center justify-center gap-1 text-xs font-medium leading-none uppercase text-muted-foreground transition-colors hover:text-foreground"
+            >
+              {isOnchainMode ? "On-chain Balance" : "Spending Balance"}
```
bumi: I was wondering if we should use a "tab"-like navigation as we do in the connection screen. 
to maybe be consistent and have it more prominent? 

Image

reneaaron: image

like this?

bumi: I actually like this better.

bumi:
can we change the URL/browser history when switching between onchain/lightning to allow the back button? 
I now used that several times when going back from the send/receive screens. 



(send/receive in this combination is rather complex also (nothing we can change I think)

reneaaron:
Done — switching modes now updates a `?mode=onchain` query param, so the browser back button restores the previous mode after going through send/receive ([5ea85300](https://github.com/getAlby/hub/commit/5ea85300)).

frontend/src/screens/wallet/index.tsx
reneaaron: On the tab styling — I tried a few variations (segmented icon-only control, larger tabs, etc.) but couldn't find one that fully convinced me on the main wallet page. Going to leave the current toggle for now and tackle this in a follow-up.

PR #2218, opened 2026-04-08
feat(frontend): add on-chain transaction details dialog

stackingsaunter:
## Summary
- open a lightweight details dialog from on-chain transaction rows instead of navigating directly to Mempool
- show the currently available on-chain transaction fields in the dialog, including status, confirmations, date, amount, and full transaction id
- keep copy and Mempool actions available from inside the dialog while preserving the updated row styling

## Test plan
- [x] cd frontend && yarn tsc:compile
- [ ] Open the Wallet on-chain transaction list and verify row clicks open the dialog
- [ ] Verify copying the transaction id works from both copy actions
- [ ] Verify the Mempool button opens the correct transaction


reneaaron:
Folded into #2217.

PR #2221, opened 2026-04-08
feat(frontend): polish on-chain pending transaction state

stackingsaunter:
## Summary
- treat unconfirmed on-chain transactions as pending in the Wallet UI without changing backend state values
- use the pending label in the details dialog while keeping the existing sending/receiving directional copy
- apply the lightning-style blue pending treatment consistently across the on-chain row and dialog states

## Test plan
- [x] cd frontend && yarn tsc:compile
- [ ] Broadcast an outgoing on-chain transaction and verify it renders with blue pending styling before confirmations
- [ ] Observe an incoming unconfirmed transaction and verify it renders as receiving/pending with blue styling
- [ ] Verify confirmed transactions still render with the existing sent/received colors

## Summary by CodeRabbit

* **Bug Fixes**
  * Transaction status labels now display "Pending" instead of "Unconfirmed" for improved clarity.
  * Visual loading indicators and styling now consistently reflect pending transaction states.

### DIFF
--- a/frontend/src/components/channels/OnchainTransactionsTable.tsx
+++ b/frontend/src/components/channels/OnchainTransactionsTable.tsx
@@ -55,7 +55,7 @@ function typeStateLabel(tx: OnchainTransaction) {
 }
 
 function transactionStatusLabel(tx: OnchainTransaction) {
-  return tx.state === "confirmed" ? "Confirmed" : "Unconfirmed";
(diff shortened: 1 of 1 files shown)


bumi:
coderabbit is right, but it was like that before.

bumi (approved):

reneaaron:
Folded into #2217.

PR #2245, opened 2026-04-21
chore: make install and update scripts more agent-friendly

rolznz:
The download was too verbose.
This lead to the agent unable to read the whole output. It then made incorrect decisions on what to do next.

## Summary by CodeRabbit

* **Improvements**
  * Added preflight validation to ensure required dependencies are available before installation begins
  * Streamlined installer and update process output for a cleaner, less verbose user experience across Linux platforms
  * Enhanced installation feedback with clearer download progress messaging

### DIFF
--- a/scripts/linux-x86_64/install.sh
+++ b/scripts/linux-x86_64/install.sh
@@ -89,8 +89,15 @@ echo "Installing to: $INSTALL_DIR"
 mkdir -p "$INSTALL_DIR"
 cd "$INSTALL_DIR" || exit 1
 
+# check bzip2 is available before downloading
+if ! command -v bzip2 > /dev/null 2>&1; then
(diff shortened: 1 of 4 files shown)


scripts/linux-x86_64/install.sh
```diff
@@ -89,8 +89,15 @@ echo "Installing to: $INSTALL_DIR"
 mkdir -p "$INSTALL_DIR"
 cd "$INSTALL_DIR" || exit 1
 
+# check bzip2 is available before downloading
+if ! command -v bzip2 > /dev/null 2>&1; then
+  echo "❌ bzip2 is required but not installed. Run: sudo apt-get install -y bzip2" >&2
```
bumi, line 94: not all systems use apt-get

rolznz: Thanks, fixed here: https://github.com/getAlby/hub/pull/2246

PR #2254, opened 2026-04-23
refactor(frontend): drop time-based home greeting

reneaaron:
## Summary

- Drops the time-of-day greeting on the Home screen (\"Good Morning, {name}!\") in favor of a static \"Home\" title that matches the sidebar nav.

## Why

The previous greeting had a timezone/hour-range bug — at 00:00 local time it still showed \"Good Morning\" because the condition was `hours < 11` without a lower bound. Figuring out the right cutoffs turned into bikeshedding (when does morning end? is 4am night or morning? is \"Good Night\" even a greeting?), so rather than keep tuning the ranges I didn't care enough to get it right and just removed the greeting entirely.

## Test plan

- [ ] Home screen header shows \"Home\" at any time of day
- [ ] No layout regression in the AppHeader

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed dynamic greeting feature from the Home screen header. The header now displays a static "Home" title instead of a personalized, time-of-day dependent greeting.

### DIFF
--- a/frontend/src/screens/Home.tsx
+++ b/frontend/src/screens/Home.tsx
@@ -12,7 +12,6 @@ import {
   CardHeader,
   CardTitle,
 } from "src/components/ui/card";
-import { useAlbyMe } from "src/hooks/useAlbyMe";
 import { useBalances } from "src/hooks/useBalances";
 import { useInfo } from "src/hooks/useInfo";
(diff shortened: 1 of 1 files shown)


im-adithya (approved):
utACK

PR #2255, opened 2026-04-23
fix(frontend): auto-balance home dashboard columns

reneaaron:
## Summary
- Replace the fixed two-column grid on the Home screen with a CSS multi-column layout (`columns-1 lg:columns-2`) so widgets flow and auto-balance regardless of which ones render.
- Several home widgets return `null` based on user state (e.g. `WhatsNewWidget`, `LatestUsedAppsWidget`), which made the static left/right split feel lopsided — this lets the browser redistribute cards automatically.
- Group the Alby ecosystem cross-promo cards (Account / Go / Extension) in a single flex wrapper so they stay together in one column via `break-inside-avoid`.

## Test plan
- [ ] Verify dashboard looks balanced on a fresh setup (no used apps, update banner hidden)
- [ ] Verify dashboard looks balanced on an active setup (used apps, update banner visible)
- [ ] Verify cards never split across columns
- [ ] Verify mobile single-column layout still stacks correctly


im-adithya (approved):
tACK

PR #2265, opened 2026-04-27
feat: user labels for transactions

### LINKED ISSUE #2058: Add way to add a label to lightning transactions
sometimes you can pay invoices that have no memo and then later look back and not sure what the payment was about. I think we should possibly use the transaction metadata field for this

reneaaron:
## Summary

Closes [#2058](https://github.com/getAlby/hub/issues/2058) — let users add custom key/value labels to lightning transactions for bookkeeping or finding payments later.

Dynamic schema (no fixed fields) stored under `metadata.user_label`, so existing NIP-47 metadata keys (`comment`, `nostr`, `offer`, …) are preserved.

### Backend

- `PATCH /api/transactions/:paymentHash/label` (full-access group) accepting `{ "labels": { "<key>": "<value>", ... } }`.
- `api.SetTransactionUserLabels` looks up the transaction, decodes existing metadata, merges or clears `user_label`, writes back via `transactionsService.SetTransactionMetadata`.
- `SetTransactionMetadata` now uses `UpdateColumn` so editing labels does not bump `updated_at` (the list orders by `updated_at desc`, so previously labeled transactions jumped to the top).
- Validation: trim whitespace, drop rows with empty key/value, key ≤ 64 chars, value ≤ 1000 chars. Total metadata still capped by existing `INVOICE_METADATA_MAX_LENGTH` (4096).
- Tests in `api/transactions_test.go`: set, preserve other metadata keys, clear, trim/drop blanks, reject long key, reject long value, does not bump updated_at.
- Wails: `WailsRequestRouter` handler for the same route so the desktop app works too.

### Frontend

- `TransactionMetadata.user_label?: Record<string, string>` added to types.
- `TransactionLabelEditor` is rendered **inline** inside the existing transaction detail dialog (no nested dialogs). Display mode shows label badges + Edit/Add button; edit mode shows key/value rows + Save/Cancel.
- Empty state opens with a single blank row. No hardcoded suggested keys.
- Key inputs autocomplete from the user's previously-used label keys via `useLabelKeySuggestions`, which derives suggestions from any `/api/transactions` responses already in SWR's cache.
- In the **transactions list** a single tag icon appears next to the timestamp when a transaction has labels — keeps row heights uniform and the timeline scannable. Full labels are visible after opening the transaction.
- Save calls the PATCH endpoint, toasts on success/error, revalidates the `/api/transactions` SWR cache.

### CSV export

- `transactions-utils.tsx` collects the union of `user_label` keys across the exported transactions and emits each one as its own `label_<key>` column (sorted, stable between exports).
- Existing `metadata` JSON column is preserved so importers like Raccoin keep working — labels are visible both as flat columns and inside the JSON blob.

### Out of scope (follow-ups)

- Searching/filtering transactions by label.
- Alby Go / Alby extension (separate issues).

## Test plan

- [x] Open a transaction → "Add labels" → fill some fields → Save → tag icon appears next to timestamp, full labels in the detail dialog
- [x] Reopen detail → "Edit" → fields pre-populated → remove a row, add a custom key → Save → updates persist after reload
- [x] Empty all fields and save → labels cleared, other metadata (e.g. zap comment) still intact
- [x] After saving labels, the transaction stays in its original position in the list (no reorder)
- [ ] Key input autocomplete suggests keys from previously-labeled transactions on the same page
- [x] Export CSV → `label_<key>` columns appear, label values populated for the right rows, empty for unlabeled ones
- [ ] Wails desktop build: same flow works (routes through `WailsRequestRouter`)
- [x] `go test ./api/...` passes

## Summary by CodeRabbit

* **New Features**
  * Add, edit, and manage custom key-value labels for transactions via an inline dialog
  * View label counts as badges in transaction list items
  * Transaction IDs now included in all transaction responses

* **Improvements**
  * CSV export now includes dynamic columns for all transaction labels and safer cell escaping
  * UI layout tweaks for clearer transaction detail spacing and modal sizing

* **Tests**
  * Added tests covering label storage, trimming, clearing, and size limits


im-adithya:
- Removed checks for individual keys and values (we anyways do a check for overall metadata)
- Changed backend to allow setting metadata on incoming transactions
- Simplified frontend components
- Renamed to `user_labels` everywhere from `user_label`/`user_labels`
- Simplified and verified CSV export to see everything's working well

im-adithya:
Things to finalize before merging:

- [x] Confirm `user_labels` (with `label_sample`, `label_foo` in csv export) naming
- [x] Confirm API tweak to return dbTransaction `ID` in list transactions and using it in change user labels call

rolznz:
Could you double check the metadata length limit is preserved in the new edit labels endpoint?

otherwise looks good to me, utACK

im-adithya:
Yup checked it first thing, we are using the same `SetTransactionMetadata` method which has the check so we're good.

reneaaron:
tACK 💯

PR #2266, opened 2026-04-28
feat: make Alby the default theme and rename default to classic

im-adithya:
## Summary by CodeRabbit

* **Style**
  * Updated default application theme to "Alby"
  * Modified available theme options in theme selector
  
* **Refactor**
  * Reorganized theme and styling system for improved consistency

### DIFF
--- a/frontend/src/themes/classic.css
+++ b/frontend/src/themes/classic.css
@@ -7,7 +7,7 @@
 /* Use generalized selector (:root, .dark) here because all themes
    should fall back to these values */
 :root,
-.theme-default {
+.theme-classic {
   --app-font-sans: "Inter Variable", ui-sans-serif, system-ui, sans-serif;
   --radius: 0.625rem;
   --background: oklch(1 0 0);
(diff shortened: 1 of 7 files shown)


reneaaron (approved):
tACK

💯

frontend/src/themes/classic.css
```diff
@@ -7,7 +7,7 @@
 /* Use generalized selector (:root, .dark) here because all themes
    should fall back to these values */
 :root,
```
reneaaron: This is a bit strange now: 

`classic` now is the base for all themes still? 🤔

reneaaron:
Pushed a follow-up refactor on top of this PR ([1af4e036](https://github.com/getAlby/hub/commit/1af4e036)) to address the asymmetry between Alby and Classic that the `:root, .theme-classic` selector was hinting at:

- Extracted neutral defaults (Inter font, 0.625rem radius, achromatic primary, original chart colors) into a new `themes/base.css` on `:root` / `.dark`, so no single theme owns the global fallback values.
- Reduced `alby.css` to a pure `.theme-alby` / `.theme-alby.dark` override layer (Figtree, larger radius, yellow primary/charts, logo vars, button gradient).
- Deleted `classic.css` entirely — Classic now falls through to the base values, which matches its meaning ("the neutral default look").
- Moved framework-level imports (`tw-animate-css`, `@custom-variant dark`) from inside the deleted theme file up to `index.css` where they're not theme-specific.

Existing users with `"default"` in localStorage continue to fall back to Alby (intended behavior — confirmed offline).

PR #2275, opened 2026-05-01
feat: just in time channels with lsps2

### LINKED ISSUE #2272: JIT channels using LSPS2 spec
Image

Image

frnandu:
fixes #2272 

## Summary by CodeRabbit

* **New Features**
  * LSPS2 liquidity source support for automatic Just‑in‑Time (JIT) channel receives; toggle available in Node settings.

* **UI/UX Improvements**
  * Liquidity source shown in About/settings with node link; new alert/components for first‑channel JIT info.
  * Receiving flow: JIT-aware limits, clearer fee/capacity alerts, suppressed “open first channel” prompts when source present.

* **Bug Fixes**
  * Invoice fees are now stored and shown correctly in transactions.

* **Documentation**
  * Added docs and example for LSPS2 addresses config.

TODO:
- [x] disable LDK trace logging on alby team cloud nodes due to log spam bug when JIT is enabled
- [x] update getting started checklist
- [x] settings option to enable/disable JIT
- [x] get default JIT provider from getalby.com api (+ minimum amount)
- [x] Add minimum amount validation on receive page (enter less than 2501 sats gives unclear message right now)
- [x] Review minimum receive + reserve amount with Megalith - reduced to 546 sats + asked to reduce to 354
- [x] Do not enable JIT if you have public channels
- [x] add JIT boolean to nwc_channel_ready event
- [x] Only set JIT provider for LDK (flow is: env variable first, otherwise try get from getalby.com, otherwise fallback to hardcoded value)
- [x] Remove hardcoded injected channel suggestions in alby/alby_service.go
- [x] Double check backwards compatibility that older hub versions filter out non-LSPS1 type
- [x] Add link to guide on ReceiveInvoice page for JIT receive fee 
- [x] remove 0-amount receive
- [x] update UI - "minimal" fee display on ReceiveInvoice page (@reneaaron 's design)


rolznz:
I get this trace log spammed consistently from LDK:
```
{"level":"trace","line":1231,"log_type":"LDK-node","module_path":"lightning_background_processor","msg":"Persisting LiquidityManager...","time":"2026-06-04T17:52:48+07:00"}
```

rolznz:
The "Getting Started" checklist should change right?

image

Suggestion: don't show the "Open your first channel" checklist item when LDK+JIT is enabled (I think available from the info response)

rolznz:
image

Here I don't get any validation warning, is there any minimum?
1, 1000 didn't work. 10,000 did.

rolznz:
I am wondering, does it make sense to have an option in settings to enable/disable this feature? (a required node restart is OK)

frnandu:
> I am wondering, does it make sense to have an option in settings to enable/disable this feature? (a required node restart is OK)

I'm all for it and even having the possibility of choosing which LSP to use from there as well.
Meanwhile for testing the env var should be enough.

frontend/src/components/TransactionItem.tsx
```diff
@@ -277,10 +277,10 @@ function TransactionItem({ tx, transactionListKey }: Props) {
           <TransactionDetailRow label="Date & Time">
             {updatedAt.format("D MMMM YYYY, HH:mm")}
           </TransactionDetailRow>
-          {tx.state != "failed" && type == "outgoing" && (
+          {tx.state != "failed" && tx.feesPaidMsat > 0 && (
             <TransactionDetailRow label="Fee">
               <FormattedBitcoinAmount amountMsat={tx.feesPaidMsat} />
-              {tx.feesPaidMsat > 0 && (
+              {type == "outgoing" && (
```
rolznz, line 283: is it intended to only show the percent for fees for outgoing payments? (I would always show it)

api/models.go
```diff
@@ -331,6 +331,7 @@ type InfoResponse struct {
 	MempoolUrl                  string              `json:"mempoolUrl"`
 	ChainDataSourceType         string              `json:"chainDataSourceType,omitempty"`
 	ChainDataSourceAddress      string              `json:"chainDataSourceAddress,omitempty"`
+	JitChannelsLiquiditySource  string              `json:"jitChannelsLiquiditySource,omitempty"`
```
rolznz, line 333: It could be nice to add this to the README with some example options for mainnet and mutinynet

frontend/src/components/TransactionItem.tsx
frnandu: The incoming transaction fees for JIT might not always be based on the amount, since if a fixed extra capacity is added by the LSP and the fees are also fixed for that extra capacity it will not matter the amount you received in relation to the paid fees so my understanding is that % in that case doesn't make so much sense, but that might be a preference thing.

frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
```diff
@@ -101,6 +121,29 @@ export default function ReceiveInvoice() {
...
+    <Alert>
+      <InfoIcon className="h-4 w-4" />
+      <AlertTitle>New channel fee expected</AlertTitle>
+      <AlertDescription>
+        <p>
+          {displayedJitFeeMsat >= 1000 ? (
+            <>
+              Around <FormattedBitcoinAmount amountMsat={displayedJitFeeMsat} />{" "}
+              will be deducted when this payment is received to open a new
+              channel.
```
rolznz: I wonder if we need a tooltip or button to open more info (in a modal for example) to teach the user about channels

frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
```diff
@@ -101,6 +121,29 @@ export default function ReceiveInvoice() {
     copyToClipboard(transaction?.invoice as string);
   };
 
+  const newChannelFeeAlert = (
+    <Alert>
+      <InfoIcon className="h-4 w-4" />
+      <AlertTitle>New channel fee expected</AlertTitle>
+      <AlertDescription>
+        <p>
+          {displayedJitFeeMsat >= 1000 ? (
```
rolznz: Why is there this conditional? why 1000 sats?

frontend/src/components/TransactionItem.tsx
rolznz: Ah right, the liquidity cost can be spread across many payments, so it doesn't make any sense. I wonder if it's worth mentioning anything like that (not necessarily here, but somewhere)

frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
frnandu: it's Msats ;)

api/models.go
frnandu: Added some info and mutinynet example, not sure if we can share the mainnet one yet.

README.md
```diff
@@ -254,6 +255,7 @@ _To configure via env, the following parameters must be provided:_
 - `NETWORK=signet`
 - `LDK_ESPLORA_SERVER=https://mutinynet.com/api`
 - `LDK_GOSSIP_SOURCE=https://rgs.mutinynet.com/snapshot` (NOTE: by default ALby Hub does not use RGS)
+- `LDK_LSPS2_ADDRESSES=03466abad0e5162c07e58c6e66a9ccac1f587944fd837d84cc3e777b8138dd9936@192.243.215.98:25030` (optional, for JIT channel receives on Mutinynet — verify current details with the LSP LQwD)
```
rolznz: What is this comment "optional, for JIT channel receives on Mutinynet — verify current details with the LSP LQwD"?

Maybe I was wrong to suggest we add example values here, if we do not have a reliable one right now

rolznz:
image

On the node page, should this also be updated ("Receive your first payment"?)

rolznz:
I did a mainnet test with Megalith / possibly feedback for his side:
- Lower than 2,500 sats doesn't work at all. 2501 sats works. I would expect as low as 1 sat should work, but the invoice would cost an additional 2,500 sats (this is fine, for me it's a fee to pay). Do you agree? (otherwise I think we need to improve our messaging)
- my reserve then was 661 / 1,025. This is super weird. I would expect my reserve to be filled, and why isn't it the default 354 sats?

--

- Receive BOLT-12 offer didn't even work EVEN when I had my channel, so I don't think we should focus on this at all right now. (paying from Phoenix). I don't think we should even consider a JIT channel via BOLT-12 until normal receives are more reliable.

frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
```diff
@@ -101,6 +118,49 @@ export default function ReceiveInvoice() {
...
+            You'll receive about{" "}
+            <FormattedBitcoinAmount amountMsat={netReceiveMsat} />
+          </AlertTitle>
+          <AlertDescription>
+            <p>
+              <FormattedBitcoinAmount amountMsat={displayedJitFeeMsat} /> of
+              this payment is a one-time fee to open a Lightning channel, which
+              lets you receive payments.{" "}
+              <ExternalLink
+                to="https://guides.getalby.com/user-guide/alby-hub/node/increase-receiving-capacity"
```
rolznz: I think we should make a new guide specifically for JIT, and maybe link to this one as an alternative option

frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
```diff
@@ -138,6 +196,9 @@ export default function ReceiveInvoice() {
                           className="text-xl"
                         />
                       </div>
+                      {isJitReceiveInvoice && displayedJitFeeMsat >= 1000 && (
```
rolznz, line 204:
```suggestion
                      {isJitReceiveInvoice && displayedJitFeeMsat > 0 && (
```

frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
```diff
@@ -54,6 +57,20 @@ export default function ReceiveInvoice() {
...
+  const requestedAmountMsat = +amountSat * 1000 || transaction?.amountMsat || 0;
+  const isNearReceivingCapacity =
+    !!hasChannelManagement && requestedAmountMsat >= 0.8 * totalReceivableMsat;
+  const isJitReceiveInvoice =
+    !!hasChannelManagement &&
+    !!lsps2Source &&
+    !!transaction &&
+    transaction.amountMsat > totalReceivableMsat;
+  const displayedJitFeeMsat = paymentDone
+    ? (invoiceData?.feesPaidMsat ?? transaction?.feesPaidMsat ?? 0)
```
rolznz: this code is a bit confusing, It could just use `invoiceData` rather than also `transaction` - since we only set `paymentDone` when `invoiceData?.settledAt` exists.

To fix this: `paymentDone` should not be a state variable. It can just be a standard boolean.

`const paymentDone = !!invoiceData?.settledAt`

frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
frnandu: Do we want to display sub sat amounts???

rolznz: it's more about just making the code easier to read, the fee will always be more than 1 sat right?

frnandu: I had few tests that fee was less.
For mainnet you're probably right that it will be always more than 1 sat, but I think for correctness sake we should keep the proper condition, otherwise you might see 0 sats fee displayed in mutinynet and such other testing environments.

rolznz:
I added minimum amount validation on receive page but only if the user has no channels.

lnclient/ldk/ldk.go
```diff
@@ -710,7 +748,11 @@ func (ls *LDKService) MakeInvoice(ctx context.Context, amountMsat int64, descrip
 
 	maxReceivable := ls.getMaxReceivable()
 
-	if amountMsat > maxReceivable {
+	isVariableAmountInvoice := amountMsat == 0
```
rolznz: I don't think we should add support for 0-amount receive in this PR. Is there a reason you added it?

frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
rolznz: using a link to our guides instead.

README.md
rolznz: I removed it.

frontend/src/screens/wallet/receive/ReceiveInvoice.tsx
```diff
@@ -62,9 +62,11 @@ export default function ReceiveInvoice() {
...
     ? info?.jitChannelsLiquiditySource
     : undefined;
-  // the configured JIT liquidity source may enforce a minimum channel size -
-  // any payment that opens a new channel must be at least this amount.
-  const lsps2MinimumChannelSizeSat = React.useMemo(() => {
+  const lsps2MinimumPaymentSizeSat = React.useMemo(() => {
+    if (info?.jitChannelsMinPaymentSizeMsat) {
+      return Math.ceil(info.jitChannelsMinPaymentSizeMsat / 1000);
+    }
+    // Fallback for older backends that only expose peer suggestions.
```
rolznz: should we remove the fallback since this isn't released yet, and only LDK supports JIT anyway?

lnclient/ldk/ldk.go
```diff
@@ -2619,6 +2624,80 @@ func (ls *LDKService) GetLiquiditySourceLsps2() string {
...
+
+	ls.lsps2MinPaymentSizeMsat = minPaymentSizeMsat
+	ls.lsps2MaxPaymentSizeMsat = maxPaymentSizeMsat
+	ls.lsps2InfoFetchedAt = time.Now()
+}
+
+func effectiveLsps2MinPaymentSizeMsat(params ldk_node.Lsps2OpeningFeeParams) (uint64, bool) {
+	paymentSizeMsat := params.MinPaymentSizeMsat
+
+	for range 8 {
```
rolznz, line 2693: what is 8 for here?

frnandu: It’s just a safety cap on the adjustment loop.

rolznz: I don't understand this loop at all. And we're only bumping the price by 1 millisat which is in my opinion basically useless.

Could you document better explain what we are doing?

rolznz: I misunderstood how the method worked. But added a small adjustment to require at least 1 sat, and added some comments

rolznz (approved):
tACK

rolznz:
@coderabbitai review

frontend/src/screens/settings/NodeSettings.tsx
```diff
@@ -0,0 +1,84 @@
...
+          />
+          <Label htmlFor="jit-channels" className="ml-2 cursor-pointer">
+            Enable JIT channels for receiving
+          </Label>
+        </div>
+        {!hasJitSource && (
+          <p className="text-sm text-muted-foreground">
+            No JIT liquidity source is available for your network, so JIT
+            channels can't be used.
+          </p>
```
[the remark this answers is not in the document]
rolznz: Minor: we already hardcode a JIT fallback in the backend, so this will never actually happen.

PR #2304, opened 2026-05-06
fix: hide transaction ID from on-chain transactions list

### LINKED ISSUE #2299: Do not show transaction IDs on wallet on-chain tab
They are not very helpful in my opinion. Maybe only show in the detail modal? and we could also show the bitcoin address there too.

reneaaron:
## Summary
- Removes the truncated transaction ID subtitle from each row in the wallet on-chain tab list — it added noise without much value at a glance.
- The full TX ID is still shown in the transaction detail modal (with copy + link to mempool), so nothing is lost.

The issue also suggested showing the bitcoin address in the detail modal. Holding off on that for now: it'd require extending `lnclient.OnchainTransaction` with an `Address` field, and LDK's `PaymentKindOnchain` doesn't expose the address — so the field would be empty for the main backend. 

Closes #2299

## Test plan
- [ ] Open the wallet on-chain tab — transaction rows show type/state + relative time + amount, no TX ID line.
- [ ] Click a transaction — detail modal still shows full TX ID with copy button and "View on Mempool" link.

### DIFF
--- a/frontend/src/components/OnchainTransactionItem.tsx
+++ b/frontend/src/components/OnchainTransactionItem.tsx
@@ -41,7 +41,6 @@ function OnchainTransactionItem({
   const typeStateText = typeStateLabel(tx);
   const statusText = isPending ? "Pending" : "Confirmed";
   const createdAt = dayjs(tx.createdAt * 1000).local();
-  const subtitle = `${tx.txId.slice(0, 10)}…${tx.txId.slice(-8)}`;
 
   const icon = (
     <div className="flex items-center">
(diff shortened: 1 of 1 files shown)


rolznz:
> extending lnclient.OnchainTransaction with an Address field, and LDK's PaymentKindOnchain doesn't expose the address — so the field would be empty for the main backend.

Thanks for checking, sounds good to hold it off for now. I asked the LDK team about this.

rolznz (approved):
utACK

PR #2305, opened 2026-05-06
chore: rename spending balance to lightning balance

### LINKED ISSUE #2297: Hub Wallet UX terminology: "Spending" vs "On-chain" balance
This is IMO inconsistent - it's conceptual vs technically correct.

Change "Spending Balance" to "Lightning Balance"

rolznz:
Closes https://github.com/getAlby/hub/issues/2297

## Summary by CodeRabbit

* **Style**
  * Updated UI terminology throughout the application, replacing "Spending Balance" with "Lightning Balance" across balance displays, channel management, payment confirmations, transaction flows, swap settings, and account backup information. Enhanced UI consistency and clarity for users.

### DIFF
--- a/frontend/src/screens/channels/IncreaseOutgoingCapacity.tsx
+++ b/frontend/src/screens/channels/IncreaseOutgoingCapacity.tsx
@@ -258,9 +258,9 @@ function NewChannelInternal({
                   </div>
                 </TooltipTrigger>
                 <TooltipContent>
-                  Configure the amount of lightning balance you need. You will
-                  need to deposit on-chain bitcoin to cover the entire channel
-                  size, plus on-chain fees.
+                  Configure the amount of spending balance you need in your new
+                  lightning channel. You will need to deposit on-chain bitcoin
+                  to cover the entire channel size, plus on-chain fees.
                 </TooltipContent>
               </Tooltip>
             </TooltipProvider>
--- a/frontend/src/screens/channels/IncreaseOutgoingCapacity.tsx
+++ b/frontend/src/screens/channels/IncreaseOutgoingCapacity.tsx
@@ -214,7 +214,7 @@ function NewChannelInternal({
       <AppHeader
         pageTitle="Open Channel with On-Chain"
         title="Open Channel with On-Chain"
-        description="Funds used to open a channel minus fees will be added to your spending balance"
+        description="Funds used to open a channel minus fees will be added to your lightning balance"
         contentRight={
           <div className="flex items-end">
             <Link to="/channels/incoming" className="underline text-sm">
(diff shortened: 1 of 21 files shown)


reneaaron (approved):
tACK

Should we also consider updating the docs as well? (maybe there is also a "search & replace" or we finally have a reason to sync gitbook to a repository?

frontend/src/screens/channels/IncreaseOutgoingCapacity.tsx
```diff
@@ -252,13 +252,13 @@ function NewChannelInternal({
...
                     <Label htmlFor="amount">
-                      Increase spending balance (sats)
+                      Increase lightning balance (sats)
                     </Label>
                     <InfoIcon className="h-4 w-4 shrink-0 text-muted-foreground" />
                   </div>
                 </TooltipTrigger>
                 <TooltipContent>
-                  Configure the amount of spending capacity you need. You will
+                  Configure the amount of lightning balance you need. You will
```
im-adithya: minor: shouldn't this be something like "configure how much you want to increase your lightning balance"?

frontend/src/screens/wallet/send/PayFromSelect.tsx
```diff
@@ -128,7 +128,7 @@ export default function PayFromSelect({ appId, onChange }: Props) {
                 {option.app ? (
                   <AppOption app={option.app} />
                 ) : (
-                  <SpendingOption />
+                  <LightningOption />
```
im-adithya, line 131: Should we then also change other components with Spending like `ReceiveToSpending` etc in the same PR?

im-adithya (approved):
tACK, just minor nits

> Should we also consider updating the docs as well? (maybe there is also a "search & replace" or we finally have a reason to sync gitbook to a repository?

I'm worried about this because some links also have "/xxx-spending-yyy" in them so we must make sure they are redirected to the new pages

frontend/src/screens/channels/IncreaseOutgoingCapacity.tsx
rolznz: I think here actually the old copy was better. I'll change it to "Configure the amount of spending balance you need in your new lightning channel"

frontend/src/screens/wallet/send/PayFromSelect.tsx
rolznz: thanks, done

rolznz:
This was not a direct regex replace, it was done mostly by AI and I reviewed the output and changed some copy in places. So I think it'd be best we can do the same for the docs if it's in a repo (I guess in 2026 we should not be manually editing all our guides by hand). Good point by @im-adithya to make sure we preserve redirects

PR #2309, opened 2026-05-07
feat: add swap in button to insufficient lightning balance alert

### LINKED ISSUE #2303: Add new "swap in" option to the insufficient balance warning
Add a new option on the left (higher priority) than the current "increase spending balance" and rename that one (since it's for opening an outbound channel?)

Reason: we currently only show an option for opening a new outbound channel. That doesn't make sense if a user already has enough channels.

Image

See the increase receiving capacity warning for comparison:

Image

rolznz:
Closes https://github.com/getAlby/hub/issues/2303

image


im-adithya (approved):
tACK

PR #2310, opened 2026-05-07
fix: allow pasting BIP-21 URL into the wallet send input

### LINKED ISSUE #2123: Support pasting BIP 21 links into the send input
e.g. bitcoin:bc1qxxxx?amount=X.XXXXXXXX rather than the user having to manually paste the address and then satoshi amount separately.

e.g. for making an on-chain payment at a BTCPayServer checkout

rolznz:
Fixes https://github.com/getAlby/hub/issues/2123

### DIFF
--- a/frontend/src/screens/wallet/Send.tsx
+++ b/frontend/src/screens/wallet/Send.tsx
@@ -26,12 +26,8 @@ export default function Send() {
   const [isLoading, setLoading] = React.useState(false);
   const [showSwapAlert, setShowSwapAlert] = React.useState(false);
 
-  React.useEffect(() => {
-    const uri = searchParams.get("bip21");
-    if (!uri) {
-      return;
-    }
-    try {
+  const handleBip21 = React.useCallback(
+    (uri: string) => {
       const bip21 = parseBip21(uri);
       if (bip21.lightning) {
         const invoice = new Invoice({ pr: bip21.lightning });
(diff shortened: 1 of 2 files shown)


frontend/src/screens/wallet/Send.tsx
```diff
@@ -48,24 +44,33 @@ export default function Send() {
...
+        throw new Error("invalid bitcoin address");
       }
+      navigate(`/wallet/send/onchain`, {
+        state: {
+          args: {
+            address: bip21.address,
+            amountSat: bip21.amountSat ? String(bip21.amountSat) : undefined,
+          },
+        },
+        replace: true,
```
im-adithya: We should have a param for replace otherwise clicking back from onchain success/pending would take you back to wallet (or whatever was before send)

rolznz: why is it different only here? we should have some consistency

Can we remove replace: true in the new handleBip21 method?

rolznz: Yeah, I think this is the proper fix - applying it now

im-adithya: Because there should be no back button when the user is coming from a bip21 link, but yeah this is simple and it doesn't hurt to show back to Send :+1:

im-adithya (approved):
tACK

PR #2324, opened 2026-05-08
chore: replace go-nostr with fork carrying CLOSED reconnect fix

reneaaron:
Depends on getAlby/go-nostr#2.


rolznz:
Closing in favor of https://github.com/getAlby/hub/pull/2331 because nbd-wtf/go-nostr is archived

PR #2344, opened 2026-05-14
Rebrand LendaSwap as Satora

luckysori:
We recently rebranded LendaSwap as Satora: https://app.satora.io.

The currently deployed link (https://lendaswap.com) just tells users to go to the new URL. Users can still access their swap history, but new swaps must go through https://app.satora.io.

image

image


rolznz:
I added a minor edit here as I couldn't push changes to this branch: https://github.com/getAlby/hub/pull/2376

PR #2369, opened 2026-05-21
fix: stop prompting for bitcoin: protocol handler on every load

reneaaron:
## Summary
- The browser-native \`registerProtocolHandler\` prompt was reappearing on every page load, because browsers re-prompt every time \`navigator.registerProtocolHandler\` is called if the user dismissed (X'd) the previous prompt without explicitly accepting or denying.
- Gate the call with \`sessionStorage\` so we ask at most once per browser session. Users who dismiss still get re-asked next session, and an explicit accept persists permanently in the browser as before.

## Test plan
- [ ] Load Hub in a fresh tab → see the bitcoin protocol handler bar once.
- [ ] Dismiss it with X, reload the same tab → bar does NOT reappear.
- [ ] Open Hub in a new tab → bar appears again.
- [ ] Accept the prompt → bar never reappears (browser remembers permanently).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented repeated browser protocol-registration prompts by ensuring the handler is registered only once per session, avoiding redundant dialogs on subsequent page loads.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getAlby/hub/pull/2369?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/hooks/useRegisterProtocolHandler.ts
+++ b/frontend/src/hooks/useRegisterProtocolHandler.ts
@@ -2,16 +2,28 @@ import React from "react";
 
 import { isHttpMode } from "src/utils/isHttpMode";
 
+const STORAGE_KEY = "bitcoin-protocol-handler-registered";
+
 export function useRegisterProtocolHandler(basePath: string) {
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


im-adithya (approved):
tACK

PR #2371, opened 2026-05-22
chore: address feedback on currency input field

rolznz:
- Removed the outer bordered wrapper around the input + context rows; InputGroup now handles its own
  border and focus ring (focus only highlights the input).
- Moved contextRows out of the bordered area so they no longer look like an input.
- Added cursor-default to the contextRows wrapper to neutralize the inherited cursor: pointer from the
   global [data-slot="field"] rule.
- Always render the leading unit addon so "sats" displays when BIP-177 is not active (in addition to
  "₿", "BTC", or fiat symbol).
- Added a toast("Switched to BTC for decimal amount") when typing/pasting a decimal auto-switches the denomination to BTC.
- Replaced the embedded buttons' default focus ring with an inset 2px ring-inset ring-ring, plus  peer-focus-visible so the adjacent divider borders go transparent on focus (cleaner ring, no clipped
  vertical stripes).
  - Rounded the rightmost button's right corners (rounded-r-md) so the inset ring follows the InputGroup's curve.
  - Changed the right-side button label to show "BTC" when BTC is active and the user's preferred sats
  unit (₿ for BIP-177, otherwise sats) when sats is active.
  
  
image

image

image

### DIFF
--- a/frontend/src/components/CurrencyInputField.tsx
+++ b/frontend/src/components/CurrencyInputField.tsx
@@ -1,5 +1,6 @@
 import { ArrowRightLeftIcon } from "lucide-react";
 import * as React from "react";
+import { toast } from "sonner";
 import {
   Field,
   FieldDescription,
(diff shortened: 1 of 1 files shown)


rolznz:
Review from @stackingsaunter - we don't want a duplicated BTC on left/right.

PR #2391, opened 2026-06-03
fix: stop retrying NIP47 info publish for deleted apps

### LINKED ISSUE #2390: Bug: Infinite re-queue to republish info of deleted app connection
- `nip47/nip47_service.go:115` → `PublishNip47Info` → `svc.db.First(&app, appId)` at `nip47/publish_nip47_info.go:74` fails with `record not found` because **the App row no longer exists** (the connection was deleted).
- On failure the queue **re-enqueues the same request with a backoff and `Attempt+1`** (`nip47_service.go:125-128`) — with **no terminal condition** for a permanently-missing app. So it retries that deleted app *forever*, ~once a minute, indefinitely.

bumi:
## Problem

The NIP47 info publish queue (`StartNip47InfoPublisher` in `nip47/nip47_service.go`) re-enqueues **every** failed publish with an incrementing backoff and **no terminal condition**.

When an app connection has been deleted, `PublishNip47Info` fails the `db.First(&app, appId)` lookup with `gorm.ErrRecordNotFound` on every attempt — so the queue item is retried **forever** (~once a minute). In the last 24h of production logs this showed up as a steady stream of `Failed to publish NIP47 info from queue` errors (`error: record not found`) concentrated on individual instances stuck retrying deleted apps.

## Fix

When the publish fails with `gorm.ErrRecordNotFound`, the app no longer exists and the info event can never be published — so we **drop the queue item** instead of requeuing. All other errors (offline relay, publish timeouts) still retry with backoff as before.

## Test

Added `TestPublishNip47Info_AppNotFound` which locks the contract the fix relies on: publishing info for a non-existent app surfaces `gorm.ErrRecordNotFound` (so the loop's `errors.Is` drop logic stays correct if the error handling is ever refactored).

`go test ./nip47/` passes.

fixes #2390 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced NIP47 info publishing with improved error detection: when an app connection is not found, the system now drops the queued request instead of retrying indefinitely, with appropriate warning logs.

* **Tests**
  * Added test case validating error handling for non-existent app connections during publishing.

### DIFF
--- a/nip47/nip47_service.go
+++ b/nip47/nip47_service.go
@@ -2,6 +2,7 @@ package nip47
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/getAlby/go-nostr"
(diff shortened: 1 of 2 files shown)


rolznz (approved):
utACK

PR #2394, opened 2026-06-04
fix: change first payment checklist item to receive-only (#2393)

### LINKED ISSUE #2393: Change checklist copy for send/receive and condition
```
{
      title: "Send or receive your first payment",
      description: "Add funds to your wallet, then make your first payment.",
      checked: hasTransaction,
      to: "/wallet",
    },
```

change it to receive only, but update the checked condition so that it also checks if the user has a non-zero balance (so that people who open outbound channels don't get a strange experience)

reneaaron:
Closes #2393

## What
- Renames the onboarding checklist item from **"Send or receive your first payment"** to **"Receive your first payment"**, with copy updated to match ("Add funds to your wallet to start using it.").
- Marks the item complete when the wallet has a non-zero spendable lightning balance, in addition to the existing "has a transaction" check.

## Why
Users who open an outbound channel end up with a spendable balance but no transaction yet. Previously the checklist item stayed unchecked, giving a confusing experience. Now it's considered done once funds are available.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved first payment checklist detection to recognize available wallet funds in addition to completed transactions.
  * Updated the checklist item from "Send or receive your first payment" to "Receive your first payment" with refreshed description.

### DIFF
--- a/frontend/src/hooks/useOnboardingData.ts
+++ b/frontend/src/hooks/useOnboardingData.ts
@@ -3,6 +3,7 @@
 import { ALBY_ACCOUNT_APP_NAME } from "src/constants";
 import { useAlbyMe } from "src/hooks/useAlbyMe";
 import { useApps } from "src/hooks/useApps";
+import { useBalances } from "src/hooks/useBalances";
 import { useChannels } from "src/hooks/useChannels";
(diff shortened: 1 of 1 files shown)


rolznz (approved):
utACK

PR #2396, opened 2026-06-04
feat: show progress while exporting transactions

### LINKED ISSUE #2386: Improve Transactions Export
I clicked "export transactions" on my main wallet page. I have thousands of TXs. It just showed nothing (I thought it didn't work until I opened the network tab)

Also maybe it would be able to configure how many months to export.

reneaaron:
Closes #2386

## Problem

Clicking **Export Transactions** on a wallet with thousands of TXs appeared to do nothing — the handler silently paginated through every page (20 transactions per request) with no feedback, so it looked broken until the download eventually popped (only visible in the network tab).

## Changes

- Export now shows a **loading toast that updates with the running transaction count** (`Exporting 2,000 transactions…`), resolving to a success toast on completion and an error toast on failure.
- Fetches **1000 transactions per page instead of 20**, cutting round-trips for large wallets so the export finishes much faster.

No backend or API changes — this is purely the progress feedback for the existing export.

## Testing

- Frontend: `tsc --noEmit` and `yarn lint` pass.
- Manual: ran the hub in an isolated worktree with 750 seeded transactions and verified the toast updates with the count, the CSV downloads, and the success toast shows.

## Summary by CodeRabbit

* **New Features**
  * Transaction exports now display persistent progress notifications in real-time, updating with each pagination step to show the current number of transactions fetched
  * Enhanced error notifications and completion feedback provide clearer communication and status updates during the export process
  * Optimized pagination page size for more efficient and faster data retrieval during exports

### DIFF
--- a/frontend/src/components/transactions-utils.tsx
+++ b/frontend/src/components/transactions-utils.tsx
@@ -1,9 +1,12 @@
 import { toast } from "sonner";
-import { LIST_TRANSACTIONS_LIMIT } from "src/constants";
 import { ListTransactionsResponse, Transaction } from "src/types";
 import { request } from "src/utils/request";
 
 const LABEL_COLUMN_PREFIX = "label_";
+
+// Use a large page size to keep the number of round-trips low when
+// exporting wallets that have thousands of transactions.
+const EXPORT_TRANSACTIONS_PAGE_SIZE = 1000;
 
 // based on https://stackoverflow.com/a/68146412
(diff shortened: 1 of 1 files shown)


reneaaron:
@rolznz Export was so fast for my local test data (a few thousand transactions) that I didn't even see the loading toast. I think increasing the page size improved overall speed a lot, I left the progress indicator it in there anyway.

rolznz (approved):
tACK

PR #2410, opened 2026-06-07
fix: ignore 404 when deleting a lightning address (#2323)

### LINKED ISSUE #2323: Error toast shows when deleting sub-wallet if lightning address was already deleted from getalby.com
getalby.com API returns 404 - we can safely ignore that code.

Wutche:
Fixes #2323.

getalby.com returns `404` when the lightning address is already deleted, so we now ignore that status and the deletion succeeds instead of showing a "Failed to delete lightning address" error toast.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of lightning address deletion; the system now gracefully handles cases where the address has already been removed, treating this scenario as a successful operation rather than an error condition.

### DIFF
--- a/alby/alby_oauth_service.go
+++ b/alby/alby_oauth_service.go
@@ -418,6 +418,13 @@ func (svc *albyOAuthService) DeleteLightningAddress(ctx context.Context, address
 		return errors.New("failed to read response body")
 	}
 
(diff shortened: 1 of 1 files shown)


rolznz (approved):
utACK, thanks for the fix!

PR #2412, opened 2026-06-08
feat: generate CLN invoice preimage ourselves to support sub-wallets

### LINKED ISSUE #2411: CLN - add sub-wallet support
"sub-wallets are currently not supported on your node backend"

For this the only requirement is that the preimage is available at invoice creation time.

daywalker90:
Fixes: #2411

We can't generate a preimage for keysend in CLN. You only get it after payment is done.

## Summary by CodeRabbit

* **New Features**
  * Isolated app support now includes PHOENIX, BARK, and CLN backends in addition to existing options.
  * CLN backend now generates and returns invoice preimages for invoices.

* **Bug Fixes**
  * Improved randomness and retry jitter handling in CLN invoice workflows.
  * Error messages for unsupported backends updated to list the expanded supported set.

### DIFF
--- a/apps/apps_service.go
+++ b/apps/apps_service.go
@@ -59,9 +59,10 @@ func (svc *appsService) CreateApp(name string, pubkey string, maxAmountSat uint6
 		if backendType != config.LDKBackendType &&
 			backendType != config.LNDBackendType &&
 			backendType != config.PhoenixBackendType &&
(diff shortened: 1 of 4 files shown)


rolznz (approved):
utACK

Cool. I think it's ok - keysend is used much any more, and it will throw an error if a non-empty preimage is passed.

Thanks!

PR #2425, opened 2026-06-09
fix: reduce wallet balance/transaction polling interval to 10s

reneaaron:
## Summary

The wallet dashboard polls `/api/balances` and `/api/transactions` every **3 seconds** via SWR's `refreshInterval`. When a hub is left open in a browser tab, this generates a high, continuous stream of identical requests (and matching backend access logs) around the clock — with little UX benefit, since SWR already revalidates on window focus, so a longer interval keeps the data fresh without flooding the backend from long-lived open tabs.

This raises the polling interval for the balances and transactions-list hooks from **3s → 10s**.

## Changes

- `useBalances`: `refreshInterval` 3000 → 10000
- `useTransactions`: `refreshInterval` 3000 → 10000

The single-transaction hook (`useTransaction`, used while waiting for a specific invoice/payment to settle) is intentionally left at 3s — there fast updates matter and the polling is short-lived.

## Impact

- ~3x fewer balance/transaction poll requests from open dashboards, reducing backend load and log volume.
- Balance changes and received-payment notifications may surface up to ~10s later when the tab is already in focus; switching to the tab still triggers an immediate refresh.

## Summary by CodeRabbit

* **Chores**
  * Updated polling configuration for balance and transaction data refresh to reduce update frequency.

### DIFF
--- a/frontend/src/hooks/useBalances.ts
+++ b/frontend/src/hooks/useBalances.ts
@@ -4,7 +4,9 @@ import { BalancesResponse } from "src/types";
 import { swrFetcher } from "src/utils/swr";
 
 const pollConfiguration: SWRConfiguration = {
-  refreshInterval: 3000,
+  // SWR also revalidates on window focus, so a longer interval keeps the
+  // balance fresh without flooding the backend from long-lived open tabs.
+  refreshInterval: 10000,
 };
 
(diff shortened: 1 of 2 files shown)


rolznz (approved):
utACK

PR #2432, opened 2026-06-10
feat: add FeeRateField component to set fee rates for swaps

### LINKED ISSUE #2429: Allow user to configure fee rate when doing internal swaps
Follow-up from https://github.com/getAlby/hub/pull/2421

Allow user to always set a fee if we fail to fetch fee rates on the frontend.

Make this consistent across all places where we rely on fee rates - so that the user always has control over fee rate (important in times of high fee rates)

Also ensure the MempoolAlert is also shown in all places that rely on fee rates.

This should be done before confirming the swap (in the previous PR we removed fee retrieval which was done AFTER the user confirmed the swap).

im-adithya:
Fixes #2429

## Summary by CodeRabbit

* **New Features**
  * Added an editable on-chain fee field with view/edit mode, automatic initialization from recommended fees, validation, priority quick-selects, fee-estimate error hinting, and an optional “View on Mempool”.
  * Internal swap-in requests can now include a fee rate, and the internal flow surfaces it in the shared fee field.
* **Refactor**
  * Unified fee-rate editing across withdraw, on-chain send, and swap flows using the shared component.
* **Bug Fixes**
  * Updated on-chain wording and hyphenation (e.g., “Confirmed on-chain”).

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -997,7 +997,7 @@ func (api *api) GetSwapOutInfo() (*SwapInfoResponse, error) {
 	}, nil
 }
 
-func (api *api) InitiateSwapOut(ctx context.Context, initiateSwapOutRequest *InitiateSwapRequest) (*swaps.SwapResponse, error) {
+func (api *api) InitiateSwapOut(ctx context.Context, initiateSwapOutRequest *InitiateSwapOutRequest) (*swaps.SwapResponse, error) {
 	lnClient := api.svc.GetLNClient()
 	if lnClient == nil {
 		return nil, ErrLNClientNotStarted
@@ -1030,7 +1030,7 @@ func (api *api) InitiateSwapOut(ctx context.Context, initiateSwapOutRequest *Ini
 	return swapOutResponse, nil
 }
 
-func (api *api) InitiateSwapIn(ctx context.Context, initiateSwapInRequest *InitiateSwapRequest) (*swaps.SwapResponse, error) {
+func (api *api) InitiateSwapIn(ctx context.Context, initiateSwapInRequest *InitiateSwapInRequest) (*swaps.SwapResponse, error) {
 	lnClient := api.svc.GetLNClient()
 	if lnClient == nil {
 		return nil, ErrLNClientNotStarted
--- a/api/api.go
+++ b/api/api.go
@@ -1050,10 +1050,11 @@ func (api *api) InitiateSwapIn(ctx context.Context, initiateSwapInRequest *Initi
 		return nil, errors.New("invalid swap amount")
 	}
 
-	swapInResponse, err := api.svc.GetSwapsService().SwapIn(amountSat, false)
+	swapInResponse, err := api.svc.GetSwapsService().SwapIn(amountSat, false, initiateSwapInRequest.InternalPayment, initiateSwapInRequest.FeeRate)
 	if err != nil {
 		logger.Logger.WithFields(logrus.Fields{
-			"amount_sat": amountSat,
+			"amount_sat":       amountSat,
+			"internal_payment": initiateSwapInRequest.InternalPayment,
 		}).WithError(err).Error("Failed to initiate swap in")
 		return nil, err
 	}
--- a/api/models.go
+++ b/api/models.go
@@ -69,8 +69,8 @@ type API interface {
 	ListSwaps() (*ListSwapsResponse, error)
 	GetSwapInInfo() (*SwapInfoResponse, error)
 	GetSwapOutInfo() (*SwapInfoResponse, error)
-	InitiateSwapIn(ctx context.Context, initiateSwapInRequest *InitiateSwapRequest) (*swaps.SwapResponse, error)
-	InitiateSwapOut(ctx context.Context, initiateSwapOutRequest *InitiateSwapRequest) (*swaps.SwapResponse, error)
+	InitiateSwapIn(ctx context.Context, initiateSwapInRequest *InitiateSwapInRequest) (*swaps.SwapResponse, error)
+	InitiateSwapOut(ctx context.Context, initiateSwapOutRequest *InitiateSwapOutRequest) (*swaps.SwapResponse, error)
 	RefundSwap(refundSwapRequest *RefundSwapRequest) error
 	GetSwapMnemonic() string
 	GetAutoSwapConfig() (*GetAutoSwapConfigResponse, error)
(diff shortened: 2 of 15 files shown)


frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -212,8 +210,9 @@ function OnchainForm({
       <div className="grid gap-2 text-sm border-t pt-6">
         {!editFee ? (
           <div className="flex items-center justify-between">
-            <p className="text-muted-foreground">On-chain Fee Rate</p>
-            <div
+            <Label>On-chain Fee Rate (sat/vB)</Label>
```
rolznz: Couldn't all this (including the edit button etc) be in this new component?

frontend/src/screens/wallet/swap/SwapInStatus.tsx
```diff
@@ -70,6 +71,7 @@ export default function SwapInStatus() {
         const payload: RedeemOnchainFundsRequest = {
           toAddress: swap.lockupAddress,
           amountSat: swap.sendAmountSat,
+          feeRate: +feeRate,
```
rolznz: the backend accepts a nullable value, I think we should only pass feeRate here if it's non-empty

`+""` will turn into 0 which will still be passed.

frontend/src/screens/wallet/send/Onchain.tsx
im-adithya: Done

frontend/src/screens/wallet/swap/SwapInStatus.tsx
im-adithya: Yeah I left it that way because after this change it can never be nil / `""` because the onchain fee rate is filled from the mempool api by default (or if it fails the field is `required`)

But will change this just to be safe ✅

frontend/src/screens/wallet/swap/index.tsx
```diff
@@ -142,7 +144,11 @@ function SwapInForm() {
         throw new Error("Error swapping in");
       }
       navigate(
-        `/wallet/swap/in/status/${swapInResponse.swapId}${swapFrom === "internal" ? "?internal=true" : ""}`
+        `/wallet/swap/in/status/${swapInResponse.swapId}${
+          swapFrom === "internal"
+            ? `?internal=true&feeRate=${encodeURIComponent(feeRate)}`
```
rolznz: Isn't it weird actually we pass the fee rate here rather than to the swap endpoint itself when creating the swap?

frontend/src/screens/wallet/send/Onchain.tsx
```diff
@@ -210,74 +187,7 @@ function OnchainForm({
...
-                >
-                  View on Mempool
-                  <ExternalLinkIcon className="w-4 h-4" />
-                </ExternalLink>
-              </div>
-            )}
-          </div>
-        )}
+        <FeeRateField feeRate={feeRate} onFeeRateChange={setFeeRate} />
       </div>
```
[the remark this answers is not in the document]
rolznz: @im-adithya I think we should remove the backendType check in frontend/src/screens/wallet/WithdrawOnchainFunds.tsx right?

frontend/src/screens/wallet/swap/index.tsx
rolznz: Then we can remove this effect that runs on page load (which is potentially a bit dodgy)

rolznz:
Nice change! :muscle:

frontend/src/screens/wallet/swap/index.tsx
rolznz: For this we should add a new parameter (e.g. ?internal=true to the api call) so that by default it works the same as now (for backward compatibility / hub CLI)

im-adithya: I added it to `InitiateSwapRequest` as `InternalPayment`

frontend/src/screens/wallet/send/Onchain.tsx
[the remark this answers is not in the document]
im-adithya: Done!

api/api.go
```diff
@@ -1050,10 +1050,11 @@ func (api *api) InitiateSwapIn(ctx context.Context, initiateSwapInRequest *Initi
...
 	}
 
-	swapInResponse, err := api.svc.GetSwapsService().SwapIn(amountSat, false)
+	swapInResponse, err := api.svc.GetSwapsService().SwapIn(amountSat, false, initiateSwapInRequest.InternalPayment, initiateSwapInRequest.FeeRate)
 	if err != nil {
 		logger.Logger.WithFields(logrus.Fields{
-			"amount_sat": amountSat,
+			"amount_sat":       amountSat,
+			"internal_payment": initiateSwapInRequest.InternalPayment,
 		}).WithError(err).Error("Failed to initiate swap in")
```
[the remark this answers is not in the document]
im-adithya: `RedeemOnchainFunds` function takes care of it, we don't have to reject beforehand

frontend/src/screens/wallet/swap/index.tsx
```diff
@@ -130,6 +132,12 @@ function SwapInForm() {
 
       const payload: InitiateSwapRequest = {
         swapAmountSat: parseInt(swapAmountSat),
+        ...(swapFrom === "internal"
+          ? {
+              internalPayment: true,
+              ...(feeRate ? { feeRate: parseInt(feeRate) } : {}),
+            }
+          : {}),
```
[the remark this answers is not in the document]
im-adithya: Fixed to align with other places using `+feeRate`, since we use FeeRateField with an Input with step 1 and min as atleast 1 we should be fine

api/models.go
```diff
@@ -169,9 +169,11 @@ type CreateLightningAddressRequest struct {
 }
 
 type InitiateSwapRequest struct {
-	SwapAmount    *uint64 `json:"swapAmount"` // deprecated
-	SwapAmountSat *uint64 `json:"swapAmountSat"`
-	Destination   string  `json:"destination"`
+	SwapAmount      *uint64 `json:"swapAmount"` // deprecated
```
rolznz, line 172: Can we rename the InitiateSwapRequest to `InitiateSwapInRequest` so it's clear

frontend/src/screens/wallet/swap/index.tsx
```diff
@@ -284,12 +294,17 @@ function SwapInForm() {
...
-            </p>
+          <div className="flex flex-col pt-4 gap-4 border-t">
+            {isInternalSwap && (
+              <FeeRateField feeRate={feeRate} onFeeRateChange={setFeeRate} />
+            )}
+            <div className="flex items-center justify-between">
+              <Label>{isInternalSwap ? "Swap Fee" : "Fee"}</Label>
+              <p className="text-muted-foreground text-sm">
+                {swapInfo.albyServiceFee + swapInfo.boltzServiceFee}%
+                {!isInternalSwap && " + on-chain fees"}
```
rolznz: couldn't we still show this and remove the conditional? because you set a fee rate, but probably won't know exactly how much you are charged

rolznz (approved):
utACK, one minor comment

frontend/src/screens/wallet/swap/index.tsx
im-adithya: Makes sense, it's just the rate :+1: Removed!

api/models.go
im-adithya: Done!

PR #2436, opened 2026-06-10
fix: filter keysend preimage from custom TLVs

### LINKED ISSUE #407: Bug: failure to make payments on podverse due to it including a custom TLV for the payment preimage
Podverse includes this TLV, when I believe it should not:
5482373484	keysend	32	Preimage as 32 bytes

Normally the node should inject this manually when making a keysend payment. It's not a "custom" TLV as an app should inject, it's a requirement for doing keysend, and rust-lightning rejects it from being passed in.

Both LND and LDK set this TLV themselves.

I think actually we should contact podverse and see if they can remove this TLV, otherwise we will need to manually filter it out on the Alby Hub side.

Anthonyushie:
## Summary

Fixes #407.

Filters the keysend preimage TLV from custom TLV records before sending payments. This prevents LDK keysend payments from failing while preserving other custom TLVs.

## Testing

- Added tests for generated and custom preimages
- Added tests confirming other custom TLVs are preserved
- Keysend transaction and controller tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed keysend payment handling to properly filter preimage records from custom metadata, ensuring only intended custom records are included in keysend transactions.

* **Tests**
  * Added comprehensive test coverage for keysend preimage record filtering behavior.

### DIFF
--- a/lnclient/lnd/lnd.go
+++ b/lnclient/lnd/lnd.go
@@ -530,8 +530,7 @@ func (svc *LNDService) SendKeysend(amountMsat uint64, destination string, custom
 		destCustomRecords[record.Type] = decodedValue
 	}
 	const MAX_PARTIAL_PAYMENTS = 16
-	const KEYSEND_CUSTOM_RECORD = 5482373484
-	destCustomRecords[KEYSEND_CUSTOM_RECORD] = preImageBytes
+	destCustomRecords[lnclient.KeysendPreimageTlvType] = preImageBytes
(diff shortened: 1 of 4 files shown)


rolznz:
Hi, thanks for the PR - but as in the issue I believe this should not be fixed on the hub side.

Anthonyushie:
Understood, thanks for clarifying. The current NIP-47 pay_keysend example includes 5482373484 in tlv_records, which led me to interpret it as valid input. I’ll close this PR

PR #2447, opened 2026-06-12
fix: fallback to outgoing payments in Phoenixd LookupInvoice

### LINKED ISSUE #2442: Phoenixd: outgoing payments permanently stuck as PENDING (LookupInvoice only queries /payments/incoming)
## Bug

Outgoing Lightning payments via Phoenixd backend are permanently displayed as **PENDING** in Alby Hub, even after Phoenixd confirms they are settled (`isPaid=true` with valid preimage).

## Root Cause

`LookupInvoice` in `lnclient/phoenixd/phoenixd.go:262` **always** queries `/payments/incoming/{paymentHash}`, even for outgoing payments:

```go
func (svc *PhoenixService) LookupInvoice(ctx context.Context, paymentHash string) (transaction *lnclient.Transaction, err error) {
    req, err := http.NewRequestWithContext(ctx, http.MethodGet, svc.Address+"/payments/incoming/"+paymentHash, nil)
    // ...
    if resp.StatusCode != http.StatusOK {
            "body":        string(body),
            "status_code": resp.StatusCode,
        }).Error("phoenixd incoming payments endpoint returned non-success code")
        return nil, fmt.Errorf("phoenixd incoming payments endpoint returned non-success code: %d %s", resp.StatusCode, string(body))
    }
```

For outgoing payments, `/payments/incoming/{hash}` returns **404** because the payment exists under `/payments/outgoing`. The reconciliation loop (`transactions_service.go:709`) then:
1. Calls `LookupInvoice` for the pending outgoing payment
2. Gets an error (404)
3. Skips the payment, leaving it as `PENDING`

Additionally, reconciliation only runs for transactions < 24h old, so older stuck payments are **permanently abandoned**.

## Evidence (v1.22.2 + Phoenixd v0.6.0)

Logs repeating every ~3 seconds on every `/api/transactions` call:

```
[ERRO] phoenixd incoming payments endpoint returned non-success code body="Not found" status_code=404
[ERRO] Failed to check transaction bolt11="lnbc347180n1p4z4x0m..."
```

All 20 outgoing payments confirmed `isPaid=true` at Phoenixd level with valid preimages. Zero are stuck at the node layer.

## Impact

- Every outgoing payment via Phoenixd shows as PENDING forever in Alby Hub UI
- Accumulates over time — older payments are never rechecked
- Users cannot distinguish between actually stuck payments and this display bug
- The 404 error spam fills logs (~1 error per second per active browser session)

## Suggested Fix

When `/payments/incoming/{hash}` returns 404, fall back to searching `/payments/outgoing` for the same payment hash:

```go
func (svc *PhoenixService) LookupInvoice(ctx context.Context, paymentHash string) (transaction *lnclient.Transaction, err error) {
    // Try incoming first (existing behavior)
    tx, err := svc.lookupIncoming(ctx, paymentHash)
    if err == nil {
        return tx, nil
    }
    // Fallback: search outgoing payments by hash
    return svc.lookupOutgoing(ctx, paymentHash)
}
```

Note: Phoenixd's `/payments/outgoing/{id}` endpoint uses UUID identifiers, not payment hashes. The fallback would need to list outgoing payments (`/payments/outgoing`) and filter by `paymentHash` field.

## Environment

- Alby Hub v1.22.2 (latest release)
- Phoenixd v0.6.0-fa318d9
- Backend: Phoenixd over HTTPS (remote)
- Deployed on Railway

## Workaround

We built a CLI tool that cross-references the Alby Hub SQLite DB against Phoenixd and updates stuck transactions directly:

```bash
# Pull Alby Hub DB, cross-reference with Phoenixd, fix stuck payments
python phoenix_cli.py desync
python phoenix_cli.py fix --apply --redeploy
```

This is a temporary workaround — the fix should be in the connector itself.

G12c4:
## Problem

Outgoing Lightning payments via Phoenixd backend are permanently displayed as **PENDING** in Alby Hub, even after Phoenixd confirms they are settled (`isPaid=true`).

Closes #2442

## Root Cause

`LookupInvoice` only queries `/payments/incoming/{paymentHash}`. For outgoing payments this returns 404, and the reconciliation loop in `transactions_service.go:718` skips the payment, leaving it as PENDING forever.

## Fix

When `/payments/incoming/{hash}` fails, fall back to `/payments/outgoingbyhash/{hash}` — a dedicated Phoenixd endpoint that looks up outgoing payments directly by payment hash (added in Phoenixd v0.x). This is an O(1) lookup, not a list+scan.

The `LNClient.LookupInvoice` interface accepts only a `paymentHash` with no transaction type hint. Ideally it would accept a type so we could skip the incoming probe for known-outgoing payments, but changing the `LNClient` interface touches every backend (LND, LDK, CLN, Bark, Cashu) and is out of scope for this bugfix. The incoming→outgoing fallback is safe, minimal, and preserves all existing behavior.

## Changes

- **`LookupInvoice`** — tries incoming first (preserving existing behavior), then falls back to outgoing
- **`lookupOutgoingPayment`** — uses `/payments/outgoingbyhash/{hash}` for direct O(1) lookup
- **`outgoingPaymentToTransaction`** — converts `OutgoingPaymentResponse` to `lnclient.Transaction` (mirrors existing `phoenixInvoiceToTransaction`)
- Removed noisy `Error()` log on incoming 404 — the error is still returned and logged by the caller in `checkUnsettledTransaction`

## Testing

- Compiles cleanly (`go build ./lnclient/phoenixd/`)
- No existing tests for the Phoenixd connector — only the LND wrapper has tests
- Verified the 404 pattern in production logs (v1.22.2 + Phoenixd v0.6.0)
- All 20 outgoing payments confirmed `isPaid=true` at Phoenixd level while Alby Hub showed them as PENDING
- Used the `/payments/outgoingbyhash` endpoint from Phoenixd source (`Api.kt`) — returns the best match (succeeded > pending > failed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved invoice lookup reliability by falling back to outgoing payment data when an incoming invoice is not found.
  * Outgoing payment results now include transaction amounts, fees, timestamps, settlement status, and descriptions.
  * Improved error messages provide clearer context for failed incoming and outgoing payment lookups.
  * Lookup fallback behavior now avoids masking errors from the incoming payment service.


rolznz (approved):
utACK

rolznz:
@coderabbitai review

PR #2455, opened 2026-06-16
fix: handle unavailable node

### LINKED ISSUE #410: Improve frontend when node is unavailable
Follow up to #404

We moved forward with not stopping the Hub when the backend is down. But during the downtime, our frontend looks pretty bad. We should handle the 500 errors from the LNClient better.

### Home
Screenshot 2024-08-06 at 4 02 29 PM

### Wallet
Screenshot 2024-08-06 at 4 02 38 PM

### Connections (minor)
Screenshot 2024-08-06 at 4 03 08 PM

### Node
Screenshot 2024-08-06 at 4 02 20 PM

Wutche:
## Description

Shows a clear unavailable state when Alby Hub cannot reach the lightning node, instead of leaving node-dependent pages stuck loading or showing broken API errors.

The unavailable state is shown for core node-backed routes like Home, Wallet, Node/Channels, and Peers. Settings and Connections remain accessible so users can still inspect configuration or recover.

Fixes #410

## Changes

- Added a shared `/api/node/status` polling hook
- Added a reusable `NodeUnavailable` screen with retry and Node settings actions
- Updated route handling to show the unavailable screen only after Hub is otherwise accessible
- Added the frontend `NodeStatus` type

## Summary by CodeRabbit

* **New Features**
  * Added unavailable node state that displays when a lightning node connection is lost, with options to retry connection or navigate to settings
  * Implemented node availability validation on key app routes; access to wallet, channels, peers, and home sections now requires an active node connection

### DIFF
--- a/frontend/src/components/NodeUnavailable.tsx
+++ b/frontend/src/components/NodeUnavailable.tsx
@@ -1,0 +1,38 @@
+import { RefreshCwIcon, ServerOffIcon, SettingsIcon } from "lucide-react";
+import AppHeader from "src/components/AppHeader";
+import { Button } from "src/components/ui/button";
+import { LinkButton } from "src/components/ui/custom/link-button";
+
+type Props = {
+  onRetry: () => void;
+};
+
+export function NodeUnavailable({ onRetry }: Props) {
+  return (
+    <>
+      <AppHeader title="Node unavailable" pageTitle="Node unavailable" />
+      <div className="flex flex-1 items-center justify-center rounded-lg bg-muted p-8">
+        <div className="flex max-w-md flex-col items-center gap-1 text-center">
+          <ServerOffIcon className="size-10 text-muted-foreground" />
+          <h2 className="mt-4 text-lg font-semibold">
+            Your lightning node is unavailable
+          </h2>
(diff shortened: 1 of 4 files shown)


rolznz:
Hi @Wutche thanks for the PR. We now have a "Healthcheck" which will show a red dot in the navbar. Could this be sufficient? Did you test this locally?

Wutche:
> Hi @Wutche thanks for the PR. We now have a "Healthcheck" which will show a red dot in the navbar. Could this be sufficient? Did you test this locally?

Hello, Sorry I didn't notice that when I was testing. I will fix that rn

Wutche:
@rolznz do you want me to keep the "healthcheck"? or I should live cos looking at it, its not that bad when i tested locally it worked fine. what do you think?

Wutche:
@rolznz 
image
do you have any alternative other than the red dot?? i applied it as an indicator that something is wrong

rolznz:
@Wutche yeah, this is a very old issue, I think the healthcheck indicator (red dot) already solves it. So maybe we can close this one. Does that sound good to you?

Wutche:
Yeah sure it sounds good I'll contribute to another one

PR #2473, opened 2026-07-10
fix: require full access api key for swaps/mnemonic

bumi:
the mnemonic could be considered a non read-only route because the mnemonic could be used.
This moves this route to the full access group to require a full access api key.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted access to the mnemonic swap endpoint to full-access requests.
  * Read-only API access can no longer retrieve mnemonic swap data.


rolznz (approved):
utACK

PR #2477, opened 2026-07-25
chore: update bark bindings to v0.12.1

PeteClubSeven:
Hi, I noticed the bark version is really old and the 0.4.0 SDK release has significant robustness improvements which users need for a better experience. I haven't written Go in quite a long time so I had Claude help me, if anything seems wrong with this I'm happy to rework it.

One thing to note, we wrote a huge amount of the payment code on the client side and a side effect is that we only migrate settled lightning receive payments, in-flight incoming payments will be dropped.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Lightning payment tracking by relying on payment states to interpret success vs failure, and by ignoring non-success movements.
  - Fixed incoming/outgoing preimage handling and corrected settlement timestamps and fee reporting using state data (with safer handling for missing values).
- **Enhancements**
  - Updated Bitcoin wallet integration to a newer Bark bindings version.
  - Improved wallet startup flow and invoice creation to reliably retrieve required preimage details.

### DIFF
--- a/lnclient/bark/bark.go
+++ b/lnclient/bark/bark.go
@@ -4,10 +4,8 @@ package bark
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -31,11 +29,14 @@ const (
 	// Subsystem name reported on movements produced for outgoing lightning
 	// payments (see bark's Subsystem::LIGHTNING_SEND).
 	lightningSendSubsystem = "lightning_send"
+	// The status a movement is created with; every other status is terminal.
+	movementStatusPending = "pending"
 	// Movement status reported once a movement has settled. A movement first
 	// appears as "pending" and is updated to this once complete.
 	movementStatusSuccessful = "successful"
-	// Movement status reported when a send was definitively not paid.
-	movementStatusFailed = "failed"
+	// LightningReceive.State values in which we hold the preimage.
+	receiveStatePreimageRevealed = "preimage-revealed"
+	receiveStateSettled          = "settled"
 	// Grace period to allow the notification loop to unwind on shutdown.
 	shutdownGracePeriod = 10 * time.Second
 )
@@ -123,7 +124,6 @@ func NewBarkService(ctx context.Context, eventPublisher events.EventPublisher, w
 
 	cfg := bark.Config{
 		ServerAddress:                config.ServerAddress,
-		Network:                      network,
 		RoundTxRequiredConfirmations: &roundTxRequiredConfirmations,
 	}
 	esploraAddress := config.EsploraAddress
@@ -135,23 +135,7 @@ func NewBarkService(ctx context.Context, eventPublisher events.EventPublisher, w
 		cfg.ServerAccessToken = &token
 	}
 
-	_, statErr := os.Stat(workDir)
-	isFirstSetup := statErr != nil && errors.Is(statErr, os.ErrNotExist)
-
-	logger.Logger.WithFields(logrus.Fields{
-		"workDir":      workDir,
-		"isFirstSetup": isFirstSetup,
-	}).Info("Opening Bark wallet")
-
-	var wallet *bark.Wallet
-	if isFirstSetup {
-		wallet, err = bark.WalletCreate(mnemonic, cfg, workDir, false)
-	} else {
-		wallet, err = bark.WalletOpen(mnemonic, cfg, workDir)
-	}
-	if err != nil {
-		return nil, fmt.Errorf("failed to open bark wallet: %w", err)
-	}
+	logger.Logger.WithField("workDir", workDir).Info("Opening Bark wallet")
 
 	// Bark provides a built-in background daemon that periodically syncs with
 	// the Ark server and blockchain, participates in rounds, and — crucially for
@@ -159,8 +143,13 @@ func NewBarkService(ctx context.Context, eventPublisher events.EventPublisher, w
 	// payment notifications and reveals the preimage, crediting the balance). We
 	// don't poll for receives ourselves; instead we observe the resulting wallet
 	// notifications (see runNotificationLoop) to emit payment-received events.
-	if err := wallet.RunDaemon(nil); err != nil {
-		logger.Logger.WithError(err).Warn("Bark daemon failed to start")
+	wallet, err := bark.WalletOpen(network, mnemonic, cfg, bark.WalletOpenArgs{
+		Datadir:           workDir,
+		RunDaemon:         true,
+		CreateIfNotExists: true,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to open bark wallet: %w", err)
 	}
 
 	loopCtx, cancelFn := context.WithCancel(context.Background())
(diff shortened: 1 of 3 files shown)


lnclient/bark/bark.go
```diff
@@ -143,12 +145,16 @@ func NewBarkService(ctx context.Context, eventPublisher events.EventPublisher, w
...
-		wallet, err = bark.WalletOpen(mnemonic, cfg, workDir)
-	}
+	// CreateIfNotExists replaces the separate WalletCreate call, so creation
+	// stays gated on the work dir being absent. The daemon is started separately
+	// below rather than via RunDaemon here, so that a daemon that fails to start
+	// leaves us with a usable wallet instead of failing the open outright.
+	wallet, err := bark.WalletOpen(network, mnemonic, cfg, bark.WalletOpenArgs{
+		Datadir:             workDir,
+		RunDaemon:           false,
+		CreateIfNotExists:   isFirstSetup,
```
rolznz: I think we can just put `true` here and remove `isFirstSetup` completely. That was for the conditional which is not needed any more

lnclient/bark/bark.go
```diff
@@ -143,12 +145,16 @@ func NewBarkService(ctx context.Context, eventPublisher events.EventPublisher, w
...
+		CreateWithoutServer: false,
```
rolznz: this is the default value, so not needed

lnclient/bark/bark.go
```diff
@@ -143,12 +145,16 @@ func NewBarkService(ctx context.Context, eventPublisher events.EventPublisher, w
...
-	} else {
-		wallet, err = bark.WalletOpen(mnemonic, cfg, workDir)
-	}
+	// CreateIfNotExists replaces the separate WalletCreate call, so creation
+	// stays gated on the work dir being absent. The daemon is started separately
+	// below rather than via RunDaemon here, so that a daemon that fails to start
+	// leaves us with a usable wallet instead of failing the open outright.
+	wallet, err := bark.WalletOpen(network, mnemonic, cfg, bark.WalletOpenArgs{
+		Datadir:             workDir,
+		RunDaemon:           false,
```
rolznz: I think this should be true so we don't need to manually call sync. Why is it set to false? can we set it to true and remove the line below which manually runs the daemon now?

lnclient/bark/bark.go
```diff
@@ -296,62 +305,85 @@ func (bs *BarkService) handleLightningReceiveMovement(movement bark.Movement) {
...
 
-	if movement.Status == movementStatusFailed {
-		bs.deliverSendResult(meta.PaymentHash, sendResult{err: errors.New("bark lightning send failed")}, func() {
+	// The movement can be canceled or failed so we should just check if it
+	// wasn't successful.
+	if movement.Status != movementStatusSuccessful {
+		reason := fmt.Sprintf("bark lightning send %s", movement.Status)
+		logger.Logger.WithFields(logrus.Fields{
+			"paymentHash": paymentHash,
+			"reason":      reason,
```
rolznz, line 306: could we log "status" here too

lnclient/bark/bark.go
```diff
@@ -296,62 +305,85 @@ func (bs *BarkService) handleLightningReceiveMovement(movement bark.Movement) {
...
 			},
 		})
 	})
 }
 
+// settledSendPreimage reads the preimage of a settled lightning send from the
+// send's own state rather than from the movement. bark records the paid invoice
+// before it finishes the movement, so the preimage is already persisted by the
+// time the successful movement is observed here.
+func (bs *BarkService) settledSendPreimage(paymentHash string) (string, error) {
```
rolznz: minor: can you start with a verb `getSettledSendPreimage`

lnclient/bark/bark.go
```diff
@@ -365,17 +397,14 @@ func (bs *BarkService) deliverSendResult(paymentHash string, res sendResult, fal
 }
 
 func paymentHashFromMovement(movement bark.Movement) (string, bool) {
-	var meta struct {
-		PaymentHash string `json:"payment_hash"`
-	}
-	if err := json.Unmarshal([]byte(movement.MetadataJson), &meta); err != nil || meta.PaymentHash == "" {
-		logger.Logger.WithError(err).WithFields(logrus.Fields{
+	if movement.PaymentHash == nil || *movement.PaymentHash == "" {
```
rolznz, line 382: this function is only used once - now that the payment hash is directly accessible (not through metadata json) I think we don't need this function any more

lnclient/bark/bark.go
```diff
@@ -534,7 +565,7 @@ func (bs *BarkService) LookupInvoice(ctx context.Context, paymentHash string) (*
 	return nil, errors.New("this method should not be called")
 }
 
-func (bs *BarkService) lightningReceiveToTransaction(receive *bark.LightningReceive) (*lnclient.Transaction, error) {
+func (bs *BarkService) lightningReceiveToTransaction(receive bark.LightningReceive) (*lnclient.Transaction, error) {
```
rolznz: minor: normally we pass a pointer here so we don't copy the whole object (for `receive`)

lnclient/bark/bark.go
```diff
@@ -551,10 +582,18 @@ func (bs *BarkService) lightningReceiveToTransaction(receive *bark.LightningRece
 		Description:     paymentRequest.Description,
 		DescriptionHash: paymentRequest.DescriptionHash,
 	}
-	if receive.PreimageRevealed {
-		now := time.Now().Unix()
-		tx.SettledAt = &now
-		tx.Preimage = receive.PaymentPreimage
+	if receive.State == receiveStatePreimageRevealed || receive.State == receiveStateSettled {
```
rolznz, line 566: do we need to check both? does the state always go to `receiveStatePreimageRevealed` first? if so can we just check this one?

rolznz: the code below becomes more complicated due to this change. If possible we should avoid it.

rolznz:
Hi @PeteClubSeven thanks for the PR! I left some minor comments.

lnclient/bark/bark.go
PeteClubSeven: I left it as false since there was already a chunk of code specifically starting the daemon but I'm happy to remove it and just set this to true 👍

lnclient/bark/bark.go
PeteClubSeven: It's used twice by `handleLightningReceiveMovement` and `handleLightningSendMovement`. Happy to inline it at both sites if you'd prefer.

lnclient/bark/bark.go
PeteClubSeven: We need both. Bark marks the movement `Successful` before it writes the settled-receive record, and `lightning_receive_state` checks that record first and falls back to the in-progress checkpoint — so depending on which side of that write we observe the movement, we get `preimage-revealed` or `settled` for the same condition. No funds are at risk either way, but I'd rather not depend on the timing.

I'll raise this internally so we can reduce the complexity here later.

PeteClubSeven:
> Hi @PeteClubSeven thanks for the PR! I left some minor comments.

Updated based on your comments. Let me know if you want any other changes.

lnclient/bark/bark.go
rolznz: Ah ok, no need to move it then. Thanks!

rolznz (approved):
tACK

PR #2501, opened 2026-08-05
fix: show icon and proper name for lightning node backend on about page

### LINKED ISSUE #2445: Improve lightning node display on about page
On the initial setup it looks good, but on the about page it's very basic. We could show the icon and proper casing.

Image

Image

rolznz:
Closes #2445

The About page showed the raw backend type (e.g. `LDK`) as plain text. It now shows the backend's icon and properly-cased name, matching the display used during initial setup.

- Moved each backend's `title` and `icon` into the shared `backendTypeConfigs` in `src/lib/backendType.tsx` (renamed from `.ts` for JSX) so the setup screen and About page use one source of truth
- `SetupNode` now consumes the shared config instead of its local duplicate — no visual change there
- About page renders the icon + title in the Lightning Node Backend section

## Summary by CodeRabbit

- **UI Improvements**
  - Added recognizable icons and display names for supported backend types.
  - Updated the setup flow to use consistent backend labels and icons.
  - Improved the About screen by showing the configured backend’s name and icon instead of its internal type value.


reneaaron (approved):
tACK

PR #2502, opened 2026-08-05
fix: Mandatory upgrade for bark SDK

PeteClubSeven:
Bumps the bark bindings from `v0.12.1` (bark 0.4.0) to `v0.15.0` (bark 0.6.0).

**This is urgent: users on < bark 0.5.0 are unable to board, and users on < 0.6.0 cannot join rounds (so cannot refresh VTXOs) or use lightning.**

We apologize for having to make a breaking release but we've been extensively testing our codebase and responding to security vulnerabilities, as have many projects in bitcoin right now. As such we made the choice to break some existing functionality and push an emergency update.

No client API changes necessary, the new SDK ships the necessary client updates for users to upgrade to the new mandatory protocol version to enable all features again.

Current clients are able to make ark payments, offboard and unilaterally exit.

As an added bonus the users get the following by upgrading from the previous release:

- **Wallet recovery from seed alone.** VTXO IDs are posted to a seed-derived recovery mailbox, so a wallet can rebuild its spendable set from the mnemonic with no local state. Runs automatically on creation.
- **Crash-safe payments** Each incoming/outgoing payment type now runs as a resumable wallet action with a persisted checkpoint, so an interrupted operation resumes on the next sync instead of facing issues.
- **Fewer stuck funds.** Maintenance refresh no longer wedges on a single unusable VTXO, force-exited VTXOs are detected and recovered, claimable exits are counted in the pending exit balance, and clients no longer disconnect from the server every 10 minutes.
- **LNURL-pay support**
- Plus hardened permissions on the wallet database and filestore.

## Summary by CodeRabbit

* **Chores**
  * Updated an internal Bitcoin integration dependency to a newer version for improved compatibility and maintenance.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	github.com/tyler-smith/go-bip39 v1.1.0
 	github.com/wailsapp/wails/v2 v2.12.0
-	gitlab.com/ark-bitcoin/bark-ffi-bindings/golang v0.12.1
+	gitlab.com/ark-bitcoin/bark-ffi-bindings/golang v0.15.0
 	golang.org/x/crypto v0.54.0
 	golang.org/x/oauth2 v0.36.0
 	google.golang.org/grpc v1.79.3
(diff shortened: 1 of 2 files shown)


PeteClubSeven:
Pinging @rolznz for visibility

PR #2510, opened 2026-08-06
fix: use scope constant in get_budget permission query

### LINKED ISSUE #2503: get_budget queries scope column with method constant
`nip47/controllers/get_budget_controller.go:29`:

```go
controller.db.Where("app_id = ? AND scope = ?", app.ID, models.PAY_INVOICE_METHOD).First(&appPermission)
```

Two small correctness issues here:

**1. The query filters a scope column with a method constant.** `app_permissions.scope` holds a scope, whose constant is `constants.PAY_INVOICE_SCOPE`, but the query passes `models.PAY_INVOICE_METHOD`. Both currently evaluate to `"pay_invoice"`, so the query works — but only by coincidence. The two are conceptually distinct (`constants.PAY_INVOICE_SCOPE` is documented as also covering `pay_keysend` and the `multi_*` payment methods), and if either constant were ever renamed this query would silently match nothing and every `get_budget` call would start reporting "no budget" instead of the real one. Every other scope lookup uses the `constants.*_SCOPE` values.

**2. The unchecked `First` result deserves a comment.** If no matching permission row exists, `appPermission` stays zero-valued, `maxAmountSat == 0`, and the handler returns the empty "no budget" result — deliberately the same response as a permission that exists with a zero max amount. That collapse is intentional and correct, but it isn't obvious from the code that ignoring the `First` result is safe. A short comment noting that both cases (no `pay_invoice` permission, or a permission with no budget set) return the same empty response would save the next reader the trace. (The one case that does differ is a genuine DB error, which also falls through to "no budget" rather than surfacing an error — minor, and only worth handling if that distinction matters to callers.)

Suggested fix: use `constants.PAY_INVOICE_SCOPE` in the `Where` clause, and add the clarifying comment above the `maxAmountSat == 0` branch.

rolznz:
Fixes #2503

The `get_budget` controller filtered the `app_permissions.scope` column with `models.PAY_INVOICE_METHOD`. That only worked because the method constant and `constants.PAY_INVOICE_SCOPE` currently share the string value `"pay_invoice"` — the two are conceptually distinct, and every other scope lookup in the codebase uses the `constants.*_SCOPE` values.

Changes:
- Query the scope column with `constants.PAY_INVOICE_SCOPE` instead of `models.PAY_INVOICE_METHOD` (no behavior change).
- Add a comment explaining why the unchecked `First` result is intentional: a missing `pay_invoice` permission and a permission with no budget set both deliberately return the same empty "no budget" response.

Verified with `go build ./nip47/...` and `go test ./nip47/controllers/`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved budget permission lookup behavior for invoice payments.
  * Missing or unavailable permissions now return a consistent empty response, matching zero-budget permissions.
  * Database failures during budget permission checks are now handled correctly and reported as internal errors instead of producing misleading results.

### DIFF
--- a/nip47/controllers/get_budget_controller.go
+++ b/nip47/controllers/get_budget_controller.go
@@ -2,9 +2,11 @@ package controllers
 
 import (
 	"context"
+	"errors"
 
 	"github.com/getAlby/go-nostr"
 	"github.com/getAlby/hub/db/queries"
+	"gorm.io/gorm"
 
 	"github.com/getAlby/hub/constants"
 	"github.com/getAlby/hub/db"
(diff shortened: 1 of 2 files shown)


rolznz:
@coderabbitai review

rolznz:
@coderabbitai review

PR #2520, opened 2026-08-07
fix: publish transaction events only after the database transaction commits

### LINKED ISSUE #2506: Payment settled/failed events are published before the enclosing DB transaction commits
`markTransactionSettled` and `markPaymentFailed` publish their events while still inside the caller's database transaction, so connected apps and the analytics pipeline can be notified of a payment that never commits.

`markTransactionSettled` (`transactions/transactions_service.go:1450`) takes the transaction-scoped handle and publishes at `:1501-1504`:

```go
	Event:      event,          // nwc_payment_sent / nwc_payment_received
	Properties: dbTransaction,
})

if dbTransaction.AppId != nil {
	...
	svc.updateAppLastSettledTransactionAt(&app, tx, &settledAt)   // more statements on tx, after the publish
}
```

`markPaymentFailed` (`:1566`) has the same shape at `:1595-1598`, and `checkBudgetUsage` — called at `:1519`, also pre-commit — can publish `nwc_budget_warning` at `:1556`.

Every caller passes `tx` from inside an `svc.db.Transaction(...)` closure, so `COMMIT` has not run yet:

- `:426`, `:435` — `SendPaymentSync`
- `:577` — `SendKeysend`
- `:745` — `checkUnsettledTransaction`
- `:821`, `:915`, `:947` — the `nwc_lnclient_payment_received` / `_sent` / `_failed` `ConsumeEvent` handlers
- `:1044` — `interceptSelfPayment`
- `:1308` — `SettleHoldInvoice`
- `:1369` — `CancelHoldInvoice`

Publication is asynchronous and immediate — `events/events.go:48-50` → `:64`, `go listener.ConsumeEvent(...)` per subscriber — and subscribers send NIP-47 notifications to connected apps over Nostr and forward events to the Alby API. So the external side effects can fire before the row is durable, and they fire regardless of whether the transaction goes on to commit.

If the enclosing transaction fails — a SQLite `database is locked`, a Postgres reset or deadlock, or an error in the statements that still execute on `tx` after the publish — an app has been told a payment settled or failed while the row reverts to PENDING or never persists. `list_transactions` / `lookup_invoice` will then contradict the notification, and notifications are not retractable, so the divergence is permanent for that payment. Needs a commit failure to manifest, so it should be rare, but the resulting state is confusing to debug after the fact.

The correct pattern already exists in the same file: `markHoldInvoiceAccepted` closes its `svc.db.Transaction` closure at `:999` and publishes at `:1006-1011`, after the commit. `CancelHoldInvoice` does both at once — it publishes `nwc_hold_invoice_canceled` post-commit at `:1384-1387`, while the `markPaymentFailed` it calls inside its own closure publishes `nwc_payment_failed` pre-commit.

### Suggested fix

Have `markTransactionSettled` / `markPaymentFailed` return the events to publish instead of publishing directly, and have each caller fire them after its `db.Transaction(...)` returns nil — matching `markHoldInvoiceAccepted`. `checkBudgetUsage` should move out of the transaction for the same reason.

Anchors are against master `355454a4`.

rolznz:
Closes #2506

## Problem

`markTransactionSettled` and `markPaymentFailed` published `nwc_payment_sent` / `nwc_payment_received` / `nwc_payment_failed` (and `checkBudgetUsage` published `nwc_budget_warning`) while still inside the caller's database transaction. Since event delivery is asynchronous and immediate, connected apps and the Alby API could be notified of a payment whose row never commits, and any subscriber reading the database in response to an event could race with the commit and observe stale state.

## Approach

Rather than threading events (or a publisher) through every call site, every function that writes transaction state now **owns its own database transaction** and publishes its events only after the commit succeeds — the pattern `markHoldInvoiceAccepted` already used. This makes the bug structurally impossible: there is no way to call these functions inside someone else's transaction anymore.

- `markTransactionSettled(dbTransaction, ...)` and `markPaymentFailed(dbTransaction, reason)` open their own transaction; the 8 call sites that only wrapped them in `db.Transaction` collapse to single calls.
- The `nwc_lnclient_payment_received` / `payment_sent` handlers no longer create a transient PENDING row just to immediately settle it. When the LNClient notifies us of a payment we have no record of (payment made or received outside of Alby Hub on an external node, or a received keysend), the new `createSettledTransactionFromNotification` inserts the row **directly in its settled state** — a single atomic insert under the same payment-hash lock / settled-dedup guard. This also removes a latent race where duplicate concurrent events on Postgres committed a zombie PENDING row.
- `markPaymentFailed` now **refuses to mark a settled transaction as failed**. This replaces `CancelHoldInvoice`'s in-transaction `ACCEPTED` re-check and additionally protects the `SendPaymentSync` error path from a racing `nwc_lnclient_payment_sent` settle (the mirror ordering of the race covered by `TestNotifications_SentAfterMarkedPaymentFailed`).
- `checkBudgetUsage` returns the budget warning event instead of publishing it; the shared `afterTransactionSettled` helper collects it for post-commit publishing.

## Behavior changes

1. Events are published only after the state is durable (the fix).
2. Failing a settled payment is an error instead of a silent state stomp.
3. Double-cancelling a hold invoice is idempotent (previously returned a not-found error once the row was already FAILED; final DB state unchanged).

## Test plan

- New: `TestDoNotMarkSettledPaymentFailed` (settled-guard; also asserts no event escapes a refused operation), `TestMarkSettled_App_BudgetWarning` (first coverage of `nwc_budget_warning`).
- Existing `notifications_test.go` covers the reworked handlers, including the unknown-payment born-settled paths and their idempotency.
- `go test ./...` passes (SQLite); Postgres runs in CI.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved payment settlement reliability by preventing duplicate or conflicting updates.
  - Prevented settled payments from being incorrectly marked as failed.
  - Improved handling of concurrent settlement and failure requests.
  - Ensured payment and budget-warning notifications are delivered only after successful updates.
  - Added support for recognizing previously unknown incoming and outgoing transactions from notifications.
  - Improved error handling for payment notifications and transaction lookups.
  - Improved handling of app budget warnings when payments approach or exceed configured limits.
  - Fixed failed keysend payments so they are recorded correctly and release reserved fees.

### DIFF
--- a/tests/mock_ln_client.go
+++ b/tests/mock_ln_client.go
@@ -81,6 +81,8 @@ type MockLn struct {
 	MakeInvoiceErrors          []error
 	PayInvoiceResponses        []*lnclient.PayInvoiceResponse
 	PayInvoiceErrors           []error
+	PayKeysendResponses        []*lnclient.PayKeysendResponse
+	PayKeysendErrors           []error
 	PaymentDelay               *time.Duration
 	Pubkey                     string
 	MockTransaction            *lnclient.Transaction
@@ -110,6 +112,13 @@ func (mln *MockLn) SendPaymentSync(payReq string, amountMsat *uint64) (*lnclient
 }
 
 func (mln *MockLn) SendKeysend(amountMsat uint64, destination string, custom_records []lnclient.TLVRecord, preimage string) (*lnclient.PayKeysendResponse, error) {
+	if len(mln.PayKeysendResponses) > 0 {
+		response := mln.PayKeysendResponses[0]
+		err := mln.PayKeysendErrors[0]
+		mln.PayKeysendResponses = mln.PayKeysendResponses[1:]
+		mln.PayKeysendErrors = mln.PayKeysendErrors[1:]
+		return response, err
+	}
 	return &lnclient.PayKeysendResponse{
 		FeeMsat: 1,
 	}, nil
--- a/transactions/app_payments_test.go
+++ b/transactions/app_payments_test.go
@@ -62,6 +62,49 @@ func TestSendPaymentSync_App_WithPermission(t *testing.T) {
 	assert.Equal(t, dbRequestEvent.ID, *transaction.RequestEventId)
 }
 
+func TestMarkSettled_App_BudgetWarning(t *testing.T) {
+	svc, err := tests.CreateTestService(t)
+	require.NoError(t, err)
+	defer svc.Remove()
+
+	app, _, err := tests.CreateApp(svc)
+	assert.NoError(t, err)
+
+	appPermission := &db.AppPermission{
+		AppId:        app.ID,
+		App:          *app,
+		Scope:        constants.PAY_INVOICE_SCOPE,
+		MaxAmountSat: 100,
+	}
+	err = svc.DB.Create(appPermission).Error
+	assert.NoError(t, err)
+
+	// settling this payment pushes the app over 80% of its 100 sat budget
+	dbTransaction := db.Transaction{
+		AppId:       &app.ID,
+		State:       constants.TRANSACTION_STATE_PENDING,
+		Type:        constants.TRANSACTION_TYPE_OUTGOING,
+		PaymentHash: tests.MockLNClientTransaction.PaymentHash,
+		AmountMsat:  90000,
+	}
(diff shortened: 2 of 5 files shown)


rolznz:
@coderabbitai review

PR #2523, opened 2026-08-08
fix: bark onboarding, migration messaging and receive settlement for bark 0.6.0

### LINKED ISSUE #2512: Update bark onboarding + migration now that user can recover offchain funds from mnemonic alone
Follow-up of https://github.com/getAlby/hub/pull/2502

rolznz:
Follow-ups to the bark 0.6.0 SDK bump (#2502).

Closes #2512

## Onboarding + backup messaging (#2512)

Offchain funds are now recoverable from the mnemonic alone via the seed-derived recovery mailbox, so:

- **Setup Security**: removed the "During beta, your funds cannot be recovered from your recovery phrase alone" bullet; bark now shows the standard "balance can be recovered with your 12-word recovery phrase" guidance.
- **Settings → Backup**: removed the "Bark Support Coming Soon / recovery phrase is not sufficient" alert.
- Added a `recoveryreport` custom node command that returns the result of the seed-recovery scan bark runs when a wallet is created from an existing recovery phrase, so users migrating to a new device can verify their funds were restored (buckets: recovered/skipped/foreign/failed/exited + completeness).
- Updated the stale comment in `api.go` about why bark still requires persistent storage (in-flight payment checkpoints and wallet metadata are local-only).

## Bug fix: receives stuck pending on bark 0.6.0

bark 0.6.0 added a `delivering` receive state between `preimage-revealed` and `settled`. The claim handler didn't recognize it, so every claimed receive was published without a preimage and rejected by the transactions service (`no preimage in payment`) — paid invoices stayed "Waiting for Payment..." forever (and no NIP-47 notification was sent). Since delivery completion emits no further movement notification, there was no second chance to settle.

Fixed by treating all states at or past preimage reveal as paid via a `receiveIsPaid` helper. It's a positive allowlist, so an unknown future state degrades to "still pending" rather than "falsely settled" (the state is the only reliable paid-signal — bark stores the preimage from invoice creation, so `PaymentPreimage` being set means nothing). Also uses bark's own `settled_at` timestamp when available.

Reproduced and verified against `ark.signet.2nd.dev`.

## Import flow cleanup

The "I don't have another Alby Hub to migrate or open channels (funds from channels will be lost!)" checkbox on the import screen was required for every backend but only applied to LDK — and is wrong when dynamic channel backups (VSS) are enabled. Removed it; the channels caveat now appears on the Security & Recovery page instead, only when a mnemonic was imported and LDK was chosen.

## Summary by CodeRabbit

* **New Features**
  * Added recovery scan status reporting, including unavailable and completed recovery results.
  * Improved recovery guidance for imported nodes.

* **Bug Fixes**
  * Improved payment settlement handling so transaction details are recorded only after a payment is confirmed.
  * Updated recovery messaging to clarify that spendable funds can be recovered from a mnemonic, while payment checkpoints and wallet metadata require persistent storage.

* **Documentation**
  * Removed outdated warnings suggesting recovery phrases were insufficient during beta.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1730,10 +1730,12 @@ func (api *api) Setup(ctx context.Context, setupRequest *SetupRequest) error {
 		return errors.New("no unlock password provided")
 	}
 
-	// Bark and Cashu both store wallet state on local disk and have no
-	// remote-backup mechanism, so they cannot run in environments without
-	// persistent volumes (e.g. Alby Cloud). The default OAuth client ID
-	// identifies a local / self-hosted deployment.
+	// Bark and Cashu both store wallet state on local disk, so they cannot
+	// run in environments without persistent volumes (e.g. Alby Cloud). Bark
+	// can recover spendable VTXOs from the mnemonic alone, but in-flight
+	// payment checkpoints and wallet metadata are local-only, so persistent
+	// storage is still required. The default OAuth client ID identifies a
+	// local / self-hosted deployment.
 	if !api.cfg.GetEnv().IsDefaultClientId() {
 		switch setupRequest.LNBackendType {
 		case config.BarkBackendType, config.CashuBackendType:
--- a/frontend/src/screens/settings/Backup.tsx
+++ b/frontend/src/screens/settings/Backup.tsx
@@ -125,16 +125,6 @@ export default function Backup() {
                   phrase, you will lose access to your funds.
                 </AlertDescription>
               </Alert>
-              {info?.backendType === "BARK" && (
-                <Alert variant="destructive">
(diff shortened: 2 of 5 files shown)


rolznz:
CC @PeteClubSeven there was a bug on our side due to not handling the new `delivering` state in 0.6.0

PeteClubSeven:
> CC @PeteClubSeven there was a bug on our side due to not handling the new `delivering` state in 0.6.0

Apologies, I figured it wouldn't matter because delivering should only be used for those running a forwarding service where lightning is received by a bark node and forwarded to the right user as an offline ark payment.

PR #2532, opened 2026-08-10
fix: validate return_to redirect URLs

rolznz:
\`return_to\` query parameters are now parsed and only http and https URLs are used for redirects (relative URLs are resolved against the hub origin):

- New shared \`safeReturnToUrl\` frontend helper, used by the new app connection screen and the connect peer screen, both when reading the query parameter and before navigating.
- The createApp API applies the same check before adding the relay/pubkey/lud16 connection parameters to the URL, with tests for \`buildReturnToUrl\`.

The production frontend build now also includes the same Content-Security-Policy meta tag that is served as a header in http mode, so the policy also applies where no HTTP headers are set, e.g. in the desktop app. The dev and prod CSP meta tags are now generated from a single \`buildCSP\` helper in \`vite.config.ts\`, with sync comments between it and \`http_service.go\`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of return URLs to prevent unsafe or unsupported redirects.
  * Preserved valid query parameters when returning to connected apps.
  * Correctly handles missing, invalid, relative, and non-HTTP(S) URLs.

* **Security**
  * Added Content Security Policy protection to production builds.
  * Improved consistency between development and production security settings.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -130,21 +130,7 @@ func (api *api) CreateApp(createAppRequest *CreateAppRequest) (*CreateAppRespons
 	responseBody.RelayUrls = relayUrls
 	responseBody.Lud16 = lightningAddress
 
-	if createAppRequest.ReturnTo != "" {
-		returnToUrl, err := url.Parse(createAppRequest.ReturnTo)
-		if err == nil {
-			query := returnToUrl.Query()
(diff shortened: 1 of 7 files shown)


rolznz:
@coderabbitai review

PR #2534, opened 2026-08-11
fix: remove legacy acceptance of empty unlock password check

rolznz:
## Summary

Tightens the unlock-password check and removes a legacy compatibility path.

- **`CheckUnlockPassword`** previously treated a missing or empty `UnlockPasswordCheck` value as a match. This dated from before the canary row was always written during setup. It now requires the stored value to be present and to equal the expected string.
- **`StartApp`** checks for the canary up front via a new `IsUnlockPasswordCheckSet()` helper and, when it is missing, stops with a clear message asking the user to restore from a backup instead of continuing to a password check it cannot verify.
- **`keys.Init`** now returns the error from reading `NostrSecretKey` instead of discarding it, so a read failure aborts rather than being read as "no key yet" and generating and saving a replacement key.

## Tests

- New: `TestCheckUnlockPassword_MissingCanaryFailsClosed` and `TestCheckUnlockPassword_NoPasswordHub` (DB-backed, SQLite + Postgres), and `TestInit_WrongPasswordDoesNotOverwriteNostrKey`.
- Updated existing config/JWT/backup tests that had used `ChangeUnlockPassword("", pw)` to initialise a password; they now set up the canary via `SaveUnlockPasswordCheck`, and `TestJWTSecret_ChangePassword` performs a real password change.
- `tests/mocks/Config.go` regenerated for the new interface method.

`go test ./...` passes; `gofmt` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved unlock-password validation by requiring a valid stored verification check.
  - Added fail-closed behavior when the verification check is missing.
  - Preserved support for empty passwords on unencrypted hubs.
  - Startup now reports clear errors when password verification data is unavailable.
  - Prevented an incorrect password from overwriting an existing secret key.
  - Encryption and decryption errors are now surfaced during key initialization.
- **Tests**
  - Added coverage for missing verification data, empty-password hubs, and incorrect-password handling.

### DIFF
--- a/api/backup_test.go
+++ b/api/backup_test.go
@@ -39,6 +39,12 @@ func TestCreateBackup(t *testing.T) {
 	cfg, err := config.NewConfig(appConfig, gormDB)
 	require.NoError(t, err)
 
+	unlockPassword := ""
+
+	// Represent a fully set-up hub: the unlock-password canary is written during
+	// setup and is required for the password check to pass.
+	require.NoError(t, cfg.SaveUnlockPasswordCheck(unlockPassword))
+
 	app := &db.App{
 		Name:      "test",
 		AppPubkey: "2b7dea2866958f17c568cf024e113db7a3baa9c253a9016889196b8d0b11c7ae",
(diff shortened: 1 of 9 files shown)


rolznz:
@coderabbitai review

PR #2536, opened 2026-08-11
fix: validate swap out invoice before payment

rolznz:
Swap out (Lightning → on-chain) now validates the invoice returned by the swap provider before it is stored and paid:

- the invoice payment hash must match the payment hash of the locally generated preimage
- the invoice amount must not exceed the requested amount plus the quoted service and miner fees (with a small rounding tolerance; the bound uses the same formula the provider uses to compute the invoice amount, and the fee quote is locked via the pair hash)
- the lockup address is checked against the swap tree, matching the checks already performed for swap in and refunds
- the invoice is verified again directly before it is paid, covering swaps resumed from the database

Also renames `AlbySwapServiceFee` to `AlbySwapServiceFeePercentage` for clarity, and adds unit tests for the new validation helpers.

### DIFF
--- a/swaps/swaps_service.go
+++ b/swaps/swaps_service.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"strconv"
 	"sync"
@@ -68,7 +69,7 @@ type SwapsService interface {
 }
 
 const (
-	AlbySwapServiceFee = 1.0
+	AlbySwapServiceFeePercentage = 1.0
 )
 
 type SwapInfo struct {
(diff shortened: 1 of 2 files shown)


swaps/swaps_service.go
```diff
@@ -379,6 +381,44 @@ func (svc *swapsService) SwapOut(amountSat uint64, destination string, autoSwap,
 	}, nil
 }
 
+// swapOutInvoiceToleranceSat covers rounding differences that can occur when
+// the swap provider converts the requested on-chain amount into an invoice amount.
+const swapOutInvoiceToleranceSat = 10
```
im-adithya, line 386: Everything else looks good but I'm just worried if this is a hacky approach (the rest is just reverse-engineered from boltz library so that should be okay)

rolznz: I had the same concern, but it checked the boltz library directly for this. Once Boltz is back online I'll test it.

PR #2540, opened 2026-08-12
fix: switch unlock rate limiter from per-IP to global

rolznz:
## Description

The rate limiter on the unlock endpoints was keyed per client IP. That identifier comes from request headers, so it is chosen by the caller rather than being a reliable client identity.

This switches to a single global rate limiter (one bucket for all callers) and applies it to every endpoint that verifies the unlock password:

- `POST /api/start`
- `POST /api/unlock`
- `POST /api/backup`
- `POST /api/mnemonic`
- `POST /api/apps`
- `POST /api/autoswap`
- `PATCH /api/unlock-password`
- `PATCH /api/auto-unlock`

The limiter allows 1 request/second with a small burst of 2, so unlocking and then immediately performing an action (e.g. creating an app) is not blocked. `/logout` no longer shares the limiter.

## Testing

- `TestUnlock_RateLimited` — the burst is served, then further requests get HTTP 429.
- `TestUnlock_RateLimitNotBypassedBySpoofedIP` — varying `X-Forwarded-For` per request does not grant a fresh bucket.
- Existing unlock/app tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened protection for password-sensitive actions, including unlocking, backup, app creation, mnemonic retrieval, and auto-swap activation, with a shared request limit.
  * Requests exceeding the limit now receive a rate-limit response, even when using different forwarded IP addresses.
  * Logout remains available without rate limiting.

* **Tests**
  * Added coverage confirming rate limiting after repeated failed unlock attempts and verifying consistent behavior across forwarded IP addresses.


rolznz:
@coderabbitai review

rolznz:
@coderabbitai review

Issue #2557, opened 2026-08-21
feat: support max fee new param to pay_invoice & bip321 pay NWC


frnandu:
https://github.com/nostr-protocol/nips/pull/2444

- [ ] pay_invoice - #2566 
- [ ] bip321 pay

PR #2566, opened 2026-08-27
feat: max fee new param to pay_invoice

frnandu:
relates #2557

## Summary by CodeRabbit

- **New Features**
  - Added an optional maximum-fee setting for invoice payments.
  - Payment requests can now specify a fee limit through supported payment interfaces.
  - Supported Lightning backends enforce the requested fee cap where available.

- **Compatibility**
  - Backends without maximum-fee support clearly reject requests that use this option.

- **Tests**
  - Added coverage verifying maximum-fee values are forwarded and enforced correctly.

### DIFF
--- a/api/models.go
+++ b/api/models.go
@@ -45,7 +45,7 @@ type API interface {
 	GetBalances(ctx context.Context) (*BalancesResponse, error)
 	ListTransactions(ctx context.Context, appId *uint, limit uint64, offset uint64, filters ListTransactionsFilters) (*ListTransactionsResponse, error)
 	ListOnchainTransactions(ctx context.Context) ([]OnchainTransaction, error)
-	SendPayment(ctx context.Context, invoice string, amountMsat *uint64, metadata map[string]interface{}, fromAppId *uint) (*SendPaymentResponse, error)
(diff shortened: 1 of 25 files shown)


rolznz:
`CalculateFeeReserveMsat` currently relies on a 1% fee. I am not sure we want to open up the attack surface here, and what is the goal? to enable users to optionally accept paying more than 1%?

frnandu:
The main goal of max_fee is not necessarily to allow fees above 1%. It is to give the NWC client an end-to-end safety bound. A client  using an NWC connection should not need to know whether the underlying wallet is LDK, LND, CLN, Cashu, or something else—or whether that implementation defaults to 1%, another value, or no explicit limit.

When handling an untrusted recipient invoice, the client should be able to say: “Do not pay more than X msat in fees.” The wallet service should then either enforce that ceiling or reject the request. This protects the user against malicious invoices or routes involving attacker-controlled, unusually expensive channels.

I agree that a client-provided value should not unintentionally weaken a stricter wallet policy. We could therefore define it as an additional ceiling: where the wallet already has a default limit, use the stricter of the wallet limit and max_fee; where no such limit exists, enforce the client limit if supported; otherwise fail the request. With max_fee omitted, existing backend behavior remains unchanged.

PR #2583, opened 2026-09-08
fix: preserve rotated oauth token across persistence failures

### LINKED ISSUE #2581: Alby Account session lost when refreshed OAuth token fails to save
## Problem

A hub lost its Alby Account session and required the user to log in manually, even though the OAuth refresh token should be long-lived.

Likely cause: in `saveToken` (`alby/alby_oauth_service.go`), the expiry, access token and refresh token are each written once with `SetUpdate`, and a failure is only logged (`Failed to save refresh token`). If that write fails, the DB keeps the **old** refresh token while the server has already rotated it. The hub then works until the access token expires, after which every refresh fails with `invalid_grant` until the user re-authenticates.

`fetchUserToken` always re-reads the token from the DB, so the freshly refreshed token is discarded even though it is the only valid credential left.

Applies to both SQLite and PostgreSQL (busy/lock timeout, connection error, or DB closing during shutdown).

## Proposed fix (minimal, no refactor)

1. **Retry the token save** a few times with a short delay (e.g. 3 attempts, ~500ms–1s apart) for each of the three keys in `saveToken`.
2. **Keep the refreshed token in memory** in `albyOAuthService` (guarded by the existing `tokenMutex`). `fetchUserToken` should prefer the in-memory token when it is newer than the DB copy, and re-attempt the save on the next call if persisting failed. This way a failed save never strands the hub with a rotated refresh token.

## How to verify

- Make `SetUpdate` for `AlbyOAuthRefreshToken` fail once during a refresh; the hub should keep working and persist the token on a later attempt.
- Run with both SQLite and `TEST_DATABASE_URI` set to PostgreSQL.

Tyagiquamar:
## Summary

- Fixes Alby Account session loss when OAuth refresh succeeds but persisting the rotated refresh token fails transiently (fixes #2581).
- Retries durable token persistence (3 attempts) and bridges failed saves with in-memory state tied to `pendingRefreshFrom`, so a consumed refresh token is not reused from the database.
- Clears in-memory OAuth state on unlink, access-token removal, and failed/wrong-account callback flows.

## Test plan

- [x] `go test ./alby -run TestFetchUserToken|TestRemoveOAuth|TestLockAndClear -count=1`
- [x] `go test ./alby -race -run TestFetchUserToken|TestRemoveOAuth|TestLockAndClear -count=20`
- [ ] CI SQLite + Postgres matrix

## Notes

- Does not close the crash window where the process dies after provider rotation but before any successful DB write; atomic OAuth config persistence would be a separate follow-up.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OAuth token refresh reliability when token storage temporarily fails.
  - Prevented concurrent refreshes from invalidating rotated refresh tokens.
  - Ensured the most current valid token is used when credentials differ.
  - Cleared cached OAuth credentials when accounts are removed or unlinked.
  - Improved recovery from transient token persistence failures.

- **Tests**
  - Added coverage for token rotation, retry behavior, concurrent refreshes, token reconciliation, and cleanup.

### DIFF
--- a/alby/alby_oauth_service.go
+++ b/alby/alby_oauth_service.go
@@ -321,6 +321,9 @@ func (svc *albyOAuthService) fetchUserToken(ctx context.Context) (*oauth2.Token,
 	if shouldUseLatestToken(svc.latestToken, svc.pendingRefreshFrom, currentToken) {
 		currentToken = svc.latestToken
 	}
+	if svc.pendingRefreshFrom != "" {
+		svc.saveToken(currentToken, svc.pendingRefreshFrom)
+	}
 
 	// only use the current token if it has at least 60 seconds before expiry
 	if currentToken.Expiry.After(time.Now().Add(time.Duration(60) * time.Second)) {
@@ -589,7 +592,6 @@ func (svc *albyOAuthService) GetAuthUrl() string {
 }
 
 func (svc *albyOAuthService) UnlinkAccount(ctx context.Context) error {
-	svc.lockAndClearTokenState()
 	ldkVssEnabled, err := svc.cfg.Get("LdkVssEnabled", "")
 	if err != nil {
 		logger.Logger.WithError(err).Error("Failed to fetch LdkVssEnabled user config")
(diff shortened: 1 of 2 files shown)


Tyagiquamar:
Hi, friendly ping for review on this PR when you have a moment. Happy to address any feedback. Thanks!

Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

rolznz:
Hi, thanks for the PR. The change looks overly complex however. I think simply adding retries should be sufficient.

PR #2587, opened 2026-09-14
feat: authenticate the web UI with a session cookie

bumi:
## Summary

Alternative to #2584. Same goal — let an HTTPS reverse proxy in front of Alby Hub use HTTP Basic Authentication — but the web UI session moves into an HttpOnly cookie instead of a custom `X-Alby-Auth` header.

A browser only ever sends one `Authorization` header, and the web UI setting it to its bearer token suppresses the Basic credentials the proxy asked the user for. A cookie is a separate header, so `Cookie: albyhub_session=…` and `Authorization: Basic …` ride along on the same request with nothing to work around.

## Why a cookie rather than a custom header

- `Cookie` is a credential that every log scrubber, proxy and cache already recognises. A custom header is scrubbed by nothing and dropped by proxies with header allowlists.
- Browsers scope cookies to origin and path, so the cross-origin redirect leak that a custom header needs `redirect: "error"` to avoid cannot happen in the first place.
- The token leaves `localStorage`, where it was readable by any XSS.
- There is no second auth mechanism to document or keep in sync: `Authorization: Bearer` stays the one thing external API clients use.

## Server

- `/api/unlock` and `/api/start` set the session cookie: `HttpOnly`, `SameSite=Strict`, `Path=/`, expiring together with the JWT.
- `Secure` is set only when the request is HTTPS (`c.Scheme()`, which honours `X-Forwarded-Proto`). Alby Hub is routinely served over plain HTTP on a local network, where a browser would silently drop a Secure cookie.
- `echojwt` looks the token up as `cookie:albyhub_session,header:Authorization:Bearer `, so external clients keep working unchanged.
- `/logout` now ends the session by expiring the cookie instead of only redirecting.
- `/api/unlock` doubles as the way to mint a token for external clients in Developer Settings. Those requests pass `createApiToken: true` and get no cookie — otherwise creating a readonly token would downgrade the session of the user creating it.
- Every `/api` response gets `Cache-Control: private, no-store` and `Vary: Cookie`. A request carrying an `Authorization` header may not be stored by a shared cache (RFC 9111 §3.5) but a cookie gets no such treatment, so it has to be said explicitly. Same fix as in #2584; needed either way.
- The info handler accepts both credentials and no longer panics on a bare `Authorization: Bearer` header (`parts[1]` on a one-element slice).

## Frontend

- The request helper stops attaching credentials — fetch sends the cookie on same-origin requests by default.
- `saveAuthToken` / `getAuthToken` / `deleteAuthToken` are gone. A one-time `deleteLegacyAuthToken()` clears the token older versions left in `localStorage`, which otherwise stays there until it expires.
- Logout goes straight to `/logout`.

## Upgrade note

Everyone is logged out once when upgrading, because the token in `localStorage` is no longer sent. Entering the unlock password again is all that is needed.

## Trade-offs

- `SameSite=Strict` is safe here because nothing auth-gated is reached through a cross-site top-level navigation — the Alby OAuth callback at `/api/alby/callback` is unauthenticated, and the HTML and assets are not gated, so the first XHR after the page loads is same-site and carries the cookie. `Lax` is the fallback if that ever changes.
- CSRF: every mutating route is a POST/PATCH/DELETE with a JSON body, which a cross-site form cannot produce, on top of `SameSite=Strict`. `middleware.CSRF` on the full access group would be belt and braces.

## Testing

- `go test ./...`
- New coverage: the cookie authenticating a request that also carries `Authorization: Basic`, the Secure flag behind an HTTPS proxy, `createApiToken` not starting a session, logout expiring the cookie with matching attributes, and `/api/info` recognising both credential types.
- `yarn lint:js`, `yarn format`, `yarn tsc:compile`


rolznz:
Closing in favor of #2590, a simplified alternative that solves this on the Caddy side only.

PR #2588, opened 2026-09-15
feat(scripts): add optional Caddy password gate to the install scripts

bumi:
## The problem

Our Caddy examples suggest wrapping Alby Hub in `basic_auth`:

```caddy
#basicauth {
    # Bob $2a$14$Zkx19XLiW6VYouLHR5NmfOFU0z2GTNmpkT/5qqR7hx4IjWJPDhjvG
#}
```

That snippet doesn't work. Alby Hub authenticates its own API with `Authorization: Bearer <JWT>`, and the frontend sets that header explicitly in [`request.ts`](https://github.com/getAlby/hub/blob/master/frontend/platform_specific/http/src/utils/request.ts#L18-L21). An explicit `Authorization` header replaces the `Authorization: Basic ...` header the browser attaches after a Basic login, so every API call reaches Caddy without Basic credentials and gets a 401. Pages load, the UI is broken. Passing the Basic header through instead doesn't help either — then Alby Hub never gets its Bearer token. One header, two layers that both want it.

## The fix

Basic Auth is used for the **first request only**. On success Caddy issues a random `HttpOnly` gate cookie, and later requests are judged on that cookie — which leaves the `Authorization` header free for Alby Hub.

```
browser --(no cookie)-------------------------> Caddy --(401 Basic)--> browser
browser --(username + password)---------------> Caddy --(Set-Cookie)-> Alby Hub
browser --(cookie + Authorization: Bearer ...)-> Caddy ---------------> Alby Hub
```

## What's in here

**`scripts/caddy-gate/`** — the canonical setup: `Caddyfile.example`, `gate.env.example`, a systemd drop-in, `rotate-gate-secret.sh`, and a README covering the mechanism, manual setup, rotation and troubleshooting.

**`scripts/linux-{x86_64,aarch64}/install.sh`** — an optional prompt after the systemd question. Flags: `--caddy-gate`, `--no-caddy-gate`, `--domain`, `--gate-user`; password read hidden from the terminal or from `ALBYHUB_GATE_PASSWORD` for non-interactive runs. Installs Caddy from the official apt repo on Debian/Ubuntu when it isn't present, generates the secrets, writes the Caddyfile, runs `caddy validate` and only then restarts Caddy.

```bash
/bin/bash -c "$(curl -fsSL .../install.sh)" -- -s --domain hub.example.com
```

**The existing `Caddyfile.example`s** — drop the broken `basicauth` snippet, explain why it can't work, point at the working config.

## Design notes for review

- **Fail-closed guard.** The `@gated` expression starts with `{http.request.cookie.__Host-albyhub_gate} != ""`. Without it, an absent cookie and an unset `GATE_SECRET` both expand to `""`, compare equal, and *every* request sails through — a missing `gate.env` would silently disable the gate. This is the single most important line in the config.
- **Authorization is only stripped when it's a Basic credential** (`@basic_credentials header Authorization Basic*`) in the gated path. Stripping it unconditionally there would delete Alby Hub's own JWT and break the API.
- **Two secrets + sliding expiry.** `GATE_SECRET_OLD` is accepted during rotation, and every gated response re-issues the cookie with the current secret, so rotating doesn't sign out anyone who's active. `GATE_SECRET_OLD` gets a random value on first install rather than an empty one, so it's never guessable.
- **`__Host-` cookie prefix** forces `Secure` + `Path=/` + no `Domain`, so a sibling subdomain can't overwrite the gate cookie. It does mean HTTPS is required, which is why the setup asks for a domain.
- **`SameSite=Lax`** rather than `Strict`, so returning from a cross-site OAuth redirect keeps the cookie. The gate isn't what authorizes state changes (Alby Hub's JWT is), so `Lax` is the right trade-off here.
- **systemd `EnvironmentFile`** is read by systemd as root before Caddy drops privileges, so `/etc/caddy/gate.env` stays `chmod 600` root-only. A `reload` does *not* re-read it — the rotation script does a full `restart`.
- **`caddy hash-password` needs a trailing newline on stdin.** It uses `ReadBytes('\n')` and returns the EOF error if there isn't one, so the script pipes with `printf '%s\n'`.

Caddy v2.8+ is required (`basicauth` → `basic_auth` rename); the script checks and bails out with a clear message on older versions.

## Framing

The docs are explicit that this is defence in depth, not permission to put Alby Hub on the public internet — a VPN is still the recommendation, and the gate is an extra layer on top of it and of Alby Hub's own unlock password. The install script also reminds you to firewall the direct port, since the gate is pointless if `:8029` is still reachable.

## Testing

Shell scripts pass `sh -n` and `shellcheck -s sh` (only the two pre-existing `SC2162` infos remain). The generated Caddyfile was rendered with sample values and reviewed. Caddy directive and matcher syntax was checked against the current docs and the `celmatcher.go` / `caddyauth/command.go` sources.

**Not verified end-to-end on a live server** — the config was never loaded into a running Caddy. Worth a manual run of the install script against a test domain before merging.

https://claude.ai/code/session_01Fx6WEWkkHFs5XGLcyLoSuj

## Summary by CodeRabbit

- **New Features**
  - Linux installers can optionally configure an HTTPS Caddy password gate for Alby Hub.
  - Added support for custom domains, gate usernames, passwords, protected secrets, and secret rotation.
  - The gate preserves Alby Hub API authentication while protecting browser access with a secure session cookie.
  - Installer behavior now reflects systemd and non-systemd deployment options and ports.

- **Documentation**
  - Added setup, troubleshooting, rotation, and hardening guidance for the Caddy password gate.
  - Updated Caddy examples with recommended authentication guidance and deployment details.


rolznz:
To simplify it I think I would undo the install.sh changes (to decouple caddy from the install) and also just remove the `Caddyfile.example` in `linux-aarch64` and `linux-x86_64`, and just link to the README of the new caddy-gate folder instead of adding new sections in each of the READMEs.

rolznz:
Closing in favor of #2590, a simplified alternative that solves this on the Caddy side only.

PR #2589, opened 2026-09-15
fix: show fee row for zero-fee outgoing payments

### LINKED ISSUE #2565: Bug: fee does not show for sent payments if fee is 0
Image

Abhinav-Malik-154:
Fixes #2565

The Fee row in `TransactionItem.tsx` was gated on `tx.feesPaidMsat > 0`,
so a genuinely fee-free outgoing payment (0 msat fee) hid the row
entirely instead of showing "Fee: 0 sats" — a real Lightning payment can
legitimately route for free.

Changed the condition to `tx.state != "failed" && type == "outgoing"`
instead, matching the pattern already used for the percentage annotation
directly below it. Since the outer condition now already guarantees
`type == "outgoing"`, the now-redundant inner check was removed too.

Verified with `yarn tsc:compile` and `yarn lint:js` — both clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Transaction details now show fee information for all non-failed payments, including incoming payments.
  - Percentage breakdowns appear only for outgoing payments with a positive amount.
  - Failed payments continue to omit fee information.
  - Transaction descriptions are displayed when available.


frontend/src/components/TransactionItem.tsx
```diff
@@ -277,13 +277,13 @@ function TransactionItem({ tx, transactionListKey }: Props) {
           <TransactionDetailRow label="Date & Time">
             {updatedAt.format("D MMMM YYYY, HH:mm")}
           </TransactionDetailRow>
-          {tx.state != "failed" && tx.feesPaidMsat > 0 && (
+          {tx.state != "failed" && type == "outgoing" && (
```
rolznz: We can have fees on incoming payments too (e.g. JIT payments), so we shouldn't filter on the type here.

Abhinav-Malik-154:
Good catch, thanks- didn't think about JIT channel fees on the receive
side. Removed the direction filter on the row itself and kept the
outgoing-only check on the percentage annotation since that's
percentage-of-amount-sent specifically, not something the original bug
report touched.

frontend/src/components/TransactionItem.tsx
```diff
@@ -277,22 +277,17 @@ function TransactionItem({ tx, transactionListKey }: Props) {
           <TransactionDetailRow label="Date & Time">
             {updatedAt.format("D MMMM YYYY, HH:mm")}
           </TransactionDetailRow>
-          {tx.state != "failed" && tx.feesPaidMsat > 0 && (
+          {tx.state != "failed" && tx.feesPaidMsat >= 0 && (
```
rolznz: Can we remove ` && tx.feesPaidMsat >= 0`?

frontend/src/components/TransactionItem.tsx
```diff
@@ -277,22 +277,17 @@ function TransactionItem({ tx, transactionListKey }: Props) {
...
                 <>
-                  &nbsp;(
-                  {((tx.feesPaidMsat / tx.amountMsat) * 100).toFixed(2)}%)
+                  &nbsp;({((tx.feesPaidMsat / tx.amountMsat) * 100).toFixed(2)}
+                  %)
                 </>
               )}
             </TransactionDetailRow>
           )}
-          {tx.description && (
```
rolznz, line 291: why was this removed?

Abhinav-Malik-154: That was accidental - got dropped during an earlier edit to the fee
condition, not intentional. Restored it now.

frontend/src/components/TransactionItem.tsx
Abhinav-Malik-154: Agreed, removed it - feesPaidMsat is never actually negative in
practice, so that check wasn't doing anything real.

Abhinav-Malik-154:
Good catch - that description block was deleted by accident during one
of the earlier edits to the fee condition, not an intentional change.
Restored it.

And yes, agreed on removing `>= 0` - feesPaidMsat is never actually
negative in practice, so that check wasn't doing anything real. Simplified
to just `tx.state != "failed"`.

Abhinav-Malik-154:
Thanks for the context makes sense this needs more thought than a
simple default swap. Happy to help implement whichever direction you two
land on.

rolznz (approved):
utACK, thanks!

Abhinav-Malik-154:
Thanks for the guidance throughout - appreciate you taking the time to
review this carefully.

PR #2593, opened 2026-09-18
fix: avoid blocking on Boltz swap info

im-adithya:
Boltz is down, so the `/wallet/receive/onchain` screen is stuck waiting for swap info. This makes the screen open with “Other Cryptocurrency” by default, only loads Boltz info when Bitcoin is selected and adds a 5s timeout so it fails gracefully instead of hanging forever.

## Summary by CodeRabbit

* **New Features**
  * Bitcoin on-chain swaps are now optional during the receive flow.
  * Added a warning when swap information is unavailable, with the Continue button disabled until the service is available.

* **Bug Fixes**
  * The receive flow now handles swap-service failures gracefully instead of displaying an error caused by missing data.
  * Swap limits and fees are shown only when current swap information is available.
  * Added a timeout for requests to the swap service.


rolznz (approved):
tACK

PR #2595, opened 2026-09-18
fix: use electrum as default chain source

### LINKED ISSUE #2563: Change default chain source to electrum
It is less resource heavy than esplora and from our testing so far works just as well.

rolznz:
Closes #2563

## What

- `LDK_ELECTRUM_SERVER` now defaults to `ssl://electrum.getalbypro.com:50002`; `LDK_ESPLORA_SERVER` has no default and becomes the opt-in override. The chain source selection in `ldk.go` is flipped accordingly (bitcoind → esplora if set → electrum), and the previously missing log line for the electrum branch is added.
- `GetChainDataSource()` (node page) follows the same order.
- `render.yaml` no longer pins `LDK_ESPLORA_SERVER`, so Render deploys follow the default.

## Address lookups

The unused-address check (`GetUnusedOnchainAddress`) and the auto swap xpub lookahead were calling the esplora REST API directly, regardless of the configured chain source. With no esplora default they would have broken, so they now go through a small `chain.AddressLookup`:

- `LDK_ESPLORA_SERVER` set → esplora `/address/:addr/txs` (same as before)
- otherwise → electrum `blockchain.scripthash.get_history` via `github.com/BoltzExchange/go-electrum` (already in the module graph through boltz-client; promoted to a direct dependency)
- bitcoind has no address index, so it also falls back to the electrum server; a warning with the server URL is logged at startup so self-hosters know to set `LDK_ELECTRUM_SERVER` to their own electrs

The swap lookahead opens one electrum connection for the whole scan instead of one request per address.

## Notes

- Non-mainnet setups that previously only set `LDK_ESPLORA_SERVER` keep working unchanged since esplora still wins when configured. The README Mutinynet section now leads with `LDK_ELECTRUM_SERVER`.
- The electrum client validates TLS certificates with the default `tls.Config`, matching LDK's rust electrum client, so a server that works for chain sync works here too.
- Verified live against the Alby electrum and esplora servers: known-used addresses → used, random fresh P2WPKH → unused, on both backends.

## Summary by CodeRabbit

- **New Features**
  - Added support for checking on-chain address activity through either Electrum or Esplora.
  - Unused addresses are now identified consistently across on-chain address and swap workflows.

- **Improvements**
  - Electrum is now the default chain source, with Esplora available as an alternative.
  - Bitcoind configurations can use Electrum for address activity lookups.

- **Documentation**
  - Updated setup guidance and Mutinynet examples to reflect the new chain-source defaults.


chain/electrum.go
```diff
@@ -0,0 +1,124 @@
...
+	history, err := e.client.GetHistory(ctx, scriptHash)
+	if err != nil {
+		return false, fmt.Errorf("failed to get address history: %w", err)
+	}
+
+	return len(history) > 0, nil
+}
+
+func (e *electrumLookup) Close() {
+	e.client.Shutdown()
```
frnandu, line 84: [P2] Fix the client shutdown leak before using it for each lookup

The pinned go-electrum client's Shutdown() closes the client listener, while TCPTransport.listen() sends the resulting socket error on an unbuffered channel with no remaining receiver. Repeating the same connect → ServerVersion → GetHistory → Shutdown sequence against a local mock server increased the goroutine count from 2 to 102 after 100 successful lookups; all 100 leaked goroutines were blocked at electrum/transport.go:68. Since the receive-address API creates and closes a client on each check, these accumulate for the lifetime of the Hub process. Please fix or replace the dependency's shutdown handling and add a regression check that repeated lookup/close cycles release their goroutines.

config/models.go
```diff
@@ -30,8 +30,8 @@ type AppConfig struct {
 	LogToFile                          bool   `envconfig:"LOG_TO_FILE" default:"true"`
 	Network                            string `envconfig:"NETWORK"`
 	LDKNetwork                         string `envconfig:"LDK_NETWORK"`
-	LDKEsploraServer                   string `envconfig:"LDK_ESPLORA_SERVER" default:"https://electrs.getalbypro.com"` // TODO: remove LDK prefix
-	LDKElectrumServer                  string `envconfig:"LDK_ELECTRUM_SERVER"`
+	LDKEsploraServer                   string `envconfig:"LDK_ESPLORA_SERVER"`
+	LDKElectrumServer                  string `envconfig:"LDK_ELECTRUM_SERVER" default:"ssl://electrum.getalbypro.com:50002"`
```
frnandu, line 34: [P2] Remove the Esplora overrides from both Raspberry Pi installers

scripts/pi-aarch64/install.sh:74 and scripts/pi-arm/install.sh:73 still write LDK_ESPLORA_SERVER=https://electrs.getalbypro.com into the generated systemd service. Because an explicitly configured Esplora server takes precedence, fresh installs through either script continue using Esplora despite this new default. Please remove both hardcoded overrides, as this PR already does for render.yaml.

frnandu (changes requested):
Please address the Electrum client goroutine leak and remove the remaining Raspberry Pi installer overrides so those installations also use the new default. The leak was reproduced against the pinned dependency with a local mock Electrum server: 100 successful lookups followed by shutdown left 100 listener goroutines blocked.

chain/electrum.go
rolznz: replaced with a simple custom electrum client

